PR #15, opened 2022-01-10
docs: replace weird emoji

mapleeit:


hanxiao:
It is actually correct. The "weird" one suggest "limited support"

hanxiao:
i intend to 
image

i can use this one ✔️ instead

hanxiao:
fixed

PR #16, opened 2022-01-10
docs(readme): polish

alexcg1:
Basic polishing

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,15 +12,15 @@
 
 <!-- start elevator-pitch -->
 
-DocArray is a library for nested, unstructured data such as text, image, audio, video, 3D mesh. It allows deep learning engineers to efficiently process, embed, search, recommend, store, transfer the data with Pythonic API.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -220,7 +220,7 @@ groundtruth = DocumentArray(
     Document(uri=d.uri, matches=[Document(uri=d.uri.replace('left', 'right'))]) for d in left_da)
 ```
 
-Here we create a new DocumentArray with real matches by simply replacing the filename, e.g. `left/00001.jpg` to `right/00001.jpg`. That's all we need: if the predicted match has the identical `uri` as the groundtruth match, then it is correct.
+Here we create a new DocArray with real matches by simply replacing the filename, e.g. `left/00001.jpg` to `right/00001.jpg`. That's all we need: if the predicted match has the identical `uri` as the groundtruth match, then it is correct.
```
hanxiao, line 223: `DocArray` is the package name
DocumentArray is the class/object name. They can not be used interchangeably

hanxiao (changes requested):
misuse of DocArray

README.md
alexcg1: Understood and fixed

PR #20, opened 2022-01-10
docs: construct nested document line by line

bwanglzu:
create nested documents with one-liner could break the `granularity`, the recommended way: 
89D97911-CB46-4D27-81BB-2A135BEE0CC8

### DIFF
--- a/docs/fundamentals/document/construct.md
+++ b/docs/fundamentals/document/construct.md
@@ -96,13 +96,10 @@ Document can be nested inside `.chunks` and `.matches`. The nested structure can
 from docarray import Document
 
 d = Document(id='d0')
-chunk = Document(id='d1')
-chunk_of_chunk = Document(id='d2')
-d.chunks.append(chunk)
-d.chunks[0].chunks.append(chunk_of_chunk)
+d.chunks.append(Document(id='d1'))
+d.chunks[0].chunks.append(Document(id='d2'))
 
-match = Document(id='d3')
-d.matches.append(match)
+d.matches.append(Document(id='d3'))
 
 print(d)
 ```
--- a/docs/fundamentals/document/construct.md
+++ b/docs/fundamentals/document/construct.md
@@ -100,6 +100,7 @@ d.chunks.append(Document(id='d1'))
 d.chunks[0].chunks.append(Document(id='d2'))
 
 d.matches.append(Document(id='d3'))
+d.matches[0].matches.append(Document(id='d4'))
 
 print(d)
 ```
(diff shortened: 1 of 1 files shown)


hanxiao:
i dont see why original code does not work

JoanFM:
> i dont see why original code does not work

Because chunks are constructed before being added to the list.

JoanFM:
> > i dont see why original code does not work
> 
> Because chunks are constructed before being added to the list.

So a Document A is created standalone (therefore granularity 0) and then by creating a Document B with A as chunk, B does not update A's granularity.

hanxiao:
can you show the code and result because i believe i solved it

nan-wang:
@hanxiao Here is the codes to show the issue

```python
from docarray import Document
d = Document(
    id='d0',
    chunks=[Document(id='d1', chunks=Document(id='d2'))],
    matches=[Document(id='d3', matches=Document(id='d4'))],
)

for c in d.chunks:
    print(f'{c.id}: g={c.granularity}')
    for cc in c.chunks:
        print(f'{cc.id}: g={cc.granularity}')

for m in d.matches:
    print(f'{m.id}: g={m.granularity}, a={m.adjacency}')
    for mm in m.matches:
        print(f'{mm.id}: g={mm.granularity}, a={mm.adjacency}')
```

The outputs is as below where 
- `d2` is expected to have `granularity=2`
- `d4` is expected to have `adjacency=2`

```diff
d1: g=1
d2: g=1
d3: g=0, a=1
d4: g=0, a=1
```

docs/fundamentals/document/construct.md
```diff
@@ -95,11 +95,11 @@ Document can be nested inside `.chunks` and `.matches`. The nested structure can
...
-d = Document(
-    id='d0',
-    chunks=[Document(id='d1', chunks=Document(id='d2'))],
-    matches=[Document(id='d3')],
-)
+d = Document(id='d0')
+d.chunks.append(Document(id='d1'))
+d.chunks[0].chunks.append(Document(id='d2'))
+
+d.matches.append(Document(id='d3'))
```
nan-wang, line 102: We can further show the adjacency as well
```suggestion
d.matches.append(Document(id='d3'))
d.matches[0].matches.append(Document(id='d4'))
```

nan-wang (changes requested):

bwanglzu:
as was discussed , it works at `g=1`, not `g>1`, such as 2

nan-wang (approved):
LGTM👍

hanxiao:
reject. it overreact. single nested works so don't rewrite all just because double nested don't work

hanxiao (changes requested):
reject as it overreacts

bwanglzu:
close the PR since not necessary.

PR #26, opened 2022-01-11
fix(pickle): hard code default pickle protocol

numb3r3:
From https://docs.python.org/3/library/pickle.html#pickle.DEFAULT_PROTOCOL

> An integer, the default protocol version used for pickling. May be less than HIGHEST_PROTOCOL. Currently the default protocol is 4, first introduced in Python 3.4 and incompatible with previous versions.
> 
> Changed in version 3.0: The default protocol is 3.
> 
> Changed in version 3.8: The default protocol is 4.

After py3.8, the default pickle protocol is changed from 3 to 4. To archive the serialization compatible, the default protocol is hard coded in this PR.

### DIFF
--- a/docarray/array/mixins/io/binary.py
+++ b/docarray/array/mixins/io/binary.py
@@ -4,7 +4,13 @@ import pickle
 from contextlib import nullcontext
 from typing import Union, BinaryIO, TYPE_CHECKING, Type, Optional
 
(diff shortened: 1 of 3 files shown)


numb3r3:
Close this PR as it is not necessary. Since `pickle.loads` will detect protocol automatically.

PR #28, opened 2022-01-11
fix(array): fix edge case on single boolean index

hanxiao:


docs/fundamentals/documentarray/access-elements.md
```diff
@@ -120,6 +120,8 @@ print(da)
 <DocumentArray (length=50) at 4513619088>
 ```
 
+Note that if the length of the boolean mask is smaller than the length of a DocumentArray, then the remaining part is padded to `False`.
```
davidbp, line 123: I would add 'The length of the boolean mask should be at least 2'.

davidbp (approved):
LGTM

PR #29, opened 2022-01-11
fix(array): fix getting docs by two ids

davidbp:
The `DocumentArray` method `__getitem__`  is incorrect in the case two ids are passed:

Minimal working example to showcase bug:

```python
In [2]: from docarray import Document, DocumentArray
   ...: 
   ...: da = DocumentArray([Document(id='1'),
   ...:                     Document(id='2'),
   ...:                     Document(id='3')])
```

Currently selecting with two doc ids produces 

```python
In [4]: da['1','2']

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/var/folders/05/h71x7gh54sx_5y43ppkq9_dw0000gq/T/ipykernel_8535/2066379862.py in <module>
----> 1 da['1','2']

~/Documents/jina_stuff/docarray/docarray/array/document.py in __getitem__(self, index)
    151                 if isinstance(_attrs, str):
    152                     _attrs = (index[1],)
--> 153                 return _docs._get_attributes(*_attrs)
    154             elif isinstance(index[0], bool):
    155                 return DocumentArray(itertools.compress(self._data, index))

~/Documents/jina_stuff/docarray/docarray/document/mixins/attribute.py in _get_attributes(self, *fields)
     20                 value = dunder_get(self, k)
     21             else:
---> 22                 value = getattr(self, k)
     23 
     24             ret.append(value)

AttributeError: 'Document' object has no attribute '2'
```

This PR solves this and produces
```python
In [3]: da['1']
Out[3]: <Document ('id',) at 1>

In [4]: da['1','2']
Out[4]: <DocumentArray (length=2) at 140552956821312>

In [5]: da['1','2','3']
Out[5]: <DocumentArray (length=3) at 140552956819056>
```

### DIFF
--- a/docarray/array/document.py
+++ b/docarray/array/document.py
@@ -146,11 +146,15 @@ class DocumentArray(AllMixins, MutableSequence[Document]):
                 and len(index) == 2
                 and isinstance(index[0], (slice, Sequence))
             ):
+
+                # edge case where 2 ids are passed
+                if isinstance(index[0], str) and index[0] in self._id2offset:
+                    return DocumentArray(self._data[self._id2offset[t]] for t in index)
+
                 _docs = self[index[0]]
                 _attrs = index[1]
                 if isinstance(_attrs, str):
                     _attrs = (index[1],)
-
                 return _docs._get_attributes(*_attrs)
             elif isinstance(index[0], bool):
                 return DocumentArray(itertools.compress(self._data, index))
--- a/docarray/array/document.py
+++ b/docarray/array/document.py
@@ -146,9 +146,12 @@ class DocumentArray(AllMixins, MutableSequence[Document]):
                 and len(index) == 2
                 and isinstance(index[0], (slice, Sequence))
             ):
-                # edge case where 2 ids are passed
-                if isinstance(index[0], str) and index[0] in self._id2offset:
+                # edge case where ids are passed as element selector
+                if isinstance(index[0], str) and all(
+                    i in self._id2offset for i in index
+                ):
                     return DocumentArray(self._data[self._id2offset[t]] for t in index)
+
                 _docs = self[index[0]]
                 _attrs = index[1]
                 if isinstance(_attrs, str):
--- a/docarray/array/document.py
+++ b/docarray/array/document.py
@@ -33,7 +33,7 @@ if TYPE_CHECKING:
 
 class DocumentArray(AllMixins, MutableSequence[Document]):
     def __init__(
-        self, docs: Optional['DocumentArraySourceType'] = None, copy: bool = False
+        self, docs: Optional["DocumentArraySourceType"] = None, copy: bool = False
     ):
         super().__init__()
         self._data = []
@@ -64,7 +64,7 @@ class DocumentArray(AllMixins, MutableSequence[Document]):
 
         :return: a Python dict.
         """
-        if not hasattr(self, '_id_to_index'):
+        if not hasattr(self, "_id_to_index"):
             self._rebuild_id2offset()
         return self._id_to_index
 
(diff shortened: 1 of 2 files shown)


docarray/array/document.py
```diff
@@ -146,11 +146,13 @@ def __getitem__(
                 and len(index) == 2
                 and isinstance(index[0], (slice, Sequence))
             ):
+                # edge case where 2 ids are passed
+                if isinstance(index[0], str) and index[0] in self._id2offset:
```
hanxiao, line 150: this logic itself is also problematic, it does not work with column selector https://docarray.jina.ai/fundamentals/documentarray/access-attributes/

hanxiao: you can try `da['2', 'id']` under your branch it will break

hanxiao: I would suggest a completely readthrough on "access attributes" chapter before fixing edge cases like this

nan-wang:
I think `da['1','2']` is confusing. It is not clear whether we want to select by `id` or to select attributes. Anyway, the part of selecting by ids does not work. I suggest remove this part for the time being. 

image

MWE:
```python

da = DocumentArray([Document(id='1'),
                    Document(id='2'),
                    Document(id='3')])

da['1', '2']
```

leads to the error,

```bash
  File "/Users/nanwang/Codes/jina-ai/docarray/toys/toy2.py", line 7, in <module>
    print(da['1', '2'])
  File "/Users/nanwang/Codes/jina-ai/docarray/docarray/array/document.py", line 154, in __getitem__
    return _docs._get_attributes(*_attrs)
  File "/Users/nanwang/Codes/jina-ai/docarray/docarray/document/mixins/attribute.py", line 22, in _get_attributes
    value = getattr(self, k)
AttributeError: 'Document' object has no attribute '2'
```

nan-wang:
One principle we can stick to is to use the first arg to select ids\index\slices and the second arg to select the attributes. If we follow this rule, the grammar for selecting by integers should be refactored to 

```diff
- da[1, 2, 3]
+ da[[1, 2, 3]]
```

This pattern is similar as the pandas `DataFrame.loc`. **The `id` in DocArray is equivalent to `index` in `pandas.DataFrame`.** Here is an example
```python
import pandas as pd

df = pd.DataFrame([['a', 1, 'xx'], ['b', 2, 'xx'], ['c', 3, 'xx']], columns=['name', 'age', 'place'])
df.set_index('name', inplace=True)

# select by one id
print(df.loc['a'])

"""output
age       1
place    xx
Name: a, dtype: object
      age place
"""

# select by multiple ids
print(df.loc[['a', 'b']])

"""output
name           
a       1    xx
b       2    xx
1
"""

# select by one id & one attribute
print(df.loc['a', 'age'])

"""output
1
"""

# select by multiple ids & one attribute
print(df.loc[['a', 'b'], 'age'])

"""output
name
a    1
b    2
Name: age, dtype: int64
"""

# select by multiple ids & multiple attribute
print(df.loc[['a', 'b'], ['age', 'place']])

"""output
      age place
name           
a       1    xx
b       2    xx
"""
```

davidbp:
This error that you point out @nan-wang  is precisely what I explained and solves this PR.

docarray/array/document.py
davidbp: Now it should be fixed. There is still the possibility that an id has exactly the same str than an element_selector or a column selector in which case I guess the notation could be ambiguous as Nan mentions below. 

I think this is a 'very stretched edge case'  but it's a possibility:

```

da = DocumentArray([Document(id='@r', text='a1'),
                    Document(id='2', text='a2'),
                    Document(id='3', text='a3')])
```

Then `da['@r']` should be a single item or a `DocumentArray` with all root items?

We could assume `id` strings do not start with `@`, otherwise the logic in getitem/setitem  is ambiguous. To me this is a reasonable assumption but I guess we should make a `Note` in the docs?

PR #35, opened 2022-01-12
fix: id_to_index initialization

numb3r3:
An empty DA instance cannot be assigned to a new DA instance, as shown in the following:
```

matches = DocumentArray()

query = Document()
query.matches = matches
```

yields
```
AttributeError: 'DocumentArray' object has no attribute '_id_to_index'
```


davidbp:
You mean a `Document` can' t have a `DocumentArray` in matches?

This works 

```python
matches = DocumentArray()
query = DocumentArray()
query.matches = matches
```

numb3r3:
> You mean a Document can' t have a DocumentArray in matches?

Not exactly, I mean "a Document can't have a **empty** DocumentArray in maches"

PR #42, opened 2022-01-12
fix: single doc set in docarray

davidbp:
Currently setting for example `.text` the attribute for a concrete `Document` in a  `DocumentArray`  results to setting the first character in the string. 

```python

# this works as expected in master
da = DocumentArray([Document(),Document(), Document()])
da[[0,1],'text'] = ['jina','jana']
print(da[:,'text'])

# this does NOT work as expected in master
da = DocumentArray([Document(),Document(), Document()])
da[[0],'text'] = 'jina'
print(da[:,'text'])

# this does NOT work as expected in master
import numpy as np
da = DocumentArray([Document(),Document(), Document()])
da[[0],'id'] = '1234'
print(da[:,'id'])
```
In master 
```
['jina', 'jana', '']
['j', '', '']
['1', 'b6ac7ce073ba11ecbdc7787b8ab3f5de', 'b6ac7d3073ba11ecbdc7787b8ab3f5de']
```

This happens because the setter has the sets the values in  `for _d, _vv in zip(_docs, _v)` and when the input data is an iterable this is not probably what the user would expect

In this PR 
```
['jina', 'jana', '']
['jina', '', '']
['1234', '92c2521e73ba11eca885787b8ab3f5de', '92c2526473ba11eca885787b8ab3f5de']
```

### DIFF
--- a/docarray/array/document.py
+++ b/docarray/array/document.py
@@ -211,6 +211,7 @@ class DocumentArray(AllMixins, MutableSequence[Document]):
         index: 'DocumentArrayIndexType',
         value: Union['Document', Sequence['Document']],
     ):
+
         if isinstance(index, (int, np.generic)) and not isinstance(index, bool):
             index = int(index)
             self._data[index] = value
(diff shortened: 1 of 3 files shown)


docarray/array/document.py
```diff
@@ -274,8 +275,11 @@ def __setitem__(
                         elif _a == 'embedding':
                             _docs.embeddings = _v
                         else:
-                            for _d, _vv in zip(_docs, _v):
-                                setattr(_d, _a, _vv)
+                            if len(_docs) == 1:
+                                setattr(_docs[0], _a, _v)
+                            else:
```
hanxiao, line 280: let me complete this logic

hanxiao (changes requested):
will take this branch

hanxiao (approved):
good to go

PR #48, opened 2022-01-14
Bind projector server to host

abduhbm:
This PR will add an option to `plot_embeddings` to bind embedding-projector server to given host. Setting this option to `0.0.0.0` will make the projector available on local network.

### DIFF
--- a/docarray/array/mixins/plot.py
+++ b/docarray/array/mixins/plot.py
@@ -106,12 +106,14 @@ class PlotMixin:
         min_image_size: int = 16,
         channel_axis: int = -1,
         start_server: bool = True,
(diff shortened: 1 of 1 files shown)


docarray/array/mixins/plot.py
```diff
@@ -106,12 +106,14 @@ def plot_embeddings(
         min_image_size: int = 16,
         channel_axis: int = -1,
         start_server: bool = True,
+        host: Optional[str] = "127.0.0.1",
```
hanxiao, line 109:
```suggestion
        host: Optional[str] = '127.0.0.1',
```

PR #51, opened 2022-01-15
refactor(array): use extend in add improve docs

hanxiao:


docarray/array/document.py
```diff
@@ -388,12 +388,10 @@ def __bool__(self):
...
 
-    def __add__(self, other: 'Document'):
+    def __add__(self, other: Union['Document', Sequence['Document']]):
         v = type(self)()
-        for doc in self:
-            v.append(doc)
-        for doc in other:
-            v.append(doc)
+        v.extend(self)
+        v.extend(other)
```
JoanFM, line 394: why not self.extend(other) and return self?

hanxiao: `__add__`
a  + b

not 

`__iadd__`
a+=b

PR #57, opened 2022-01-17
feat: use schema-ed json as default to_dict/json

hanxiao:
close #56


hanxiao:
do you know that `random_docs` have embeddings?

Issue #72, opened 2022-01-21
The misuse of smart union in pydantic


numb3r3:
The string value `"1"` in `doc.tags` will be automatically converted to bool `True` when using `.to_pydantic_model`. This would not make sense. 

The reproduce script:

```python
from docarray import Document

doc = Document(tags={'a': '1', 'b': 1, 'c': '2'})
print(f'origin:')
print(doc.tags)

pd_doc_tags = doc.to_pydantic_model().tags
print(f'pydantic: ')
print(pd_doc_tags)


proto_tags = doc.to_dict(protocol='protobuf')['tags']
print(f'protobuf: ')
print(proto_tags)


json_tags = doc.to_dict(protocol='jsonschema')['tags']
print(f'jsonschema')
print(proto_tags)
```

yields the following output:
```
origin:
{'a': '1', 'b': 1, 'c': '2'}
pydantic: 
{'a': True, 'b': True, 'c': 2.0}
protobuf: 
{'a': '1', 'c': '2', 'b': 1.0}
jsonschema
{'a': '1', 'c': '2', 'b': 1.0}
```

hanxiao:
This to be honest is an upstream issue, i don't think we can solve it?

numb3r3:
One possible quick solution is to disable the smart union in the pydantic model definition

```
class PydanticDocument(BaseModel):
    ...
    tags: Optional[Dict[str, '_StructValueType']] -> tags: Optional[Dict[str, ANY]]
```

https://github.com/jina-ai/docarray/blob/main/docarray/document/pydantic_model.py#L30-L50

But I'm not very clear whether this change will not destroy other API/functions.

numb3r3:
With this PR https://github.com/jina-ai/docarray/pull/73
```
origin:
{'a': '1', 'b': 1, 'c': '2'}
pydantic: 
{'a': '1', 'b': 1, 'c': '2'}
protobuf: 
{'b': 1.0, 'c': '2', 'a': '1'}
jsonschema
{'b': 1.0, 'c': '2', 'a': '1'}
```

PR #73, opened 2022-01-21
fix(pydantic): turn on pydantic smart union

numb3r3:
attempt to address https://github.com/jina-ai/docarray/issues/72

### DIFF
--- a/docarray/document/pydantic_model.py
+++ b/docarray/document/pydantic_model.py
@@ -38,7 +38,7 @@ class PydanticDocument(BaseModel):
     text: Optional[str]
     weight: Optional[float]
     uri: Optional[str]
-    tags: Optional[Dict[str, '_StructValueType']]
+    tags: Optional[Dict[str, Any]]
     offset: Optional[float]
     location: Optional[List[float]]
     embedding: Optional[Any]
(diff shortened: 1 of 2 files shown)


docarray/document/pydantic_model.py
```diff
@@ -38,7 +38,7 @@ class PydanticDocument(BaseModel):
     text: Optional[str]
     weight: Optional[float]
     uri: Optional[str]
-    tags: Optional[Dict[str, '_StructValueType']]
+    tags: Optional[Dict[str, Any]]
```
hanxiao, line 41: This however makes `tags` unschema

hanxiao: @JoanFM

JoanFM: I think is right having tags unschema, it is the free part of Document

hanxiao: but the "free-part" still have to be  a valid JSON

hanxiao: the right way is to turn on pydantic smart union, as in https://pydantic-docs.helpmanual.io/usage/model_config/#smart-union

hanxiao:
the original title of the PR is misleading, the default behavior is not "smart union" according to https://pydantic-docs.helpmanual.io/usage/model_config/#smart-union

This PR in fact turns on the "smart union".

hanxiao (approved):

PR #78, opened 2022-01-26
feat: storage option

winstonww:
To better keep track of the TODO items, I have compiled the TODOs from the code into the following summary:

- Investigate into get/set attribute support (e.g. `da[1, ['text', 'id']]`)
- Clarify, Investigate and refactor the current set attribute logic
- Add test coverage for `AllMixin`
- Combine DocumentArray and its meta object into one single schema
- Investigate and use/test base class' `_set_doc_value_pairs`; add test coverage
- Currently, `vectorIndexConfig` is disabled to allow docs with no embeddings to be added. But this needs to change as it also disables vector search functionality.  Will stay tuned and address this in the next step.
- Add match function

This PR incorporated some previous changes from @alaeddine-13  on test setup and also some initial suggestions from @numb3r3. Thanks for the feedback.

### DIFF
--- a/docarray/array/mixins/setitem.py
+++ b/docarray/array/mixins/setitem.py
@@ -72,6 +72,8 @@ class SetItemMixin:
             else:
                 self._set_doc_by_id(index, value)
         elif isinstance(index, slice):
+            # TODO: handle this case da[1:3] = DocumentAray.empty(10)
+            # where length of value is longer than the slice
             self._set_docs_by_slice(index, value)
         elif index is Ellipsis:
             self._set_doc_value_pairs(self.flatten(), value)
@@ -79,8 +81,52 @@ class SetItemMixin:
             if (
                 isinstance(index, tuple)
                 and len(index) == 2
-                and isinstance(index[0], (slice, Sequence))
+                and (
+                    isinstance(index[0], (slice, Sequence, str, int))
+                    or index[0] is Ellipsis
+                )
+                and isinstance(index[1], (str, Sequence))
             ):
+                # TODO: this is added because we are still trying to figure out the proper way
+                # to set attribute and to get test_path_syntax_indexing_set to pass.
+                # we may have to refactor the following logic
+                from ..weaviate import DocumentArrayWeaviate
+                from ..memory import DocumentArrayInMemory
+
+                # NOTE: this check is not proper way to handle, but a temporary hack.
+                # writing it this way to minimize effect on other docarray classs and
+                # to make it easier to remove/refactor the following block
+                if isinstance(self, (DocumentArrayWeaviate, DocumentArrayInMemory)):
+                    if index[1] in self:
+                        # we first handle he case when second item in index is an id not attr
+                        _docs = DocumentArrayInMemory(
+                            self[index[0]]
+                        ) + DocumentArrayInMemory(self[index[1]])
+                        self._set_doc_value_pairs(_docs, value)
+                        return
+
+                    _docs = self[index[0]]
+                    if isinstance(_docs, Document):
+                        _docs = DocumentArrayInMemory(_docs)
+                        # because we've augmented docs dimension, we do the same for value
+                        value = (value,)
+
+                    attrs = index[1]
+                    if isinstance(attrs, str):
+                        attrs = (attrs,)
+                        # because we've augmented attrs dimension, we do the same for value
+                        value = (value,)
+
+                    for attr in attrs:
+                        if any(not hasattr(_d, attr) for _d in _docs):
+                            raise ValueError(
+                                f'`{attr}` is neither a valid id nor attribute name'
+                            )
+
+                    for _a, _v in zip(attrs, value):
+                        self._set_docs_attrs(_docs, _a, _v)
+                    return
+
                 if isinstance(index[0], str) and isinstance(index[1], str):
                     # ambiguity only comes from the second string
                     if index[1] in self:
--- a/docarray/array/mixins/setitem.py
+++ b/docarray/array/mixins/setitem.py
@@ -109,10 +109,12 @@ class SetItemMixin:
                     if isinstance(_docs, Document):
                         _docs = DocumentArrayInMemory(_docs)
                         # because we've augmented docs dimension, we do the same for value
+                        print('ok')
                         value = (value,)
 
                     attrs = index[1]
                     if isinstance(attrs, str):
+                        print('okk')
                         attrs = (attrs,)
                         # because we've augmented attrs dimension, we do the same for value
                         value = (value,)
--- a/docarray/array/mixins/setitem.py
+++ b/docarray/array/mixins/setitem.py
@@ -72,8 +72,6 @@ class SetItemMixin:
             else:
                 self._set_doc_by_id(index, value)
         elif isinstance(index, slice):
-            # TODO: handle this case da[1:3] = DocumentAray.empty(10)
-            # where length of value is longer than the slice
             self._set_docs_by_slice(index, value)
         elif index is Ellipsis:
             self._set_doc_value_pairs(self.flatten(), value)
--- a/docarray/array/mixins/setitem.py
+++ b/docarray/array/mixins/setitem.py
@@ -106,21 +106,23 @@ class SetItemMixin:
                         return
 
                     _docs = self[index[0]]
+
+                    if not _docs:
+                        return
+
                     if isinstance(_docs, Document):
                         _docs = DocumentArrayInMemory(_docs)
                         # because we've augmented docs dimension, we do the same for value
-                        print('ok')
                         value = (value,)
 
                     attrs = index[1]
                     if isinstance(attrs, str):
-                        print('okk')
                         attrs = (attrs,)
                         # because we've augmented attrs dimension, we do the same for value
                         value = (value,)
 
                     for attr in attrs:
-                        if any(not hasattr(_d, attr) for _d in _docs):
+                        if not hasattr(_docs[0], attr):
                             raise ValueError(
                                 f'`{attr}` is neither a valid id nor attribute name'
                             )
--- a/docarray/array/mixins/setitem.py
+++ b/docarray/array/mixins/setitem.py
@@ -90,15 +90,18 @@ class SetItemMixin:
                 # TODO: this is added because we are still trying to figure out the proper way
                 # to set attribute and to get test_path_syntax_indexing_set to pass.
                 # we may have to refactor the following logic
-                from ..weaviate import DocumentArrayWeaviate
-                from ..memory import DocumentArrayInMemory
 
                 # NOTE: this check is not proper way to handle, but a temporary hack.
                 # writing it this way to minimize effect on other docarray classs and
                 # to make it easier to remove/refactor the following block
-                if isinstance(self, (DocumentArrayWeaviate, DocumentArrayInMemory)):
+                if self.__class__.__name__ in {
+                    'DocumentArrayWeaviate',
+                    'DocumentArrayInMemory',
+                }:
+                    from ..memory import DocumentArrayInMemory
+
                     if index[1] in self:
-                        # we first handle he case when second item in index is an id not attr
+                        # we first handle the case when second item in index is an id not attr
                         _docs = DocumentArrayInMemory(
                             self[index[0]]
                         ) + DocumentArrayInMemory(self[index[1]])
--- a/docarray/array/storage/base/getsetdel.py
+++ b/docarray/array/storage/base/getsetdel.py
@@ -36,20 +36,29 @@ class BaseGetSetDelMixin(ABC):
 
     def _get_docs_by_slice(self, _slice: slice) -> Iterable['Document']:
         """This function is derived from :meth:`_get_doc_by_offset`
+        Override this function if there is a more efficient logic
 
-        Override this function if there is a more efficient logic"""
+        :param _slice: the slice used for indexing
+        :return: an iterable of document
+        """
         return (self._get_doc_by_offset(o) for o in range(len(self))[_slice])
 
     def _get_docs_by_offsets(self, offsets: Sequence[int]) -> Iterable['Document']:
         """This function is derived from :meth:`_get_doc_by_offset`
+        Override this function if there is a more efficient logic
 
-        Override this function if there is a more efficient logic"""
+        :param offsets: the offsets used for indexing
+        :return: an iterable of document
+        """
         return (self._get_doc_by_offset(o) for o in offsets)
 
     def _get_docs_by_ids(self, ids: Sequence[str]) -> Iterable['Document']:
         """This function is derived from :meth:`_get_doc_by_id`
+        Override this function if there is a more efficient logic
 
-        Override this function if there is a more efficient logic"""
+        :param ids: the ids used for indexing
+        :return: an iterable of document
+        """
         return (self._get_doc_by_id(_id) for _id in ids)
 
     # Delitem APIs
@@ -64,15 +73,17 @@ class BaseGetSetDelMixin(ABC):
 
     def _del_docs_by_slice(self, _slice: slice):
         """This function is derived and may not have the most efficient implementation.
-
-        Override this function if there is a more efficient logic"""
+        Override this function if there is a more efficient logic
+        :param _slice: the slice used for indexing
+        """
         for j in range(len(self))[_slice]:
             self._del_doc_by_offset(j)
 
     def _del_docs_by_mask(self, mask: Sequence[bool]):
         """This function is derived and may not have the most efficient implementation.
-
-        Override this function if there is a more efficient logic"""
+        Override this function if there is a more efficient logic
+        :param mask: the boolean mask used for indexing
+        """
         for idx, m in enumerate(mask):
             if not m:
                 self._del_doc_by_offset(idx)
@@ -98,6 +109,9 @@ class BaseGetSetDelMixin(ABC):
         """This function is derived and may not have the most efficient implementation.
 
         Override this function if there is a more efficient logic
+        :param _slice: the slice used for indexing
+        :param value: the value docs will be updated to
+        :raises TypeError: error raised when right-hand assignment is not an iterable
         """
         if not isinstance(value, Iterable):
             raise TypeError(
@@ -112,12 +126,14 @@ class BaseGetSetDelMixin(ABC):
         """This function is derived and may not have the most efficient implementation.
 
         Override this function if there is a more efficient logic
+        :param docs: the docs to update
+        :param values: the value docs will be updated to
         """
         for _d, _v in zip(docs, values):
             _d._data = _v._data
 
         for _d in docs:
-            if _d not in docs:
+            if _d not in self:
                 root_d = self._find_root_doc(_d)
             else:
                 # _d is already on the root-level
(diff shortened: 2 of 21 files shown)


docarray/array/storage/weaviate/seqlike.py
```diff
@@ -0,0 +1,112 @@
...
+        """Add the other document to this :class:`DocumentArrayWeaviate`
+        :param other: the other document to add
+        :return: the new :class:`DocumentArrayWeaviate` but with other added
+        """
+        v = type(self)()
+        for doc in self:
+            v.append(doc)
+        for doc in other:
+            v.append(doc)
+        return v
```
alaeddine-13, line 101: I think this makes sense to DocumentArrayInMemory but not for storage backends.
Since storage DocumentArrays are tied to storage configurations, I don't think it should be possible to instantiate them implicitly (without config) like this.

winstonww: Yes, good point. For reasons you mentioned, I will remove both `__add__` for this and sqlite.

docarray/array/mixins/setitem.py
```diff
@@ -72,15 +72,61 @@ def __setitem__(
             else:
                 self._set_doc_by_id(index, value)
         elif isinstance(index, slice):
+            # TODO: handle this case da[1:3] = DocumentAray.empty(10)
+            # where length of value is longer than the slice
```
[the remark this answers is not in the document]
winstonww: I also think raising exception is easier. But I think since this operation is currently allowed in `DocumentArrayMemory`, correct me if I'm wrong but from the discussion we had since this is doable we should handle this case also in other storages? We can make handle this case a lower priority? @numb3r3  @hanxiao

docarray/array/mixins/setitem.py
```diff
@@ -72,15 +72,63 @@ def __setitem__(
...
+                    _docs = self[index[0]]
+                    if isinstance(_docs, Document):
+                        _docs = DocumentArrayInMemory(_docs)
+                        # because we've augmented docs dimension, we do the same for value
+                        print('ok')
+                        value = (value,)
+
+                    attrs = index[1]
+                    if isinstance(attrs, str):
+                        print('okk')
```
alaeddine-13, line 117:
```suggestion
```

docarray/array/mixins/setitem.py
```diff
@@ -72,15 +72,63 @@ def __setitem__(
...
+                            self[index[0]]
+                        ) + DocumentArrayInMemory(self[index[1]])
+                        self._set_doc_value_pairs(_docs, value)
+                        return
+
+                    _docs = self[index[0]]
+                    if isinstance(_docs, Document):
+                        _docs = DocumentArrayInMemory(_docs)
+                        # because we've augmented docs dimension, we do the same for value
+                        print('ok')
```
alaeddine-13, line 112:
```suggestion
```

docarray/array/mixins/setitem.py
```diff
@@ -72,15 +72,63 @@ def __setitem__(
...
+
+                    attrs = index[1]
+                    if isinstance(attrs, str):
+                        print('okk')
+                        attrs = (attrs,)
+                        # because we've augmented attrs dimension, we do the same for value
+                        value = (value,)
+
+                    for attr in attrs:
+                        if any(not hasattr(_d, attr) for _d in _docs):
```
alaeddine-13, line 123: Do we have to perform this check on all docs ?I believe the first one is enough ?

docarray/array/mixins/setitem.py
```diff
@@ -72,15 +72,63 @@ def __setitem__(
...
+                            raise ValueError(
+                                f'`{attr}` is neither a valid id nor attribute name'
+                            )
+
+                    for _a, _v in zip(attrs, value):
+                        self._set_docs_attrs(_docs, _a, _v)
```
numb3r3, line 131: This for loop is not efficient for storage backend. The iteration on the document level could be better.

```python
for _d in _docs:
    for _a, _v in zip(attrs, value):
        setattr(_d, _a, _v)
    self[_d.id] = _d
```

docarray/array/storage/base/getsetdel.py
```diff
@@ -112,12 +126,14 @@ def _set_doc_value_pairs(
...
         Override this function if there is a more efficient logic
+        :param docs: the docs to update
+        :param values: the value docs will be updated to
         """
         for _d, _v in zip(docs, values):
             _d._data = _v._data
 
         for _d in docs:
-            if _d not in docs:
+            if _d not in self:
```
numb3r3, line 136: BTW, this is a good catch. 👍

docarray/array/storage/sqlite/seqlike.py
```diff
@@ -97,8 +97,3 @@ def __eq__(self, other):
             and type(self._config) is type(other._config)
             and self._config == other._config
         )
-
-    def __add__(self, other: Union['Document', Sequence['Document']]):
-        v = type(self)(self, storage='sqlite')
-        v.extend(other)
-        return v
```
numb3r3, line 104: @alaeddine-13 @davidbp Is this revision is decided?

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+@dataclass
+class WeaviateConfig:
+    """This class stores the config variables to initialize
+    connection to the Weaviate server"""
+
+    client: Optional[Union[str, weaviate.Client]] = None
+    n_dim: Optional[int] = None
+    array_id: Optional[str] = None
```
numb3r3, line 36: `array_id` is not an explicit name to tell the user what's the meaning? And how to use it. How about renaming as `name`

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,290 @@
...
+    Dict,
+    Sequence,
+    Optional,
+    TYPE_CHECKING,
+    Union,
+    Tuple,
+    List,
+)
+
+import weaviate
```
numb3r3, line 16: move the `import weaviate` under `_init_storage()` to avoid the dependency missing problem.

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,290 @@
...
+                doc_schemas = self._get_schema_by_name(self._get_weaviate_class_name())
+            self._client.schema.create(doc_schemas)
+            return doc_schemas
+
+        doc_schemas = self._get_schema_by_name(cls_name)
+        if self._client.schema.contains(doc_schemas):
+            return doc_schemas
+
+        self._client.schema.create(doc_schemas)
+        return doc_schemas
```
numb3r3, line 162:
```suggestion
        doc_schemas = self._get_schema_by_name(cls_name or self._get_weaviate_class_name())
        if not self._client.schema.contains(doc_schemas):
            self._client.schema.create(doc_schemas)

        return doc_schemas
```

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,290 @@
...
+        if config.client is None:
+            self._client = weaviate.Client('http://localhost:8080')
+        elif isinstance(config.client, str):
+            self._client = weaviate.Client(config.client)
+        else:
+            self._client = config.client
+
+        if config.array_id is not None and docs is not None:
+            raise ValueError(
+                'only one of array_id or docs can be provided for initialization'
```
numb3r3, line 73: what's the reason for this constraint?

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,290 @@
...
+        return self._schemas['classes'][1]
+
+    def _doc2weaviate_create_payload(self, value: 'Document'):
+        """Return the payload to store :class:`Document` into weaviate
+
+        :param value: document to create a payload for
+        :return: the payload dictionary
+        """
+        if value.embedding is None:
+            embedding = [0]
```
numb3r3, line 267:
```suggestion
            embedding = [0] * config.n_dim
```

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,290 @@
...
+        elif isinstance(value.embedding, scipy.sparse.spmatrix):
+            embedding = value.embedding.toarray()
+        else:
+            embedding = value.embedding
+
```
numb3r3, line 272: to support `pytorch` tensor embedding

```suggestion
from ...math.ndarray import to_numpy_array
embedding = to_numpy_array(embedding)
```

docarray/array/storage/weaviate/getsetdel.py
```diff
@@ -0,0 +1,206 @@
...
+        return Document.from_base64(resp['properties']['_serialized'])
+
+    def _setitem(self, wid: str, value: Document):
+        """Helper method for setting an item with weaviate as storage
+
+        :param wid: weaviate id
+        :param value: the new document to update to
+        """
+        if self._client.data_object.exists(wid):
+            self._client.data_object.delete(wid)
```
numb3r3, line 36: I would suggest moving `delete` ops after the `self._doc2weaviate_create_payload`.

numb3r3 (changes requested):

docarray/array/mixins/setitem.py
winstonww: Removed.

docarray/array/mixins/setitem.py
```diff
@@ -72,15 +72,63 @@ def __setitem__(
...
+                and (
+                    isinstance(index[0], (slice, Sequence, str, int))
+                    or index[0] is Ellipsis
+                )
+                and isinstance(index[1], (str, Sequence))
             ):
+                # TODO: this is added because we are still trying to figure out the proper way
+                # to set attribute and to get test_path_syntax_indexing_set to pass.
+                # we may have to refactor the following logic
+                from ..weaviate import DocumentArrayWeaviate
```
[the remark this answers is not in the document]
winstonww: Agree `DocumentArrayWeaviate` should not be imported here. But this is done for now since I do not want to change the behavior of `DocumentArraySqlite`. I have also added comments to note this. If you have better alternative temporary solutions, please feel free to let me know.

docarray/array/storage/weaviate/backend.py
winstonww: Addressed.

docarray/array/storage/weaviate/getsetdel.py
winstonww: Addressed, thanks.

docarray/array/storage/weaviate/backend.py
winstonww: In the case when `cls_name` is not provided (i.e. when user did not specify the class name), as noted in the comment, I believe a new class should be created as we do not want to use previous/existing schema that may contain non-empty data?

docarray/array/storage/weaviate/backend.py
winstonww: hm I think if the given `array_id` has non-empty data, then it might be hard to decide what data this DocArray should be initialized to? I think if we extend `docs` to an already existing data given by `array_id` then we are not really initializing to either but performing a new `extend`?

docarray/array/storage/weaviate/backend.py
winstonww: Sounds good, I will rename it.

docarray/array/storage/weaviate/backend.py
winstonww: Will do, thanks.

docarray/array/mixins/setitem.py
winstonww: The difficulty with the suggested approach is that `_d` can be a non-root document. As a result, `self[_d.id] = ` currently will not set the expected Document. We may be able to come up with ways to alter all `_set_doc_by_x` to work on non-root documents, but this can be a very elaborate change and I think it may be easier to tackle this in next version as noted in the description/comment.

fyi, I have tried using the `_set_doc_value_pairs` approach with setattr. But also have issues with it because of the `flatten` method. This needs to be further investigated.

docarray/array/storage/weaviate/backend.py
winstonww: Thanks, addressed.

docarray/array/mixins/setitem.py
winstonww: Yes, one is sufficient.

docarray/array/mixins/setitem.py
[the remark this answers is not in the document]
numb3r3: @winstonww use `self.__class__.__name__ = 'DocumentArrayWeaviate'` in the if-condition

docarray/array/storage/weaviate/backend.py
numb3r3: Yes, I agree. But my suggestions do not change the logic.

docarray/array/storage/weaviate/backend.py
numb3r3: in this case, the given DocArray could be appended into the `array_id` collection?

docarray/array/storage/sqlite/seqlike.py
alaeddine-13: For me `__add__` only makes sense for memory document array.

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,294 @@
...
+        self._schemas = self._load_or_create_weaviate_schema(config.name)
+        self._offset2ids, self._offset2ids_wid = self._get_offset2ids_meta()
+
+        if docs is None or config.name:
+            return
+
+        elif isinstance(
+            docs, (DocumentArray, Sequence, Generator, Iterator, itertools.chain)
+        ):
+            self.extend(Document(d, copy=True) for d in docs)
```
alaeddine-13, line 87: Is it needed to do this here ?
Can't we just do `self.extend(docs)` ?

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,294 @@
...
+                    'class': cls_name + 'Meta',
+                    "vectorizer": "none",
+                    'vectorIndexConfig': {'skip': True},
+                    'properties': [
+                        {
+                            'dataType': ['string[]'],
+                            'name': '_offset2ids',
+                            'indexInverted': False,
+                        },
+                    ],
```
alaeddine-13, line 137: Does it make sense to create a property of `string` instead where the whole column consists a mapping between offsets to the document IDs. Then, we no more need to load offset2ids in memory and just need to keep it in the database.
We just need to convert offsets to uuids when we want to do the mapping (maybe something like `uuid.uuid5(uuid.NAMESPACE_URL, str(offset))`)

docarray/array/storage/weaviate/backend.py
numb3r3: ah, you are right. forgot my code.

docarray/array/storage/weaviate/backend.py
winstonww: I am not aware of a reason to do this,  thus have changed this to `self.extend(docs)`.

docarray/array/storage/weaviate/backend.py
winstonww: As discussed offline, will align with Sqlite behavior but will also raise this for more feedback

docarray/array/storage/weaviate/backend.py
winstonww: Do you mean creating a new field of type `string` in the `Document` schema (i.e. schema with name `_class_name`, not `_meta_name`) that maps offset -> document ID? If so, yes combining the two schemas into one is the plan going forward but will be tackled in next phase (as detailed in google docs/description). 

Currently, what blocks us from doing that is that unlike SQLite where we can update the entire column with a single SQL statement, in Weaviate's case I don't think it's possible to update an entire column (or batch update every property of objects in a class) atm (their Python client only allows adding objects in batch, not updating/deleting). 

If I do not understand you correctly, would you please provide an example (e.g. `daw.insert(3, doc)` or `del daw[4]`)  to illustrate what would happen in your suggested case? Thank you.

alaeddine-13: Basicly I meant something like:
```
                {
                    'class': cls_name + 'Meta',
                    "vectorizer": "none",
                    'vectorIndexConfig': {'skip': True},
                    'properties': [
                        {
                            'dataType': ['string[]'],
                            'name': '_offset2ids',
                            'indexInverted': False,
                        },
                    ]
                }
```
and implementing `def _offset2id(self, offset)` so that it converts offset to a weaviate id first (like in wmap) and then retrieves to id associated with offset.
But now I get your point, weaviate doesn't support batch operations and this will be problematic in some cases.

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,298 @@
...
+            .do()
+            .get('data', {})
+            .get('Get', {})
+            .get(self._meta_name, [])
+        )
+
+        if not resp:
+            return [], None
+        elif len(resp) == 1:
+            return resp[0]['_offset2ids'], resp[0]['_additional']['id']
```
alaeddine-13, line 211:
```suggestion
            return resp[0].get('_offset2ids', []), resp[0].get('_additional', {}).get('id')
```

docarray/array/storage/weaviate/backend.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+@dataclass
+class WeaviateConfig:
+    """This class stores the config variables to initialize
+    connection to the Weaviate server"""
+
+    client: Optional[Union[str, weaviate.Client]] = None
+    n_dim: Optional[int] = None
+    name: Optional[str] = None
```
alaeddine-13, line 36: In SqliteConfig, there's a parameter `serialize_config` and currently the serialization protocol is being changed to protobuf.
Let's add the same parameter here or create a `BaseConfig` class that has it and make both Config classes inherit from it

docarray/array/storage/weaviate/backend.py
winstonww: Since the length of `resp` is 1, at this point we should expect `_offset2ids` and `_additional__id` to be present in the `dict`, hence I think an error should be raised if these fields are not present like in the current case. More importantly, it would also be easier to debug if we error out early as opposed to going with the suggested approach of accepting invalid values which may have potential of causing issues later down the execution.

docarray/array/storage/weaviate/backend.py
winstonww: Sure, `serialize_config` is made a parameter for now because different storages may have different serialization protocol by default (I'm using `pickle` in this case because currently `protobuf` is having issues with `to_base64`). While it's possible to also address this with `BaseConfig`, there may also be considerations with ordering kw/non-kw arguments and stuff. To make it simple and not to over-engineer, I just made it a parameter.

docarray/array/mixins/setitem.py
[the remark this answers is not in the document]
winstonww: Sure, I will remove the TODO and if such corner case comes up we can always open another issue if needed.

numb3r3 (approved):
LGTM

PR #79, opened 2022-01-26
fix: del with ids

samsja:
Current version of docarray has a wrong behavior while deleting item via ids ==>

@JohannesMessner and I added some test to cover all of the del case from the documentation + a fix when deleting by ids

### DIFF
--- a/docarray/array/storage/memory/backend.py
+++ b/docarray/array/storage/memory/backend.py
@@ -44,7 +44,7 @@ class BackendMixin(BaseBackendMixin):
         self._needs_id2offset_rebuild = False
 
     def _init_storage(
-        self, _docs: Optional["DocumentArraySourceType"] = None, copy: bool = False
+        self, _docs: Optional['DocumentArraySourceType'] = None, copy: bool = False
     ):
         from ... import DocumentArray
 
--- a/docarray/array/storage/memory/backend.py
+++ b/docarray/array/storage/memory/backend.py
@@ -26,8 +26,9 @@ class BackendMixin(BaseBackendMixin):
 
         :return: a Python dict.
         """
-        if not hasattr(self, '_id_to_index'):
+        if self._needs_id2offset_rebuild:
             self._rebuild_id2offset()
+
         return self._id_to_index
 
     def _rebuild_id2offset(self) -> None:
@@ -40,10 +41,16 @@ class BackendMixin(BaseBackendMixin):
             d.id: i for i, d in enumerate(self._data)
         }  # type: Dict[str, int]
 
+        self._needs_id2offset_rebuild = False
+
     def _init_storage(
-        self, _docs: Optional['DocumentArraySourceType'] = None, copy: bool = False
+        self, _docs: Optional["DocumentArraySourceType"] = None, copy: bool = False
     ):
         from ... import DocumentArray
+
+        self._needs_id2offset_rebuild = True
+        # self._id2offset needs to be rebuilt after every insert or delete
+        # this flag allows to do it lazily and cache the result
 
         self._data = []
         if _docs is None:
@@ -53,13 +60,11 @@ class BackendMixin(BaseBackendMixin):
         ):
             if copy:
                 self._data = [Document(d, copy=True) for d in _docs]
-                self._rebuild_id2offset()
             elif isinstance(_docs, DocumentArray):
                 self._data = _docs._data
                 self._id_to_index = _docs._id2offset
             else:
                 self._data = list(_docs)
-                self._rebuild_id2offset()
         else:
             if isinstance(_docs, Document):
                 if copy:
--- a/docarray/array/storage/memory/backend.py
+++ b/docarray/array/storage/memory/backend.py
@@ -63,6 +63,7 @@ class BackendMixin(BaseBackendMixin):
         from ... import DocumentArray
 
         self._data = []
+        self._id_to_index = {}
         if _docs is None:
             return
         elif isinstance(
--- a/docarray/array/storage/memory/backend.py
+++ b/docarray/array/storage/memory/backend.py
@@ -74,11 +74,11 @@ class BackendMixin(BaseBackendMixin):
                 self._data = [Document(d, copy=True) for d in _docs]
             elif isinstance(_docs, DocumentArray):
                 self._data = _docs._data
-                self._id_to_index = _docs._id2offset
             else:
                 self._data = list(_docs)
 
             if isinstance(_docs, DocumentArrayInMemory):
+                self._id_to_index = _docs._id2offset
                 self._needs_id2offset_rebuild = _docs._needs_id2offset_rebuild
         else:
             if isinstance(_docs, Document):
--- a/docarray/array/storage/memory/backend.py
+++ b/docarray/array/storage/memory/backend.py
@@ -61,6 +61,7 @@ class BackendMixin(BaseBackendMixin):
         self, _docs: Optional['DocumentArraySourceType'] = None, copy: bool = False
     ):
         from ... import DocumentArray
+        from ...memory import DocumentArrayInMemory
 
         self._data = []
         self._id_to_index = {}
@@ -76,6 +77,9 @@ class BackendMixin(BaseBackendMixin):
                 self._id_to_index = _docs._id2offset
             else:
                 self._data = list(_docs)
+
+            if isinstance(_docs, DocumentArrayInMemory):
+                self._needs_id2offset_rebuild = _docs._needs_id2offset_rebuild
         else:
             if isinstance(_docs, Document):
                 if copy:
--- a/docarray/array/storage/memory/backend.py
+++ b/docarray/array/storage/memory/backend.py
@@ -6,7 +6,9 @@ from typing import (
     Sequence,
     Optional,
     TYPE_CHECKING,
+    Callable,
 )
+import functools
 
 from ..base.backend import BaseBackendMixin
 from .... import Document
@@ -15,6 +17,17 @@ if TYPE_CHECKING:
     from ....types import (
         DocumentArraySourceType,
     )
+
+
+def needs_id2offset_rebuild(func) -> Callable:
+    # self._id2offset needs to be rebuilt after every insert or delete
+    # this flag allows to do it lazily and cache the result
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        self._needs_id2offset_rebuild = True
+        return func(self, *args, **kwargs)
+
+    return wrapper
 
 
 class BackendMixin(BaseBackendMixin):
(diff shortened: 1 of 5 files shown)


docarray/array/storage/memory/getsetdel.py
```diff
@@ -26,7 +26,7 @@ def _del_docs_by_slice(self, _slice: slice):
 
     def _del_doc_by_id(self, _id: str):
         del self._data[self._id2offset[_id]]
-        self._id2offset.pop(_id)
```
hanxiao, line 29: calling `_rebuild_id2offset` is expensive. in general every inconsistency problem can be solved by calling this method, but we choose to avoid that for reducing the overhead.

What is exact problem of the current `_id2offset.pop` logic?

samsja: Line 28 when we remove the Document from the data the indices from all of the following document shift by one which is not reflected by self._id2offset.pop. Technically instead of rebuilding the offset we could decrement the indices of only the documents that are after the one that just got deleted. But unless we are missing something, that wouldn't be a big gain (still O(n)).5

samsja: Currently that is what happening :

```python
da = DocumentArray([Document(id=f"doc{i}") for i in range(4)])
del da["doc0"]
del da["doc1"]
print(da[:,"id"]) >>> ['doc1', 'doc3']
```

when calling del da["doc1"] it is the "doc2" which is deleted because the id2offset did not shift the indices

hanxiao: you are right, but this suggest `insert` logic is also wrong

hanxiao: `docarray/array/storage/memory/seqlike.py:L16`

samsja: Yep you are right, it is fix now in the last commit

docarray/array/storage/memory/getsetdel.py
```diff
@@ -14,23 +14,22 @@ class GetSetDelMixin(BaseGetSetDelMixin):
...
+        self._needs_id2offset_rebuild = True
 
     def _del_all_docs(self):
         self._data.clear()
         self._id2offset.clear()
 
     def _del_docs_by_slice(self, _slice: slice):
         del self._data[_slice]
-        self._rebuild_id2offset()
+        self._needs_id2offset_rebuild = True
```
numb3r3, line 25: change the `_needs_id2offset_rebuild` to a decorator?

docarray/array/storage/memory/backend.py
```diff
@@ -40,11 +41,17 @@ def _rebuild_id2offset(self) -> None:
             d.id: i for i, d in enumerate(self._data)
         }  # type: Dict[str, int]
 
+        self._needs_id2offset_rebuild = False
+
     def _init_storage(
-        self, _docs: Optional['DocumentArraySourceType'] = None, copy: bool = False
+        self, _docs: Optional["DocumentArraySourceType"] = None, copy: bool = False
```
hanxiao, line 47: single quote

tests/unit/array/mixins/test_del.py
```diff
@@ -0,0 +1,42 @@
...
+    assert doc_to_delete not in docs
+
+
+@pytest.mark.parametrize(
+    ["deleted_ids", "expected_ids"],
+    [
+        (["1", "2", "3", "4"], ["5", "6", "7", "8", "9"]),
+        (["2", "4", "7", "1"], ["3", "5", "6", "8", "9"]),
+    ],
+)
```
hanxiao, line 39: - use `black -s`
- or a better way is to install pre-commit hook via 
```python
# at docarray root
pre-commit install
```

you will see this option auto-checked on every commit
image

samsja: I already installed the pre-commit hook, this is already formatted by black

docarray/array/storage/memory/getsetdel.py
samsja: I changed it in the last commit, could you review ?

docarray/array/storage/memory/getsetdel.py
```diff
@@ -41,16 +40,16 @@ def _set_doc_by_id(self, _id: str, value: 'Document'):
...
 
+    @needs_id2offset_rebuild
     def _set_docs_by_slice(self, _slice: slice, value: Sequence['Document']):
         self._data[_slice] = value
-        self._rebuild_id2offset()
 
+    @needs_id2offset_rebuild
     def _set_doc_value_pairs(
         self, docs: Iterable['Document'], values: Iterable['Document']
     ):
```
alaeddine-13, line 50: why does this need rebuilding the offset ?

docarray/array/storage/memory/seqlike.py
```diff
@@ -55,6 +57,6 @@ def __add__(self, other: Union['Document', Sequence['Document']]):
         v.extend(other)
         return v
 
+    @needs_id2offset_rebuild
     def extend(self, values: Iterable['Document']) -> None:
         self._data.extend(values)
-        self._rebuild_id2offset()
```
alaeddine-13, line 60: I think extending doesn't require rebuilding because we just need to add more offsets

alaeddine-13 (commented):
In general, I think that the operations that require rebuilding the id2offset are the delete and insert. Extend and append really don't require doing so. If we append/extend and the id2offset, is ready, then we update it efficiently without rebuilding, otherwise we don't update and wait for the next `rebuilding`

docarray/array/storage/memory/seqlike.py
samsja: Done in the last commit

docarray/array/storage/memory/backend.py
```diff
@@ -40,26 +54,28 @@ def _rebuild_id2offset(self) -> None:
...
             _docs, (DocumentArray, Sequence, Generator, Iterator, itertools.chain)
         ):
             if copy:
                 self._data = [Document(d, copy=True) for d in _docs]
-                self._rebuild_id2offset()
             elif isinstance(_docs, DocumentArray):
                 self._data = _docs._data
                 self._id_to_index = _docs._id2offset
             else:
                 self._data = list(_docs)
```
alaeddine-13, line 78: I'm not sure but does it make sense to copy the `self._needs_id2offset_rebuild` at this point ? (but only if it's also a `DocumentArrayInMemory`)
So maybe we can do:
```suggestion
            elif isinstance(_docs, DocumentArray):
                self._data = _docs._data
                self._id_to_index = _docs._id2offset
            else:
                self._data = list(_docs)
            
            if isinstance(_docs, DocumentArrayInMemory):
                self._needs_id2offset_rebuild = _docs._needs_id2offset_rebuild
```

docarray/array/storage/memory/seqlike.py
```diff
@@ -57,4 +59,6 @@ def __add__(self, other: Union['Document', Sequence['Document']]):
 
     def extend(self, values: Iterable['Document']) -> None:
         self._data.extend(values)
-        self._rebuild_id2offset()
+        # extend _id_to_index, no rebuilt needed
+        last_idx = len(self._id_to_index)
+        self._id_to_index.update({d.id: i + last_idx for i, d in enumerate(values)})
```
alaeddine-13, line 64: Let's also handle the case where the `_id_to_index` is not rebuilt.
I guess this should take care of it:
```suggestion
        # extend _id_to_index, no rebuilt needed
        last_idx = len(self._id2offset)
        self._id_to_index.update({d.id: i + last_idx for i, d in enumerate(values)})
```

samsja: completely agree

docarray/array/storage/memory/getsetdel.py
samsja: no need, it is fixed

alaeddine-13 (approved):
LGTM

docarray/array/storage/memory/backend.py
```diff
@@ -40,26 +54,32 @@ def _rebuild_id2offset(self) -> None:
...
             return
         elif isinstance(
             _docs, (DocumentArray, Sequence, Generator, Iterator, itertools.chain)
         ):
             if copy:
                 self._data = [Document(d, copy=True) for d in _docs]
-                self._rebuild_id2offset()
             elif isinstance(_docs, DocumentArray):
                 self._data = _docs._data
                 self._id_to_index = _docs._id2offset
```
hanxiao, line 77: i think this part is wrong, not all `DocumentArray` has `_id_to_index` or `_id2offset`

docarray/array/storage/memory/backend.py
```diff
@@ -40,26 +54,32 @@ def _rebuild_id2offset(self) -> None:
...
+        self._id_to_index = {}
         if _docs is None:
             return
         elif isinstance(
             _docs, (DocumentArray, Sequence, Generator, Iterator, itertools.chain)
         ):
             if copy:
                 self._data = [Document(d, copy=True) for d in _docs]
-                self._rebuild_id2offset()
             elif isinstance(_docs, DocumentArray):
```
hanxiao, line 75:
```suggestion
            elif isinstance(_docs, DocumentArrayinMemory):
```

hanxiao (dismissed):
wrong logic in copy constructor

docarray/array/storage/memory/seqlike.py
```diff
@@ -56,5 +58,7 @@ def __add__(self, other: Union['Document', Sequence['Document']]):
         return v
 
     def extend(self, values: Iterable['Document']) -> None:
-        self._data.extend(values)
-        self._rebuild_id2offset()
+        last_idx = len(self._id2offset)
+        for i, d in enumerate(values):
+            self._id_to_index[d.id] = last_idx + i
+            self._data.append(d)
```
hanxiao, line 64: this looks slower than the old version. If the intention to play safe, then simply delete `extend` method completely, Python will automatically delegate back to using `for-loop append`. That is, the only purpose to implement `extend` is to have faster method, if one can not come up with one, then there is no need to implement that

docarray/array/storage/memory/backend.py
samsja: well either the DocumentArray inherit from BackendMixin and has `_id_to_index` or `_id2offset` (because it is defined in the this class), either they don't inherit and they don't call this method. So in both case everything is fine ? Might be missing something

samsja: I agree, that was here before so I might be an other bug ? we should delete line 77 as the index _id_to_index will be automatically built on first call.

docarray/array/storage/memory/backend.py
samsja: fix in last commit

docarray/array/storage/memory/backend.py
samsja: fix in last commit

PR #81, opened 2022-01-26
test(sqlite): add more test to cover sqlite backend

hanxiao:


tests/unit/array/mixins/test_empty.py
```diff
@@ -1,7 +1,11 @@
+import pytest
+
 from docarray import DocumentArray
+from docarray.array.sqlite import DocumentArraySqlite
 
 
-def test_empty_non_zero():
+@pytest.mark.parametrize('da_cls', [DocumentArray, DocumentArraySqlite])
+def test_empty_non_zero(da_cls):
```
alaeddine-13, line 8: although empty is possible to `DocumentArraySqlite` (basicly because we allow config-free construction via temp files), I don't think it should be supported for storage backends in general (and no backend in particular).
If we think about it, where does `DocumentArrayWeaviate.empty(10)` store ? Should docs go to the default `localhost:8080` server configuration ?

hanxiao: ref1: https://github.com/RaRe-Technologies/sqlitedict/blob/master/sqlitedict.py
ref2: https://github.com/osoken/sqlitecollections

they are support default connection as temp file, so do we. let's follow this behavior

tests/unit/array/mixins/test_embed.py
```diff
@@ -40,10 +43,11 @@
 
 
 @pytest.mark.parametrize('framework', ['onnx', 'keras', 'pytorch', 'paddle'])
-@pytest.mark.parametrize('da', [DocumentArray])
+# @pytest.mark.parametrize('da', [DocumentArray, DocumentArraySqlite])
+@pytest.mark.parametrize('da', [DocumentArraySqlite])
```
alaeddine-13, line 47:
```suggestion
@pytest.mark.parametrize('da', [DocumentArray, DocumentArraySqlite])
```

tests/unit/array/mixins/test_embed.py
```diff
@@ -40,10 +43,11 @@
...
 @pytest.mark.parametrize('N', [2, 1000])
 @pytest.mark.parametrize('batch_size', [1, 256])
-@pytest.mark.parametrize('to_numpy', [True, False])
+@pytest.mark.parametrize('to_numpy', [True])
```
alaeddine-13, line 50:
```suggestion
@pytest.mark.parametrize('to_numpy', [True, False])
```
just so that we don't forget

docarray/array/mixins/group.py
```diff
@@ -64,3 +64,57 @@ def batch(
 
         for i in range(n_batches):
             yield self[ix[i * batch_size : (i + 1) * batch_size]]
+
+    def batch_indices(
+        self,
+        batch_size: int,
+        shuffle: bool = False,
+    ) -> Generator[list, None, None]:
```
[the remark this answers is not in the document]
davidbp: It's not, it can be removed or kept as an alternative to`batch_ids` since docarray suports both indexing by ids and indexing by indices. Probably for in_memory it's faster to do it by index and in sqlite to do it by ids.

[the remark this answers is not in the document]
alaeddine-13: removed, we can restore it later if needed

winstonww (approved):
As #78 may introduce a few minor changes in the expected test behavior,  I have discussed with @alaeddine-13 before merging #78 and we have agreed upon them. 

Otherwise, LGTM!

docarray/array/mixins/getitem.py
```diff
@@ -92,6 +92,11 @@ def __getitem__(
                         _attrs = (index[1],)
                     return _docs._get_attributes(*_attrs)
             elif isinstance(index[0], bool):
+                if len(index) != len(self):
+                    raise IndexError(
+                        f'Boolean mask index is required to have the same length as {len(self)}, '
+                        f'but receiving {len(index)}'
+                    )
```
hanxiao, line 99:
```suggestion

```

docarray/array/mixins/group.py
```diff
@@ -64,3 +64,30 @@ def batch(
...
+        """
+        Creates a `Generator` that yields `lists of ids` of size `batch_size` until `docs` is fully traversed along
+        the `traversal_path`.  Note, that the last batch might be smaller than `batch_size`.
+
+        :param batch_size: Size of each generated batch (except the last one, which might be smaller, default: 32)
+        :param shuffle: If set, shuffle the Documents before dividing into minibatches.
+        :yield: a Generator of `np.ndarray`, each in the length of `batch_size`
```
hanxiao, line 79: has to be revised

docarray/array/mixins/group.py
```diff
@@ -64,3 +64,30 @@ def batch(
 
         for i in range(n_batches):
             yield self[ix[i * batch_size : (i + 1) * batch_size]]
+
+    def batch_ids(
+        self,
+        batch_size: int,
+        shuffle: bool = False,
+    ) -> Generator[list, None, None]:
```
hanxiao, line 72:
```suggestion
    ) -> Generator[List[int], None, None]:
```

docarray/array/mixins/group.py
```diff
@@ -64,3 +64,30 @@ def batch(
...
         for i in range(n_batches):
             yield self[ix[i * batch_size : (i + 1) * batch_size]]
+
+    def batch_ids(
+        self,
+        batch_size: int,
+        shuffle: bool = False,
+    ) -> Generator[List[str], None, None]:
+        """
+        Creates a `Generator` that yields `lists of ids` of size `batch_size` until `self` is fully .
```
samsja, line 74: small typo
```suggestion
        Creates a `Generator` that yields `lists of ids` of size `batch_size` until `self` is empty .
```

docarray/array/mixins/io/common.py
```diff
@@ -1,4 +1,4 @@
-from typing import Union, TextIO, BinaryIO, TYPE_CHECKING, Type
+from typing import Union, TextIO, BinaryIO, TYPE_CHECKING, Type, Optional
```
samsja, line 1: Optional is not used
```suggestion
from typing import Union, TextIO, BinaryIO, TYPE_CHECKING, Type
```

docarray/array/storage/sqlite/binary.py
```diff
@@ -0,0 +1,134 @@
...
+
+        :param file: File or filename or serialized bytes where the data is stored.
+        :param protocol: protocol to use. 'pickle-array' is not supported for DocumentArraySqlite
+        :param compress: compress algorithm to use
+        :param _show_progress: show progress bar, only works when protocol is `pickle` or `protobuf`
+        :param streaming: if `True` returns a generator over `Document` objects.
+        In case protocol is pickle the `Documents` are streamed from disk to save memory usage
+        :return: a DocumentArray object
+        """
+        _check_protocol(protocol)
```
hanxiao, line 38: i think a better way to do this is the following:
1. moving `_check_protocol(protocol)` to `DocumentArray` level mixin
2. add class variable `available_protocols `, e.g.
```python
class IOMixin:

   available_protocols = ['a', 'b', 'c', 'd']
   def _check_protocols(self, value: str):
       if value not in self.available_protocols:
          raise ...

   def from_bytes():
      self._check_protocols

```
3. Finally, all DA subclass just need to specify a white list by overriding `available_protocols` class variable

hanxiao (commented):
implement `available_protocols` on the top-level

docarray/array/mixins/setitem.py
```diff
@@ -76,127 +76,16 @@ def __setitem__(
...
-                        f'the same as the index length: {len(index)}'
-                    )
-                if isinstance(value, Document):
-                    for si in index:
-                        self[si] = value  # leverage existing setter
-                else:
-                    for si, _val in zip(index, value):
-                        self[si] = _val  # leverage existing setter
+                for si, _val in zip(index, value):
+                    self[si] = _val  # leverage existing setter
```
samsja, line 87: Invalid types in the sequence should be caught and not fail silently; particularly things like da[1.0, 2.0] coming from np or torch might be a pitfall.

```suggestion
            else:
                raise IndexError(f"{index} should be either a sequence of bool, int or str")
```

docarray/array/storage/sqlite/binary.py
alaeddine-13: @hanxiao I'll still need to subclass the mix and override methods like `load_binary` because I need to change the default protocol. Is it okay to implement your suggestion while still overriding the methods ?

There's another option which is adding attribute `default_protocol` in the same way, and modifying the protocol parameter to be None (will bad for autocompletion).
Your thoughts ?

docarray/array/mixins/setitem.py
```diff
@@ -207,3 +96,104 @@ def __setitem__(
...
+            # second is a list of attributes:
+            elif (
+                isinstance(idx2, Sequence)
+                and all(isinstance(attr, str) for attr in idx2)
+                and all(hasattr(self[idx1], attr) for attr in idx2)
+            ):
+                for attr, _v in zip(idx2, value):
+                    self._set_doc_attr_by_id(idx1, attr, _v)
+            else:
+                raise ValueError(f'`{idx2}` is neither a valid id nor attribute name')
```
samsja, line 117:
```suggestion
                raise IndexError(f'`{idx2}` is neither a valid id nor attribute name')
```

docarray/array/mixins/setitem.py
```diff
@@ -207,3 +96,104 @@ def __setitem__(
...
+            # second is a list of attributes:
+            elif (
+                isinstance(idx2, Sequence)
+                and all(isinstance(attr, str) for attr in idx2)
+                and all(hasattr(self[idx1], attr) for attr in idx2)
+            ):
+                for attr, _v in zip(idx2, value):
+                    self._set_doc_attr_by_id(idx1, attr, _v)
+            else:
+                raise ValueError(f'`{idx2}` must be an attribute or list of attributes')
```
samsja, line 134:
```suggestion
                raise IndexError(f'`{idx2}` must be an attribute or list of attributes')
```

docarray/array/mixins/setitem.py
```diff
@@ -207,3 +96,104 @@ def __setitem__(
...
+                return
+
+            for _a, _v in zip(attributes, value):
+                if _a in ('tensor', 'embedding'):
+                    if _a == 'tensor':
+                        _docs.tensors = _v
+                    elif _a == 'embedding':
+                        _docs.embeddings = _v
+                    for _d in _docs:
+                        self._set_doc_by_id(_d.id, _d)
```
samsja, line 170:
```suggestion
                if _a == 'tensor':
                    _docs.tensors = _v
                elif _a == 'embedding':
                    _docs.embeddings = _v
                    for _d in _docs:
                        self._set_doc_by_id(_d.id, _d)
```

docarray/array/mixins/setitem.py
```diff
@@ -207,3 +96,104 @@ def __setitem__(
...
+        _docs = self[traversal_paths]
+        if not _docs:
+            return
+
+        for _a, _v in zip(attributes, value):
+            if _a in ('tensor', 'embedding'):
+                if _a == 'tensor':
+                    _docs.tensors = _v
+                elif _a == 'embedding':
+                    _docs.embeddings = _v
```
samsja, line 191:
```suggestion
            if _a == 'tensor':
                _docs.tensors = _v
            elif _a == 'embedding':
                _docs.embeddings = _v
```

docarray/array/mixins/setitem.py
samsja: oups not used to it yet :confused:

docarray/math/ndarray.py
```diff
@@ -70,19 +68,17 @@ def ravel(value: 'ArrayType', docs: Sequence['Document'], field: str) -> None:
 
     if use_get_row:
         emb_shape0 = value.shape[0]
-        for d, j in zip(docs, range(emb_shape0)):
+        for i, (d, j) in enumerate(zip(docs, range(emb_shape0))):
```
samsja, line 71: Why is it needed ? variable i seems to be never used

docarray/math/ndarray.py
```diff
@@ -70,19 +68,17 @@ def ravel(value: 'ArrayType', docs: Sequence['Document'], field: str) -> None:
...
-            setattr(d, field, row)
+            docs[d.id, field] = row
     elif isinstance(value, (list, tuple)):
         for d, j in zip(docs, value):
-            setattr(d, field, j)
+            docs[d.id, field] = j
     else:
 
         emb_shape0 = value.shape[0]
         for i, (d, j) in enumerate(zip(docs, range(emb_shape0))):
```
samsja, line 80: Same, why is it needed ? variable i seems to be never used

docarray/array/mixins/setitem.py
alaeddine-13:
```
for _d in _docs:
    self._set_doc_by_id(_d.id, _d)
```
should be common to both branches

docarray/array/mixins/setitem.py
alaeddine-13: fixed

PR #82, opened 2022-01-26
refactor(deps): move cicd requirement to extra_requires

hanxiao:


davidbp:
LGTM, I think it's convenient, thanks!

PR #84, opened 2022-01-26
fix(hashing): ignore casting to float

### LINKED ISSUE #83: Feature hashing fails with float strings `inf` and `nan`
To produce:

```python
>>> from docarray import Document
>>> x = Document(text="to infinity and beyond")
>>> x.embed_feature_hashing()
  File "<stdin>", line 1, in <module>
  File "~/docarray-env/docarray/docarray/document/mixins/featurehash.py", line 40, in embed_feature_hashing
    _hash_column(f_id, val, n_dim, max_value, idxs, data, table)
  File "~/docarray-env/docarray/docarray/document/mixins/featurehash.py", line 63, in _hash_column
    table[col] += np.sign(h) * col_val
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

abduhbm:
Fixes #83 

```python
>>> x = Document(text="float test 2.56")
>>> x.get_vocabulary()
Counter({'float': 1, 'test': 1, '2': 1, '56': 1})
```

### DIFF
--- a/docarray/document/mixins/featurehash.py
+++ b/docarray/document/mixins/featurehash.py
@@ -67,19 +67,16 @@ def _any_hash(v):
     try:
         return int(v)  # parse int parameter
     except ValueError:
-        try:
-            return float(v)  # parse float parameter
-        except ValueError:
-            if not v:
-                # ignore it when the parameter is empty
+        if not v:
+            # ignore it when the parameter is empty
+            return 0
+        if isinstance(v, str):
+            v = v.strip()
+            if v.lower() in {'true', 'yes'}:  # parse boolean parameter
+                return 1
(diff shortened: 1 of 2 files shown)


docarray/document/mixins/featurehash.py
```diff
@@ -67,19 +67,16 @@ def _any_hash(v):
     try:
         return int(v)  # parse int parameter
     except ValueError:
-        try:
-            return float(v)  # parse float parameter
```
numb3r3, line 71: We can also check the parsed value, and raise exception then
```
v = float(v)
if math.isnan(v) or math.isinf(v):
    raise ValueError
```

PR #91, opened 2022-02-03
feat: allow empty to get kwargs

davidbp:
This PR allows `.empty` to work with storage backends that need "configuration information".

Consider the following setting

```python
import numpy as np

n_docs = 500
n_features = 128
```

A user can do 

```python
# works in sqlite
storage='sqlite'
da = DocumentArray(storage=storage).empty(n_docs)
da[:, 'embedding'] = [x for x in np.random.random((n_docs,n_features))]
assert da[:,'embedding'].shape == (n_docs, n_features)
da.embeddings = np.random.random((n_docs,n_features))
assert da[:,'embedding'].shape == (n_docs, n_features)
```

but can't do 
```python
storage='pqlite'
da_pqlite = DocumentArray(storage=storage, config={'n_dim': n_features}).empty(n_docs)
da_pqlite[:, 'embedding'] = [x for x in np.random.random((len(da_pqlite),n_features))]
assert da_pqlite[:,'embedding'].shape == (500, 128)
da_pqlite.embeddings = np.random.random((len(da_pqlite),n_features))
assert da_pqlite[:,'embedding'].shape == (500, 128)
```

Because the information from `config` is ignored with `.empty` is called.

With this PR users can do

```python
storage='pqlite'
da_pqlite = DocumentArray(storage=storage).empty(n_docs, config={'n_dim': n_features})
da_pqlite[:, 'embedding'] = [x for x in np.random.random((len(da_pqlite),n_features))]
assert da_pqlite[:,'embedding'].shape == (500, 128)
da_pqlite.embeddings = np.random.random((len(da_pqlite),n_features))
assert da_pqlite[:,'embedding'].shape == (500, 128)
```

### DIFF
--- a/docarray/array/mixins/io/binary.py
+++ b/docarray/array/mixins/io/binary.py
@@ -24,6 +24,8 @@ class BinaryIOMixin:
         compress: Optional[str] = None,
         _show_progress: bool = False,
         streaming: bool = False,
+        *args,
+        **kwargs,
     ) -> Union['DocumentArray', Generator['Document', None, None]]:
         """Load array elements from a compressed binary file.
 
@@ -49,9 +51,13 @@ class BinaryIOMixin:
                 protocol=protocol,
                 compress=compress,
                 _show_progress=_show_progress,
+                *args,
+                **kwargs,
             )
         else:
-            return cls._load_binary_all(file_ctx, protocol, compress, _show_progress)
+            return cls._load_binary_all(
+                file_ctx, protocol, compress, _show_progress, *args, **kwargs
+            )
 
     @classmethod
     def _load_binary_stream(
@@ -60,6 +66,8 @@ class BinaryIOMixin:
         protocol=None,
         compress=None,
         _show_progress=False,
+        *args,
+        **kwargs,
     ) -> Generator['Document', None, None]:
         """Yield `Document` objects from a binary file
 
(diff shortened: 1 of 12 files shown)


hanxiao (changes requested):
only `DocumentArray` class method need to be extended. not `Documet` class method

docarray/array/mixins/io/binary.py
```diff
@@ -49,6 +49,8 @@ def load_binary(
                 protocol=protocol,
                 compress=compress,
                 _show_progress=_show_progress,
+                *args,
+                **kwargs,
```
alaeddine-13, line 53: I don't agree on this
Actually we can't really stream a storage document array right ?

davidbp:
> only `DocumentArray` class method need to be extended. not `Documet` class method

Ok, reverted

docarray/array/mixins/io/json.py
```diff
@@ -53,21 +55,21 @@ def load_json(
...
+            return cls(constructor(v, protocol=protocol) for v in fp)
 
     @classmethod
     def from_json(
-        cls: Type['T'], file: Union[str, TextIO], protocol: str = 'jsonschema', **kwargs
+        cls: Type['T'],
+        file: Union[str, TextIO],
+        protocol: str = 'jsonschema',
     ) -> 'T':
-        return cls.load_json(file, protocol=protocol, **kwargs)
```
hanxiao, line 62: if this is targeting master, then this diff shows you over-delete something which supposed to be there

davidbp: good catch, reverted

hanxiao (approved):

PR #94, opened 2022-02-03
perf: implement append to avoid invalidating id2offset

alaeddine-13:
slack discussion: https://jinaai.slack.com/archives/C02AC4T8Y5T/p1643893118678839

### DIFF
--- a/docarray/array/storage/memory/seqlike.py
+++ b/docarray/array/storage/memory/seqlike.py
@@ -24,7 +24,7 @@ class SequenceLikeMixin(MutableSequence[Document]):
         """
         self._data.append(value)
         if not self._needs_id2offset_rebuild:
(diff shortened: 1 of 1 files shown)


docarray/array/storage/memory/seqlike.py
```diff
@@ -17,6 +17,15 @@ def insert(self, index: int, value: 'Document'):
...
         self._data.insert(index, value)
 
+    def append(self, value: 'Document'):
+        """Append `doc` to the end of the array.
+
+        :param value: The doc needs to be appended.
+        """
+        self._data.append(value)
+        if not self._needs_id2offset_rebuild:
+            self._id_to_index[value.id] = len(self) - 1
```
hanxiao, line 27: can you implement faster `extend` as well?

PR #96, opened 2022-02-04
feat: transformer embed support

bwanglzu:


docarray/array/mixins/embed.py
```diff
@@ -44,6 +44,12 @@ def _set_embeddings_keras(
         with device:
             for b_ids in self.batch_ids(batch_size):
                 r = embed_model(self[b_ids, 'tensor'], training=False)
+                if not isinstance(r, tf.Tensor):
```
JoanFM, line 47: wouldn't it make more sense to apply positive logic?

JoanFM (changes requested):

docarray/array/mixins/embed.py
bwanglzu: as discussed it is because we want to import within the `else` branch.

bwanglzu:
as was discussed with @davidbp will fix weaviate storage errors after #93

PR #99, opened 2022-02-06
feat(storage): add Qdrant support

generall:
Open questions:

- work with offsets
- creation of a new collection on select operations
- Distances
- Mixin inter-dependencies


docarray/array/storage/qdrant/getsetdel.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    def scan(self) -> Iterator['Document']:
+        offset = None
+        while True:
+            response = self.client.http.points_api.scroll_points(
+                name=self.collection_name,
+                scroll_request=ScrollRequest(
+                    offset=offset,
+                    limit=SCROLL_BATCH_SIZE,
```
alaeddine-13, line 107: can we add this to the config instead ?

docarray/array/storage/qdrant/getsetdel.py
```diff
@@ -0,0 +1,118 @@
+from typing import List, Iterable, Iterator
+
+import scipy.sparse
+from qdrant_client import QdrantClient
+from qdrant_openapi_client.exceptions import UnexpectedResponse
+from qdrant_openapi_client.models.models import PointIdsList, PointsList, ScrollRequest
+
+from docarray import Document
+from docarray.array.storage.base.getsetdel import BaseGetSetDelMixin
+from docarray.types import ArrayType
```
alaeddine-13, line 10:
```suggestion
if TYPE_CHECKING:
    from docarray.types import ArrayType
```

docarray/array/storage/qdrant/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @property
+    def collection_name(self) -> str:
+        return self._config.collection_name
+
+    @property
+    def config(self):
+        return self._config
+
+    def client(self) -> QdrantClient:
```
alaeddine-13, line 41:
```suggestion
    @property
    def client(self) -> QdrantClient:
```

alaeddine-13 (dismissed):
Thanks for your contribution!
I submitted some suggestions.
We also need to add qdrant to `docarray/array/document.py` to enable this syntax:
`da = DocumentArray(storage='qdrant')`
Furthermore, in order to ensure the quality of the Qdrant DocumentArray, tests supported should be activated for Qdrant. We can add a `start_qdrant` fixture in tests/unit/array/conftest.py to start qdrant docker image on module level. Or we can add it to the same docker compose file as weaviate and rename the fixture from start_weaviate to `start_storage`

Find method will be added soon for storage backends that support ANN, it would be great to add this to qdrant as well

docarray/array/storage/qdrant/backend.py
```diff
@@ -0,0 +1,100 @@
...
+        DocumentArraySourceType,
+    )
+
+
+@dataclass
+class QdrantConfig:
+    distance: Distance = field(default=Distance.COSINE)
+    collection_name: Optional[str] = None
+    connection: Optional[Union[str, QdrantClient]] = field(default="localhost:6333")
+    n_dim: Optional[int] = field(default=1)
```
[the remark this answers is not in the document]
alaeddine-13: let's raise a `ValueError`  in that case

docarray/array/storage/qdrant/getsetdel.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    def _qdrant_to_document(self, qdrant_record: dict) -> 'Document':
+        return Document.from_base64(
+            qdrant_record['payload']['_serialized'], **self.serialization_config
+        )
+
+    def _document_to_qdrant(self, doc: Document, _id: str = None) -> dict:
+        return dict(
+            id=_id or doc.id,
```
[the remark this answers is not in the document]
alaeddine-13: Alright, the current behaviour of DocumentArray is to rely on `value.id` as the new ID and `_id` as the existing, old ID.
So basicly, in case both IDs are different, this should change the ID of the Document in the storage backend.

docarray/array/storage/qdrant/getsetdel.py
```diff
@@ -0,0 +1,122 @@
...
+    def n_dim(self) -> int:
+        raise NotImplementedError()
+
+    @property
+    def collection_name(self) -> str:
+        raise NotImplementedError()
+
+    @property
+    def scroll_batch_size(self) -> int:
+        raise NotImplementedError()
```
numb3r3, line 35: are these functions are necessary? I found they are implemented in `StorageMixins`

docarray/array/storage/qdrant/seqlike.py
```diff
@@ -0,0 +1,67 @@
...
+        raise NotImplementedError()
+
+    @property
+    def collection_name(self) -> str:
+        raise NotImplementedError()
+
+    @property
+    def config(self):
+        raise NotImplementedError()
+
```
numb3r3, line 20: same

generall:
Fixed:
- wrong result processing in `FindMixin`
- increase number of open files limit for tests: there are a lot of collections created, should not be required regular work
- Implemented data uploaded with batches - there is 32mb limit for single request, uploading of images should be splitted into multiple batches

docarray/array/storage/qdrant/helper.py
```diff
@@ -22,4 +24,6 @@ def embedding_to_array(
         if embedding.ndim > 1:
             embedding = np.asarray(embedding).squeeze()
 
+        if np.all(embedding == 0):
+            embedding = embedding + EPSILON
```
JoanFM, line 28: Isnt this a little expensive? Also in some cases embeddings are not even numpy arrays and may live in GPU

docarray/array/storage/qdrant/find.py
```diff
@@ -0,0 +1,92 @@
...
+        'QdrantArrayType',
+        np.ndarray,
+        tensorflow.Tensor,
+        torch.Tensor,
+        Sequence[float],
+    )
+
+
+class FindMixin:
+    @property
```
JoanFM, line 31: I think if u declare as abstract method it will not allow u to instantiate the class and will fail faster.

docarray/array/storage/qdrant/helper.py
alaeddine-13: Qdrant does not accept null vectors because cosine distance can't be computed
Basicly we have 2 options:

- prohibit users from using null vectors as embedding
- implicitly add Epsilon to allow computation
your thoughts ?

numb3r3: I prefer the second one, as same as we have done in weaviate.

docarray/array/storage/qdrant/find.py
alaeddine-13: done

hanxiao (approved):

PR #101, opened 2022-02-08
fix pqlite 

numb3r3:
- use `serilize_config` instead, same as in `DocumentArraySqlite`
- override `__getstate__` and `__setstate__` for supporting pickle
- pass `test_construct.py`
- [x] support all unittest

### DIFF
--- a/docarray/array/storage/pqlite/backend.py
+++ b/docarray/array/storage/pqlite/backend.py
@@ -1,10 +1,13 @@
-from dataclasses import dataclass, asdict
-from pathlib import Path
+import itertools
+from dataclasses import dataclass, asdict, field
 from typing import (
     Union,
     Dict,
     Optional,
     TYPE_CHECKING,
+    Sequence,
+    Generator,
+    Iterator,
 )
 
 from ..base.backend import BaseBackendMixin
@@ -20,7 +23,7 @@ if TYPE_CHECKING:
 class PqliteConfig:
     n_dim: int = 1
     metric: str = 'cosine'
-    serialize_protocol: str = 'pickle'
+    serialize_config: Dict = field(default_factory=dict)
     data_path: Optional[str] = None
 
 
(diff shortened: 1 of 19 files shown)


docarray/array/storage/pqlite/backend.py
```diff
@@ -53,11 +58,56 @@ def _init_storage(
...
+        if _docs is None:
+            return
+        elif isinstance(
+            _docs, (DocumentArray, Sequence, Generator, Iterator, itertools.chain)
+        ):
             self.clear()
-            self.extend(docs)
+            self.extend(_docs)
+        else:
+            if isinstance(_docs, Document):
```
hanxiao, line 76:
```suggestion
            self.clear()
            if isinstance(_docs, Document):
```

docarray/array/storage/pqlite/backend.py
```diff
@@ -53,11 +58,56 @@ def _init_storage(
...
+        )
+
+    def _get_storage_infos(self) -> Dict:
+        storage_infos = super()._get_storage_infos()
+        return {
+            'Backend': 'PQLite (https://github.com/jina-ai/pqlite)',
+            'Distance Metric': self._pqlite.metric.name,
+            'Data Path': self._config.data_path,
+            'Serialization Protocol': self._config.serialize_config.get(
+                'protocol', 'pickle'
```
alaeddine-13, line 110: doesn't look like a good place to put a default value
let's center all of these somewhere, or just modify the factory of serialize config

numb3r3: fixed

docarray/array/storage/pqlite/backend.py
numb3r3: fixed

docarray/array/storage/pqlite/backend.py
```diff
@@ -20,7 +23,7 @@
 class PqliteConfig:
     n_dim: int = 1
```
alaeddine-13, line 24: this pattern is not supported in weaviate and I guess the same reason applies for pqlite
Actually, implicitly setting vectors of dim=1 by default is problematic, because users will try to add empty Documents and then find errors when they want to update embeddings.
Instead, what we should do is make n_dim a positional argument and raise an error when no config object is passed
What do you think ?

numb3r3: @alaeddine-13 Agree, I have fixed it, please give check!

PR #103, opened 2022-02-08
refactor(backend): change weaviate config

hanxiao:


docarray/array/storage/weaviate/backend.py
```diff
@@ -55,22 +55,15 @@ def _init_storage(
...
-        self.n_dim = config.n_dim
-        self.serialize_config = config.serialize_config
+            config = dataclass_from_dict(WeaviateConfig, config)
 
-        import weaviate
+        from ... import DocumentArray
 
-        if config.client is None:
-            config.client = 'http://localhost:8080'
+        self._n_dim = config.n_dim
```
alaeddine-13, line 65: I don't think n_dim should become private
I suggest the following
```
@property
    def n_dim(self) -> int:
        return self._config.n_dim
```

PR #104, opened 2022-02-09
fix: weaviate da del behavior

winstonww:
Align the del/destructor behavior of `DocumentArrayWeaviate` with `DocumentArraySqlite` (delete data only when `self._persist` is `True`, that is when table name/name in provided in `config`).

Also, adding a check to assert `name` provided is capitalized. This is because if `name` is not capitalized, Weaviate will automatically capitalize the name but users may not be aware of this special case.

### DIFF
--- a/docarray/array/storage/registry.py
+++ b/docarray/array/storage/registry.py
@@ -1,0 +1,4 @@
+from collections import defaultdict
+
+
+_REGISTRY = defaultdict(lambda: defaultdict(list))
--- a/docarray/array/storage/weaviate/backend.py
+++ b/docarray/array/storage/weaviate/backend.py
@@ -19,6 +19,7 @@ import weaviate
 from ..base.backend import BaseBackendMixin
 from .... import Document
 from ....helper import dataclass_from_dict
+from ..registry import _REGISTRY
 
 if TYPE_CHECKING:
     from ....types import (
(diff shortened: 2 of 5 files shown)


JoanFM (changes requested):
What is this del doing? Are we sure this behavior is desired when garbage collector kicks in?

winstonww:
@JoanFM Yes, an issue is indeed observed and is under investigation detailed in the discussion [here](https://jinaai.slack.com/archives/C02AC4T8Y5T/p1644324667978449). Thoughts appreciated, thanks

docarray/array/storage/weaviate/backend.py
```diff
@@ -65,6 +65,11 @@ def _init_storage(
         self._n_dim = config.n_dim
         self._serialize_config = config.serialize_config
 
+        if config.name and config.name != config.name.capitalize():
+            raise ValueError('weaviate class name has to be capitalized')
```
numb3r3, line 69: Given this error report, the user would still not know how to correct the configurations. Could you give more clues in the output?

winstonww: Yep, addressed thank you.

docarray/array/storage/registry.py
```diff
@@ -0,0 +1,4 @@
+from collections import defaultdict
+
+
+_REGISTRY = defaultdict(lambda: defaultdict(list))
```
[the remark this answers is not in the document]
winstonww: This is suggested by @alaeddine-13  to fix the issue detailed [here](https://jinaai.slack.com/archives/C02AC4T8Y5T/p1644399473166809?thread_ts=1644324667.978449&cid=C02AC4T8Y5T)/

alaeddine-13 (approved):
LGTM

PR #106, opened 2022-02-09
feat: add series as index container

alaeddine-13:
feat: add series as index container


setup.py
```diff
@@ -55,6 +55,8 @@
             'uvicorn',
             'weaviate-client~=3.3.0',
             'pqlite>=0.2.3',
+            'pandas',
+            'pqlite>=0.2.3',
```
hanxiao, line 59: `pqlite` is already in line 57

docarray/array/storage/memory/seqlike.py
```diff
@@ -66,7 +67,5 @@ def __add__(self, other: Union['Document', Iterable['Document']]):
         return v
 
     def extend(self, values: Iterable['Document']) -> None:
-        values = list(values)  # consume the iterator only once
-        last_idx = len(self._id2offset)
-        self._data.extend(values)
-        self._id_to_index.update({d.id: i + last_idx for i, d in enumerate(values)})
+        for v in values:
+            self._data[v.id] = v
```
hanxiao, line 71: another option is:
```
values = list(values)
tmp = pd.Series(values, index=(d.id for d in values))
self._data = self._data.append(tmp)
```

Otherwise, there is no need to implement `extend`, it will automatically call `append()`/ `insert()`, simply remove it

PR #110, opened 2022-02-11
feat: integrate match and find function

numb3r3:
**WANING**: To enable the `match` function to work,   the implementation of `find()` needs to allow the query input as `DocumentArray` (since we need to support **exclude_self** parameter).

```
import numpy as np
N = 100
D = 8

da = DocumentArray([Document(id=f'{i}', embedding=x) for i, x in zip(range(N), np.random.rand(N, D))])

q = Document(embedding=np.random.random(D))

### query is `Document`
result = da.find(q, limit=10)
q.match(da, limit=10)

### query is `DocumentArray`
qa = DocumentArray(q)
result = da.find(qa, limit=10)
qa.match(da, limit=10)

qa = DocumentArray([q] * 2, limit=10)
result = da.find(qa, limit=10)
qa.match(da)

### query is `ndarray`
qx = np.random.random(D)
result = da.find(qx, limit=10)

qx = np.random.rand(5, D)
result = da.find(qx, limit=10)
```

## DocumentArrayPqlite

```
pq_da = DocumentArray(da, storage='pqlite', config={'n_dim': D})
result = pq_da.find(qa, limit=10)
qa.match(pq_da)
print(qa.summary())

```

### DIFF
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -3,9 +3,11 @@ from ...math.helper import top_k, minmax_normalize, update_rows_x_mat_best
 from ...score import NamedScore
 import numpy as np
 
+
 if TYPE_CHECKING:
     from ...types import T, ArrayType
-    from ... import Document, DocumentArray
+
+    # from ... import Document, DocumentArray
 
 
 class FindMixin:
@@ -33,6 +35,7 @@ class FindMixin:
         **kwargs,
     ) -> 'DocumentArray':
         from ...math.ndarray import to_numpy_array, get_array_type
+        from ... import Document, DocumentArray
 
         if isinstance(query, (dict, str)):
             return self.filter(query, **kwargs)
@@ -147,6 +150,8 @@ class FindMixin:
         else:
             dist, idx = self._search(query, cdist, _limit, normalization, metric_name)
 
+        from ... import Document, DocumentArray
+
         result = DocumentArray([Document(id=q.id) for q in query])
         for _q, _ids, _dists in zip(result, idx, dist):
             num_matches = 0
@@ -186,7 +191,7 @@ class FindMixin:
         :return: distances and indices
         """
         dists = cdist(query.embeddings, self.embeddings, metric_name)
-        dist, idx = top_k(dists, min(limit, len(query)), descending=False)
+        dist, idx = top_k(dists, min(limit, len(self)), descending=False)
         if isinstance(normalization, (tuple, list)) and normalization is not None:
             # normalization bound uses original distance not the top-k trimmed distance
             min_d = np.min(dists, axis=-1, keepdims=True)
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -1,4 +1,4 @@
-from typing import overload, Optional, Union, List, Dict, Tuple, Callable, TYPE_CHECKING
+from typing import overload, Optional, Union, Dict, Tuple, Callable, TYPE_CHECKING
 from ...math.helper import top_k, minmax_normalize, update_rows_x_mat_best
 from ...score import NamedScore
 import numpy as np
@@ -7,7 +7,7 @@ import numpy as np
 if TYPE_CHECKING:
     from ...types import T, ArrayType
 
-    # from ... import Document, DocumentArray
+    from ... import Document, DocumentArray
 
 
 class FindMixin:
@@ -31,7 +31,7 @@ class FindMixin:
 
     def find(
         self: 'T',
-        query: Union['DocumentArray', 'Document', 'ArrayType', Callable, Dict, str],
+        query: Union['DocumentArray', 'Document', 'ArrayType', Dict, str],
         **kwargs,
     ) -> 'DocumentArray':
         from ...math.ndarray import to_numpy_array, get_array_type
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -39,11 +39,6 @@ class FindMixin:
 
         if isinstance(query, (DocumentArray, Document)):
 
-            print("\n\nHEREEEE!!!!\n\n")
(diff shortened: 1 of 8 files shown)


docarray/array/mixins/find.py
```diff
@@ -0,0 +1,268 @@
...
+
+            return self.search(query, **kwargs)
+
+        try:
+            _, _ = get_array_type(query)
+            q_mat = to_numpy_array(query)
+            query = DocumentArray([Document(embedding=x) for x in q_mat])
+            if q_mat.ndim == 1:
+                query = DocumentArray(Document(embedding=q_mat))
+                return self.search(query, **kwargs)[0].matches
```
alaeddine-13, line 55: I think it's better not to create Documents when searching by vectors. Maybe we can keep this section custom for pqlite and weaviate so that we just search by vectors, without having any documents created

docarray/array/mixins/find.py
```diff
@@ -0,0 +1,268 @@
...
+        from ... import Document, DocumentArray
+
+        if isinstance(query, (dict, str)):
+            return self.filter(query, **kwargs)
+        elif isinstance(query, (DocumentArray, Document)):
+            if isinstance(query, Document):
+                query = DocumentArray(query)
+                return self.search(query, **kwargs)[0].matches
+
+            return self.search(query, **kwargs)
```
alaeddine-13, line 47: I think we should either return a document array of matches or a list of match document arrays, not root documents

docarray/array/mixins/match.py
```diff
@@ -65,170 +65,20 @@ def match(
...
+            batch_size=batch_size,
+            exclude_self=exclude_self,
+            only_id=only_id,
+            use_scipy=use_scipy,
+            device=device,
+            num_worker=num_worker,
+        )
+
+        for _r, _d in zip(result, self):
+            _d.matches = _r.matches
```
alaeddine-13, line 84: why do we have to copy the query inside `search` and then fill the matches again at the level of match ?
I think search should return a list of match document arrays (if len(query) > 1) or a document array of matches (if len(query) == 1)

alaeddine-13:
I also think that weaviate is not affected by this PR right ?

docarray/array/mixins/find.py
numb3r3: > Maybe we can keep this section custom for pqlite and weaviate so that we just search by vectors

By this way, the implementation of `search` must **validate the shape of query** to determin how to return the results. This is where error usually happens. So, in the latest commit, the signature of function `search()` looks like:
```
def search(
        self: 'T',
        query: 'DocumentArray',
        ...
) -> List['DocumentArray']
```
The search function always accepts a da query and returns a list of matched da .

docarray/array/mixins/find.py
numb3r3: Addressed

docarray/array/mixins/find.py
numb3r3: > without having any documents created

I made some revisions that the new-created docs will only be used as auxiliary, and will not being included in the result.

docarray/array/mixins/match.py
numb3r3: > I also think that weaviate is not affected by this PR right ?

the weaviate `search` has been refactored in this PR

davidbp:
Weaviate backend can’t handle tests with a vector of zeros (I guess because cosine distance is not defined in this case). I’ve added a 
```python
        is_all_zero = np.all(query == 0)
        if is_all_zero:
            query = query + EPSILON
```
where  `EPSILON = 1.0e-9`

docarray/array/mixins/find.py
```diff
@@ -0,0 +1,270 @@
...
+        :param normalization: a tuple [a, b] to be used with min-max normalization,
+                                the min distance will be rescaled to `a`, the max distance will be rescaled to `b`
+                                all values will be rescaled into range `[a, b]`.
+        :param metric_name: if provided, then match result will be marked with this string.
+        :return: distances and indices
+        """
+
+        # Ensure query is a 'matrix' since cdist(X,Y) computes all pairwise distances
+        # between rows of X and rows of Y.
+        if query.embeddings.ndim != 2:
```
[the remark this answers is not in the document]
numb3r3: or this can be address by updating method of `unravel` in `math.ndarray.py`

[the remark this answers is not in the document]
davidbp: one of the  tests that you added yesterday precisely broke because of this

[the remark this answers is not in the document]
davidbp: unravel does too much stuff to achieve what I want in that line of code.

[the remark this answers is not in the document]
numb3r3: I see, what about using this commit https://github.com/jina-ai/docarray/pull/110/commits/f3a92d778ea44819c0cf191e8205257cd2b1971e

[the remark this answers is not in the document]
davidbp: seems to work

tests/unit/array/mixins/test_match.py
```diff
@@ -68,6 +69,23 @@ def doc_lists_to_doc_arrays(doc_lists, *args, **kwargs):
...
+    D1, D2 = doc_lists_to_doc_arrays(doc_lists)
+
+    if config:
+        da = DocumentArray(D2, storage=storage, config=config)
+    else:
+        da = DocumentArray(D2, storage=storage)
+
+    D1.match(da, limit=limit)
+    for m in D1[:, 'matches']:
+        assert len(m) == limit
```
alaeddine-13, line 86: does it make sense to have further checks on the order ?

numb3r3: done

davidbp (approved):
LGTM

hanxiao (approved):

PR #111, opened 2022-02-11
fix(weaviate): create documents in a batch

StefanBogdan:
At the moment some unit tests fail, this can be caused by the fact that objects from batch are not flushed. In unit tests, after all the documents are imported/added, it needs to flush the remaining ones by using the `DocumentArrayWeaviate.flush_weaviate_batch()` method, [here](https://github.com/semi-technologies/docarray/blob/main/docarray/array/storage/weaviate/backend.py#L333-L336).


davidbp:
Thanks for the PR @StefanBogdan, will take a look at it on monday, and come back to you.

docarray/array/storage/weaviate/backend.py
```diff
@@ -322,3 +329,8 @@ def _get_storage_infos(self) -> Dict:
             'Serialization Protocol': self._config.serialize_config.get('protocol'),
             **storage_infos,
         }
+
+    def flush_weaviate_batch(self) -> None:
+        """Flush existing objects in Weaviate Batch to the Weaviate Backend Server"""
```
numb3r3, line 334: Without flushing, the objects will be in-memory batch of client-side? or backend server side?

docarray/array/storage/weaviate/backend.py
```diff
@@ -76,8 +76,19 @@ def _init_storage(
...
+                dynamic=True,
+                timeout_retries=5,
+            )
         else:
-            self._client = config.client
+            self._client: weaviate.Client = config.client
+            if self._client.batch.batch_size is None:
+                self._client.batch.configure(
+                    batch_size=100,
+                    dynamic=True,
```
numb3r3, line 89: What's the meaning of `dynamic=True` ?

docarray/array/storage/weaviate/backend.py
StefanBogdan: In memory of client-side.

docarray/array/storage/weaviate/backend.py
StefanBogdan: After the batch creation the `batch_size` would be set to the `recommended` number objects. [Here](https://weaviate-python-client.readthedocs.io/en/latest/weaviate.batch.html#weaviate.batch.Batch) is the docs page, it is CASE III.

numb3r3: I see, thanks

docarray/array/storage/weaviate/backend.py
```diff
@@ -174,21 +185,17 @@ def _load_or_create_weaviate_schema(self):
...
-            )
+            
+        # this call will either create or replace an object (if uuid exists)
+        # since Meta class has a single property: '_offset2ids', we can replace
+        # the object instead of updating it.
+        self._client.batch.add_data_object(
+            data_object={'_offset2ids': self._offset2ids},
+            class_name=self._meta_name,
+            uuid=self._offset2ids_wid,
+        )
```
numb3r3, line 198: Is it a good idea to use context manager `with self._client.batch` instead, so as to automatically flushing batch?

StefanBogdan: One would use the context manager when importing/creating data in the same function, otherwise it will flush the objects once it exits the context manager. In this case it would, if the batch is empty and we call this method it will flush the batch with only one object. This is not something we want, because it is the same as creating the object explicitly.

PR #112, opened 2022-02-13
fix: use utf-8 encoding for csv reading

jacobowitz:
I am not 100% sure about this PR, so please think about it **carefully**.

In Jina the chatbot example is currently broken in Windows, because the csv dataset cant be read correctly. This file [here]('https://static.jina.ai/chatbot/dataset.csv') .

Reading this file with `DocumentArray.from_csv` will result in this error:

`
File "c:\users\tobias\pycharmprojects\hello_jina\venv\lib\site-packages\docarray\document\generators.py", line 302, in _sample
    for i in iterable:
  File "C:\Users\tobias\AppData\Local\Programs\Python\Python38\lib\csv.py", line 111, in next
    row = next(self.reader)
  File "C:\Users\tobias\AppData\Local\Programs\Python\Python38\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 3204: character maps to <undefined>
`

The problem is that the csv file is opened without providing any encoding. This means [Python will use some OS dependant encoding](https://docs.python.org/3/library/functions.html#open). Linux correctly assumes 'utf-8', but Windows tries `cp1252` and fails.

Thus this PR hard codes the encoding to `utf-8`, which should work in most cases I think. I am not sure about some undesired side effects in some Operating systems though.

### DIFF
--- a/docarray/document/generators.py
+++ b/docarray/document/generators.py
@@ -133,6 +133,7 @@ def from_csv(
         parameters together. If you don't know the dialect and the default one does not work for you,
         you can try set it to ``auto``.
     :param encoding: encoding used to read the CSV file. By default, ``utf-8`` is used.
+
     :yield: documents
 
     """
--- a/docarray/document/generators.py
+++ b/docarray/document/generators.py
@@ -117,6 +117,7 @@ def from_csv(
     size: Optional[int] = None,
     sampling_rate: Optional[float] = None,
     dialect: Union[str, 'csv.Dialect'] = 'excel',
+    encoding: str = 'utf-8',
     *args,
     **kwargs,
 ) -> Generator['Document', None, None]:
@@ -131,6 +132,7 @@ def from_csv(
         predefined dialects in your system, or could be a :class:`csv.Dialect` class that groups specific formatting
         parameters together. If you don't know the dialect and the default one does not work for you,
         you can try set it to ``auto``.
+    :param encoding: encoding used to read the CSV file. By default, ``utf-8`` is used.
     :yield: documents
 
     """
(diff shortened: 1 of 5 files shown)


JoanFM (commented):
Maybe we can simply add `encoding` as an option with `utf-8` as default

docarray/document/generators.py
```diff
@@ -139,7 +139,7 @@ def from_csv(
     if hasattr(file, 'read'):
         file_ctx = nullcontext(file)
     else:
-        file_ctx = open(file, 'r')
+        file_ctx = open(file, 'r', encoding='utf-8')
```
numb3r3, line 142: And the writing function `save_embeddings_csv` has the save issue.

numb3r3 (changes requested):

samsja:
FYI: that's how pandas deal with it : https://github.com/pandas-dev/pandas/blob/b365376044c28f8f0a6767c1657bd62327d7ae09/pandas/io/parsers/readers.py#L1054
I am not sure what the  `sys.getfilesystemencoding()` is returning on windows

JoanFM:
> FYI: that's how pandas deal with it : https://github.com/pandas-dev/pandas/blob/b365376044c28f8f0a6767c1657bd62327d7ae09/pandas/io/parsers/readers.py#L1054 I am not sure what the `sys.getfilesystemencoding()` is returning on windows

This would make sense, lets implement this

numb3r3:
> This would make sense, lets implement this

I would not implement like this way. To make sure the encoding of the dumped files is **system independently**, I recommend using `UTF-8` as the default without special reason.

numb3r3 (approved):

JoanFM (approved):

PR #114, opened 2022-02-14
refactor(graphql): strawberry to input

JoanFM:
- [x] make sure the docs script is run out of the box https://docarray.jina.ai/advanced/graphql-support/

### DIFF
--- a/docarray/document/mixins/strawberry.py
+++ b/docarray/document/mixins/strawberry.py
@@ -13,7 +13,7 @@ class StrawberryMixin:
     def to_strawberry_type(self) -> 'StrawberryDocument':
         """Convert a Document object into a Strawberry type."""
         from ..strawberry_type import StrawberryDocument as SD
-        from ..strawberry_type import NameScoreItem, NamedScore
+        from ..strawberry_type import _NameScoreItem, _NamedScore
 
         _p_dict = {}
         for f in self.non_empty_fields:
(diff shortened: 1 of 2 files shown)


deepankarm (changes requested):
We need both type & input.

samsja:
I don't think there is a way to create a strawberry type from a strawberry object or vice versa even if object seems to be a type under the hood but instantiate differently. There is no method to do it proprely and and did not find a hack by looking at the source code. So I propose to keep one source of truth and create both type and object from it

JoanFM:
Why do we need both? is that clear @samsja ?

samsja:
> Why do we need both? is that clear @samsja ?

I am not sure to understand 100% all of the concept but what I understand is that `input` is more restrictive than  `type`. 

On the core part we can't use the `type` we need `input` and on the docarray part they seems to need the `type`.

So I need to check with docarray team if they really need the `strawberry.type` but I guess so.

@alaeddine-13 maybe ?

JohannesMessner:
> Why do we need both? is that clear @samsja ?

If my understanding is correct, one is what you put into a query, and one is what you get out. So even if, on a technical level, we could get away with using just one for both purposes, we probably shouldn't, because they are conceptually quite different.

JohannesMessner (approved):

PR #116, opened 2022-02-15
refactor(array): change primary view to dict and 2nd view to list

alaeddine-13:
feat: use list for offset2id
Goal: Have a unified offset2id data structure for all storage backends and pull out the implementation details of all offset-related operations from the implementations of storage backends.

For now, we are relying on python list because all operations that we currently need will be delegated to built in methods of list: There shouldn't be any pure python for loops interacting with offset2ids.

TODOs:

- [x] fix tests for DocumentArrayInMemory
- [x] leverage offset2ids in DocumentArrayWeaviate
- [x] leverage offset2ids in DocumentArrayPQLite
- [x] leverage offset2ids in DocumentArraySqlite
- [x] take care of operations that can be done in batches (for example, pqlite allows batch updates), _del_docs_by_ids not implemented in base, ...
- [x] allow save/load of offset2ids on `__init__` and `__del__`
* weaviate: meta schema
* sqlite: use another column
* pqlite: sqlite table
- [x] __getstate__ and __setstate__ to also include offset2ids
- [x] TODOs
- [x] Update create storage backend documentation

### DIFF
--- a/docarray/array/storage/base/getsetdel.py
+++ b/docarray/array/storage/base/getsetdel.py
@@ -22,6 +22,11 @@ class BaseGetSetDelMixin(ABC):
 
         Other methods implemented a generic-but-slow version that leverage the methods above.
         Please override those methods in the subclass whenever a more efficient implementation is available.
+        Mainly, if the backend storage supports operations in batches, you can implement the following methods:
+            - :meth:`._get_docs_by_ids`
+            - :meth:`._set_docs_by_ids`
+            - :meth:`._del_docs_by_ids`
+
     """
 
     # Getitem APIs
@@ -79,9 +84,8 @@ class BaseGetSetDelMixin(ABC):
         Override this function if there is a more efficient logic
         :param _slice: the slice used for indexing
         """
-        # TODO: rebuild offset2ids once
-        for _id in self._offset2ids.get_id(_slice):
-            self._del_doc(_id)
+        ids = self._offset2ids.get_id(_slice)
+        self._del_docs(ids)
 
     def _del_docs_by_mask(self, mask: Sequence[bool]):
         """This function is derived and may not have the most efficient implementation.
@@ -89,9 +93,8 @@ class BaseGetSetDelMixin(ABC):
         :param mask: the boolean mask used for indexing
         """
         # TODO: rebuild offset2ids once
-        idxs = list(itertools.compress(self._offset2ids, (not _i for _i in mask)))
-        for _idx in reversed(idxs):
-            self._del_doc(_idx)
+        ids = list(itertools.compress(self._offset2ids, (not _i for _i in mask)))
+        self._del_docs(ids)
 
     def _del_all_docs(self):
         """This function is derived and may not have the most efficient implementation.
@@ -99,6 +102,19 @@ class BaseGetSetDelMixin(ABC):
         Override this function if there is a more efficient logic"""
         self._clear_storage()
         self._offset2ids = Offset2ID()
+
+    def _del_docs_by_ids(self, ids):
+        """This function is derived from :meth:`_del_doc_by_id`
+        Override this function if there is a more efficient logic
+
+        :param ids: the ids used for indexing
+        """
+        for _id in ids:
+            self._del_doc_by_id(_id)
+
+    def _del_docs(self, ids):
+        self._del_docs_by_ids(ids)
+        self._offset2ids.delete_by_ids(ids)
 
     @abstractmethod
     def _clear_storage(self):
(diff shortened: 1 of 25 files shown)


JoanFM:
Hey @alaeddine-13 

Can u add more details on why this is the chosen implementation?

alaeddine-13:
> Hey @alaeddine-13
> 
> Can u add more details on why this is the chosen implementation?

description updated

docarray/array/storage/base/helper.py
```diff
@@ -0,0 +1,42 @@
...
+        del self.offset2id[self.offset2id.index(_id)]
+
+    def index(self, _id):
+        return self.offset2id.index(_id)
+
+    def delete_by_offset(self, position):
+        del self.offset2id[position]
+
+    def insert(self, position, data_id):
+        self.offset2id.insert(position, data_id)
```
numb3r3, line 27: Since the basic data container is a Dict now. I have to decide whether do we need to check the inserted `data_id` is unique?

docarray/array/storage/base/helper.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def clear(self):
+        self.offset2id.clear()
+
+    def save(self):
```
numb3r3, line 32: The `serialization` and `deserialization` are needed for storage backend, right?

docarray/array/storage/base/getsetdel.py
```diff
@@ -76,35 +84,74 @@ def _del_docs_by_slice(self, _slice: slice):
...
+    def _clear_storage(self):
         ...
 
+    # Setitem API
+
+    def _set_doc_by_offset(self, offset: int, value: 'Document'):
+        self._set_doc(self._offset2ids.get_id(offset), value)
+
+    def _set_doc(self, _id: str, value: 'Document'):
+        if _id != value.id:
```
JoanFM, line 129: Here we should think what happens if the `set_doc_by_id` is not successful.

Maybe we should update offset after and if no exception is raised

docarray/array/storage/base/helper.py
alaeddine-13: this is the last step, taking care of it currently

docarray/array/storage/base/getsetdel.py
```diff
@@ -76,35 +84,74 @@ def _del_docs_by_slice(self, _slice: slice):
...
+
+        :param ids: the ids used for indexing
+        """
+        for _id, doc in zip(ids, docs):
+            self._set_doc_by_id(_id, doc)
+
+    def _set_docs(self, ids, docs: Iterable['Document']):
+        docs = list(docs)
+        self._set_docs_by_ids(ids, docs)
+        mismatch_ids = {_id: doc.id for _id, doc in zip(ids, docs) if _id != doc.id}
```
JoanFM, line 152: This seems like an extensive iteration, why do we require ids? why don't just trust the ones from the `docs`?

docarray/array/storage/base/getsetdel.py
```diff
@@ -185,10 +234,15 @@ def _set_doc_attr_by_id(self, _id: str, attr: str, value: Any):
         :param attr: the attribute of document to update
         :param value: the value doc's attr will be updated to
         """
+        if attr == 'id' and value is None:
+            raise ValueError(
+                'pop id from Document stored in a DocumentArray is not allowed'
```
JoanFM, line 239: This is a weird message, when am I asking to `pop`?

docarray/array/storage/base/helper.py
```diff
@@ -0,0 +1,57 @@
+from typing import Iterator, List, Tuple, Dict
+
+
+class Offset2ID:
+    def __init__(self):
+        self.offset2id = []
```
JoanFM, line 6: weird name. `Offset2ID` is the name. So this should be simply named `container` or `ids`

JoanFM (changes requested):

docarray/array/storage/base/getsetdel.py
alaeddine-13: fixed

docarray/array/storage/memory/getsetdel.py
```diff
@@ -53,28 +28,17 @@ def _set_doc_value_pairs(
...
 
-    def _get_docs_by_slice(self, _slice: slice) -> Iterable['Document']:
-        return self._data[_slice]
+    def _get_docs_by_ids(self, ids: Sequence[str]) -> Iterable['Document']:
+        return (self._data[_id] for _id in ids)
 
-    def _get_docs_by_offsets(self, offsets: Sequence[int]) -> Iterable['Document']:
-        return (self._data[t] for t in offsets)
+    def _clear_storage(self):
+        self._data.clear()
```
numb3r3, line 38: Where the `offset2ids` is cleared?

docarray/array/storage/base/helper.py
alaeddine-13: right
It doesn't make sense anyways for storage backends
so we don't allow docs with same ID ?

docarray/array/storage/memory/getsetdel.py
alaeddine-13: this method is only about clearinng the storage, not offset2ids
It is cleared in a parent class

docarray/array/storage/base/helper.py
numb3r3: I would prefer to disable documents have same ID

docarray/array/storage/base/helper.py
davidbp: renamed to `ids`

docarray/array/storage/pqlite/getsetdel.py
```diff
@@ -44,11 +44,11 @@ def _load_offset2ids(self):
             data_path=self._config.data_path, in_memory=False
         )
         self._offsetmapping.create_table()
-        self._offset2ids.offset2id = self._offsetmapping.get_all_ids()
+        self._offset2ids._ids = self._offsetmapping.get_all_ids()
```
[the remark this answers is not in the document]
JoanFM: or build another instance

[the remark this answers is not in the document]
JoanFM: if u set `_ids` u should do:

```python
self._offset2ids = Offset2ID(self._offsetmapping.get_all_ids())
```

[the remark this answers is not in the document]
alaeddine-13: it's public now

hanxiao (approved):

PR #118, opened 2022-02-17
[WIP] test: reduce test time

winstonww:
Work in Progress


hanxiao:
based on this number, it seems some tests are missing in this refactoring, could you revise your base script
image

winstonww:
The decrease in codecov test coverage seen before is because some tests results did not get picked up by codecov even though tests have actually run.

For instance, in [this](https://github.com/jina-ai/docarray/pull/118/commits/4291445007f1e01bc8117b3e57d70a538f9a8526) commit, the corresponding codecov build for [this](https://github.com/jina-ai/docarray/runs/5244266051?check_suite_focus=true) batch of tests does not seem to appear [here](https://codecov.io/gh/jina-ai/docarray/commit/4291445007f1e01bc8117b3e57d70a538f9a8526/build), even though builds for a few other test batches are correctly shown. 

It's not very clear to me why the above only occurred for some new test batches, it might be the case that the `name` provided in codecov action is too long?  But one possible fix is to remove this optional name field.

PR #120, opened 2022-02-18
fix: max_rel args in evaluation

makram93:
Finetuner.fit internally calls docarray evaluation with all default metrics but with no arguments. This causes the `f1_score_at_k` and `recall_at_k` to fail as no `max_rel` argument is passed. It should be added to the kwargs at each iteration of the docs as not every doc has same number of matches

### DIFF
--- a/docarray/array/mixins/evaluation.py
+++ b/docarray/array/mixins/evaluation.py
@@ -74,8 +74,7 @@ class EvaluationMixin:
 
             binary_relevance = [1 if hash_fn(m) in desired else 0 for m in d.matches]
 
-            if 'max_rel' not in kwargs:
-                kwargs['max_rel'] = len(gd.matches)
+            kwargs['max_rel'] = len(gd.matches)
 
             r = metric_fn(binary_relevance, **kwargs)
             d.evaluations[metric_name] = NamedScore(
--- a/tests/unit/array/mixins/test_eval_class.py
+++ b/tests/unit/array/mixins/test_eval_class.py
@@ -105,3 +105,30 @@ def test_adding_noise():
 
     for d in da2:
         assert 0.0 < d.evaluations['precision_at_k'].value < 1.0
+
+
+@pytest.mark.parametrize(
+    'metric_fn, kwargs',
+    [
+        ('recall_at_k', {}),
+        ('f1_score_at_k', {}),
+    ],
+)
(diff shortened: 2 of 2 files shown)


docarray/array/mixins/evaluation.py
```diff
@@ -74,8 +74,7 @@ def evaluate(
 
             binary_relevance = [1 if hash_fn(m) in desired else 0 for m in d.matches]
 
-            if 'max_rel' not in kwargs:
-                kwargs['max_rel'] = len(gd.matches)
+            kwargs['max_rel'] = len(gd.matches)
```
JoanFM, line 77: I still think max rel needs to be respected. I think we need the logic to tell if max_rel was in kwargs from the beginning or not

JoanFM (changes requested):

docarray/array/mixins/evaluation.py
JoanFM: if we have `max_rel` in original args it means that the user assumes is the same for all `docs`

JoanFM: Otherwise u do not give this power to the user

docarray/array/mixins/evaluation.py
```diff
@@ -52,7 +52,9 @@ def evaluate(
 
         metric_name = metric_name or metric_fn.__name__
         results = []
+        global_max_rel = kwargs.pop('max_rel', None)
```
JoanFM, line 55: `caller_max_rel` seems better naming

JoanFM (approved):

docarray/array/mixins/evaluation.py
```diff
@@ -65,19 +67,17 @@ def evaluate(
                     f'Document {d!r} or {gd!r} has no matches, please check your Document'
                 )
 
-            desired = {hash_fn(m) for m in gd.matches}
-            if len(desired) != len(gd.matches):
+            targets = gd.matches[:max_rel]
+            desired = {hash_fn(m) for m in targets}
+            if len(desired) != len(targets):
                 warnings.warn(
                     f'{hash_fn!r} may not be valid, as it maps multiple Documents into the same hash. '
```
alaeddine-13, line 74: I'm not sure what is the desired behaviour here but it looks like the warning is no more reachable.

alaeddine-13 (approved):

Issue #130, opened 2022-02-22
bug: `_ipython_display_` stacking tensors/embeddings


davidbp:
When printing `DocumentArray` information in a jupyter notebook, which ends up calling `_ipython_display` which calls `summary` currently the codebase stacks embeddings/ tensors.

This does not work and provides `ValueError: all input arrays must have the same shape`

```python
from docarray import DocumentArray,Document
import numpy as np
da = DocumentArray([Document(tensor=np.zeros(3)), Document(tensor=np.zeros(4))])
da._ipython_display_()
```

but this works as expected

```python
In [4]: from docarray import DocumentArray,Document
   ...: import numpy as np
   ...: da = DocumentArray([Document(tensor=np.zeros(3)), Document(tensor=np.zeros(3))])
   ...: da._ipython_display_()
   ...: 
   ...: 
             Documents Summary             
                                           
  Length                 2                 
  Homogenous Documents   True              
  Common Attributes      ('id', 'tensor')  
                                           
                      Attributes Summary                       
                                                               
  Attribute   Data type      #Unique values   Has empty value  
 ───────────────────────────────────────────────────────────── 
  id          ('str',)       2                False            
  tensor      ('ndarray',)   2                False            
                                                               
          Storage Summary          
                                   
  Class     DocumentArrayInMemory  
  Backend   In Memory      
```

### Why this happens

When plotting to a jupyter notebook `_ipython_display_` is called which calls `summary` which calls 
`all_attrs_values = self._get_attributes(*all_attrs_names)`. If there are `tensor` or `embedding` fields then `all_attrs_names` contains them. This implies `.tensors` or `.emdeddings` can be called which will break since data can't be stacked.

### Workaround

Never call  `.tensors` or `.emdeddings` which is actually quite dangerous for big datasets because it will allocate the memory for all the vectors.

PR #132, opened 2022-02-22
feat: add method to extract metadata information

davidbp:
Refactor `_ipython_display_` to avoid https://github.com/jina-ai/docarray/issues/130 .

### problem

When plotting to a jupyter notebook `_ipython_display_` is called which calls `summary` which calls 
`all_attrs_values = self._get_attributes(*all_attrs_names)`.

- BUG 🐛: If there are `tensor` or `embedding` fields then `all_attrs_names` contains them. This implies `.tensors` or `.emdeddings` can be called which will break since data can't be stacked.
- PERFORMANCE REFACTOR: Even if the previous bug is solved it does not seem reasonable to load all embeddings in RAM just to get the length and if any of those elements is a None.

### Workaround Usability

When computing embedding stacking or .match it should be resilient to:
- `None` fields: In this case return correct result taking into account missing vectors.
- `dimension mismatch`: In this case throw and error.

### Workaround Performance

Never call  `.tensors` or `.emdeddings` which is actually quite dangerous for big datasets because it will allocate the memory for all the vectors. 

### Notes

I guess the behaviour could change in `DocumentArrayInmemory` vs the rest. Specially for `DocumentArray` with other storage backends (not memory) iterating over the data is infeasible for simply printing `DocumentArray` information.

### DIFF
--- a/docarray/array/mixins/plot.py
+++ b/docarray/array/mixins/plot.py
@@ -19,6 +19,52 @@ class PlotMixin:
         """Displays the object in IPython as a side effect"""
         self.summary()
 
+    def _update_tables_exhaustive(self, tables, all_attrs_names):
+        from rich.table import Table
+        from rich import box
+
+        attr_table = Table(box=box.SIMPLE, title='Attributes Summary')
+        attr_table.add_column('Attribute')
+        attr_table.add_column('Data type')
+        attr_table.add_column('#Unique values')
+        attr_table.add_column('Has empty value')
+
+        import ipdb
+
+        ipdb.set_trace()
+        all_attrs_values = self._get_attributes(*all_attrs_names)
+
+        if len(all_attrs_names) == 1:
+            all_attrs_values = [all_attrs_values]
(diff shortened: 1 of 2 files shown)


docarray/array/mixins/plot.py
```diff
@@ -73,30 +120,12 @@ def summary(self):
...
-                    str(tuple(_set_type_a)),
-                    str(len(_a)),
-                    str(any(_aa is None for _aa in _a)),
-                )
-            tables.append(attr_table)
+        if all_attrs_names:
+            if isinstance(self, DocumentArrayInMemory):
+                self._update_tables_exhaustive(tables, all_attrs_names)
+            else:
+                self._update_tables_information(tables, all_attrs_names)
```
alaeddine-13, line 128: I believe we should avoid these checks and rely on inheritance and overriding methods (and if we apply the suggestion, the method should just return information and shouldn't handle the tables I guess)

Issue #128, opened 2022-02-21
Add protocol and compression info DocumentArray filename


davidbp:
When storing a `DocumentArray` there is a protocol and compression used. This implies that users need to remember how they serialized a DocumentArray to be able to load it back from binary. 

We can add this information in the string of the filename of the serialiized docarray and use it when loading from binary.

PR #138, opened 2022-02-23
feat: add compression and protocol in file name

davidbp:
Now there is the issue that if a `DocumentArray` is stored as binary users need to remember compression and protocol to load it back.

This PR adds this information as extensions in the file.

Example
```python
da = DocumentArray([Document(tensor=np.array([1,2,3])),  Document(tensor=np.array([7,8,9]))])
da.save_binary( 'my_docarray.protobuf.lz4')
```
will save
```bash
my_docarray.protobuf.lz4
```
Thefore,  later on users don't need to recall which `protocol` and `compress` were used. 
Since the information is in the string this PR parses it so that users don't need to pass this info as keyword arguments to `load_binary`. They can simply do:

```python
da.load_binary('my_docarray.protobuf.lz4')
```

This PR solves https://github.com/jina-ai/docarray/issues/128

### DIFF
--- a/docarray/array/mixins/io/binary.py
+++ b/docarray/array/mixins/io/binary.py
@@ -1,6 +1,7 @@
 import base64
 import io
 import os.path
+import os
 import pickle
 from contextlib import nullcontext
 from typing import Union, BinaryIO, TYPE_CHECKING, Type, Optional, Generator
@@ -10,7 +11,6 @@ from ....helper import (
     get_compress_ctx,
     decompress_bytes,
     protocol_and_compress_from_file_path,
-    add_protocol_and_compress_to_file_path,
 )
 
 if TYPE_CHECKING:
@@ -42,7 +42,15 @@ class BinaryIOMixin:
         :param streaming: if `True` returns a generator over `Document` objects.
         In case protocol is pickle the `Documents` are streamed from disk to save memory usage
         :return: a DocumentArray object
+
+        .. note::
+            If `file` is `str` it can specify `protocol` and `compress` as file extensions.
+            This functionality assumes `file=file_name.$protocol.$compress` where `$protocol` and `$compress` refer to a
+            string interpolation of the respective `protocol` and `compress` methods.
+            For example if `file=my_docarray.protobuf.lz4` then the binary data will be loaded assuming `protocol=protobuf`
+            and `compress=lz4`.
         """
+
         if isinstance(file, io.BufferedReader):
             file_ctx = nullcontext(file)
         elif isinstance(file, bytes):
@@ -62,6 +70,7 @@ class BinaryIOMixin:
                 _show_progress=_show_progress,
             )
         else:
+            print(protocol, compress)
             return cls._load_binary_all(
                 file_ctx, protocol, compress, _show_progress, *args, **kwargs
             )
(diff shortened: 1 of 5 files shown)


docarray/helper.py
```diff
@@ -321,3 +322,69 @@ def dataclass_from_dict(klass, dikt):
...
+    ('protobuf', 'gzip')
+
+    >>> protocol_and_compress_from_file_path('/Documents/docarray_fashion_mnist.protobuff')
+    ('protobuf', None)
+
+    >>> protocol_and_compress_from_file_path('/Documents/docarray_fashion_mnist.gzip')
+    (None, gzip)
+    """
+
+    ALLOWED_PROTOCOLS = {'pickle', 'protobuf', 'protobuf-array', 'pickle-array'}
```
JoanFM, line 350: I think this should be defined somewhere more globally to make sure is better mantainable?

docarray/helper.py
```diff
@@ -321,3 +322,69 @@ def dataclass_from_dict(klass, dikt):
         if isinstance(dikt, (tuple, list)):
             return [dataclass_from_dict(klass.__args__[0], f) for f in dikt]
         return dikt
+
+
+def protocol_and_compress_from_file_path(
+    file_path: str,
+    default_protocol: Optional[str] = None,
```
JoanFM, line 329: what if it is not found and this is None, will it fail?

davidbp: No, since `load_binary` have default values.

docarray/helper.py
davidbp: Agree, maybe in `docarray/array/mixins/io/binary.py`

numb3r3 (changes requested):

tests/unit/test_helper.py
```diff
@@ -0,0 +1,48 @@
...
+    if protocol:
+        file_path_extended += '.' + protocol
+    if compress:
+        file_path_extended += '.' + compress
+
+    protocol, compress = protocol_and_compress_from_file_path(file_path_extended)
+    if protocol:
+        assert protocol in ['protobuf', 'protobuf-array', 'pickle', 'pickle-array']
+    if compress:
+        assert compress in ['lz4', 'bz2', 'lzma', 'zlib', 'gzip']
```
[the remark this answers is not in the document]
davidbp: both things should be tested, just added now

numb3r3 (approved):
LGTM 👍

docarray/array/mixins/io/binary.py
```diff
@@ -191,18 +208,33 @@ def save_binary(
...
         """
         if isinstance(file, io.BufferedWriter):
             file_ctx = nullcontext(file)
         else:
+            _protocol, _compress = protocol_and_compress_from_file_path(file)
+
+            if _protocol is not None:
+                protocol = _protocol
+            if _compress is not None:
+                compress = _compress
```
alaeddine-13, line 236: I think we should enforce the presence of both, protocol and compress, in order to use any of them, as documented.

docarray/helper.py
```diff
@@ -321,3 +325,66 @@ def dataclass_from_dict(klass, dikt):
...
+
+    protocol = default_protocol
+    compress = default_compress
+
+    file_extensions = [e.replace('.', '') for e in pathlib.Path(file_path).suffixes]
+    for extension in file_extensions:
+        if extension in ALLOWED_PROTOCOLS:
+            protocol = extension
+        elif extension in ALLOWED_COMPRESSIONS:
+            compress = extension
```
alaeddine-13, line 361:
```suggestion
    m = re.match('^.*\.([^.]+)\.([^.]+)$', file_path)
    if m and m.groups() == 2 and m.group(1) in ALLOWED_PROTOCOLS and m.group(2) in ALLOWED_COMPRESSIONS:
        protocol, compress = m.group(1), m.group(2)
```

docarray/array/mixins/io/binary.py
davidbp: you can decide to use No compression

docarray/helper.py
davidbp: You can decide no compression. I'm allowing the order not to be relevant but the values that is why I did it how I did.

PR #141, opened 2022-02-24
refactor(graphql): strawberry to input

samsja:
We need both strawberry `input` and `type` for `DocumentArray`.

Indeed in strawberry [input](https://strawberry.rocks/docs/types/input-types) and [type](https://strawberry.rocks/docs/types/object-types) are two differents concepts (even if from a code perspective they are very similar).

Briefly : 

* `type` is used to select fields in request / response
* ` input` is used  when "posting" a document. (We need it in core because we always rely on posting a da when calling an endpoint)

So core need both definition in `docarray`. Moreover other users working with docarray might need both as well.

Right now we defined `StrawberryDocumentInput` without correlation with `StrawberryDocument`(the type). So some code is repeated. This is normal, both objects and semantically different even if today they are almost equal it could be different in the future. ( see this  [commit](https://github.com/jina-ai/docarray/pull/114/commits/ecc648866cfd354e3461cd94d1b1c7126b17ba99) where I unified both that I finally revert)

original conversion [here](https://github.com/jina-ai/docarray/pull/114)

done with @JohannesMessner

### DIFF
--- a/docarray/document/mixins/strawberry.py
+++ b/docarray/document/mixins/strawberry.py
@@ -13,7 +13,7 @@ class StrawberryMixin:
     def to_strawberry_type(self) -> 'StrawberryDocument':
         """Convert a Document object into a Strawberry type."""
         from ..strawberry_type import StrawberryDocument as SD
-        from ..strawberry_type import NameScoreItem, NamedScore
+        from ..strawberry_type import _NameScoreItem, _NamedScore
 
         _p_dict = {}
         for f in self.non_empty_fields:
@@ -22,7 +22,7 @@ class StrawberryMixin:
                 _p_dict[f] = v.to_strawberry_type()
             elif f in ('scores', 'evaluations'):
                 _p_dict[f] = [
-                    NameScoreItem(k, NamedScore(**v.to_dict())) for k, v in v.items()
+                    _NameScoreItem(k, _NamedScore(**v.to_dict())) for k, v in v.items()
                 ]
 
             else:
--- a/docarray/document/mixins/strawberry.py
+++ b/docarray/document/mixins/strawberry.py
@@ -13,7 +13,7 @@ class StrawberryMixin:
     def to_strawberry_type(self) -> 'StrawberryDocument':
         """Convert a Document object into a Strawberry type."""
         from ..strawberry_type import StrawberryDocument as SD
-        from ..strawberry_type import _NameScoreItem, _NamedScore
+        from ..strawberry_type import NameScoreItem, NamedScore
 
         _p_dict = {}
         for f in self.non_empty_fields:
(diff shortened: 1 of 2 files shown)


hanxiao (changes requested):
```python
class _StrawberryDocument:
    id: Optional[str] = None
    parent_id: Optional[str] = None
    granularity: Optional[int] = None
    adjacency: Optional[int] = None
    blob: Optional[Base64] = None
    tensor: Optional[NdArray] = None
    mime_type: Optional[str] = None
    text: Optional[str] = None
    weight: Optional[float] = None
    uri: Optional[str] = None
    tags: Optional[JSONScalar] = None
    offset: Optional[float] = None
    location: Optional[List[float]] = None
    embedding: Optional[NdArray] = None
    modality: Optional[str] = None
    evaluations: Optional[List[_NameScoreItem]] = None
    scores: Optional[List[_NameScoreItem]] = None
    chunks: Optional[List['StrawberryDocument']] = None
    matches: Optional[List['StrawberryDocument']] = None

@strawberry.type
class StrawberryDocument(_StrawberryDocument):
    ...

@strawberry.input
class StrawberryDocumentInput(_StrawberryDocument):
    ...
```

hanxiao (dismissed):
refer to `tests/unit/test_strawberry.py` for the test

samsja:
> ```python
> class _StrawberryDocument:
>     id: Optional[str] = None
>     parent_id: Optional[str] = None
>     granularity: Optional[int] = None
>     adjacency: Optional[int] = None
>     blob: Optional[Base64] = None
>     tensor: Optional[NdArray] = None
>     mime_type: Optional[str] = None
>     text: Optional[str] = None
>     weight: Optional[float] = None
>     uri: Optional[str] = None
>     tags: Optional[JSONScalar] = None
>     offset: Optional[float] = None
>     location: Optional[List[float]] = None
>     embedding: Optional[NdArray] = None
>     modality: Optional[str] = None
>     evaluations: Optional[List[_NameScoreItem]] = None
>     scores: Optional[List[_NameScoreItem]] = None
>     chunks: Optional[List['StrawberryDocument']] = None
>     matches: Optional[List['StrawberryDocument']] = None
> 
> @strawberry.type
> class StrawberryDocument(_StrawberryDocument):
>     ...
> 
> @strawberry.input
> class StrawberryDocumentInput(_StrawberryDocument):
>     ...
> ```

I don't agree that we should have one source of truth for both. The input kind is more like an init function we don't want the user to be allowed to override some private field. In the future we could want to delete some field from the input and keep them in the type. 

For me the StrawberryDocument is like the pydantic models and the StrawberryDocumentInput is the public interface of the Document.

I first stared to join them in this commit https://github.com/jina-ai/docarray/pull/141/commits/69ac6bb2ab7900d4c1d90c2a03371acfdbedbd92 but I revert it later
 

Moreover

JoanFM (approved):

PR #142, opened 2022-02-24
fix(array): fix text getter

hanxiao:
- text getter
- plot local audio/video file (e.g., `../data/helllo.mp4`) by `display()`


JoanFM (changes requested):
Let's add the test

KunalVaghani90:
hi

PR #143, opened 2022-02-24
feat: return iterable on @ selector 

alaeddine-13:
**Do not merge**
This PR is only for testing purposes


JoanFM (changes requested):
I would suggest then to create an abstract class:

```python
class DocumentIterable(Iterable[Document]):
 ...

```

that implements all the methods that do not to rely on the `uniqueness` of `IDS`. Like this there will be much better clearer of what is what Plus many of the accessors could be shared.

hanxiao:
close due to #239

PR #144, opened 2022-02-24
fix(pushpull): raise error if push failed

delgermurun:
* also added ability change server host

related https://github.com/jina-ai/hubble/issues/301


numb3r3 (changes requested):
Let's add a test to cover this change.

delgermurun:
@numb3r3 

> Let's add a test to cover this change.

Added tests.

numb3r3 (approved):
LGTM 👍

PR #150, opened 2022-02-28
feat: query language supporting filter

numb3r3:
The example usages code:

```python

docs = DocumentArray.empty(5)
docs[0].text = 'hello'
docs[1].text = 'world'
docs[2].tags['x'] = 0.3
docs[2].tags['y'] = 0.6
docs[3].tags['x'] = 0.8

# NOTE: you can use `find` instead of `_filter`

result = docs.find({'text': {'$eq': 'hello'}})
print(DocumentArray(result).texts)

result = docs.find({'tags__x': {'$gte': 0.5}})
print(DocumentArray(result)[:, 'tags'])

result = docs.find({'$or': {'tags__x': {'$gte': 0.1}, 'tags__y': {'$gte': 0.5}}})
print(DocumentArray(result)[:, 'tags'])

result = docs.find({'tags__x': {'$gte': 0.1, '$lte': 0.5}})
print(DocumentArray(result)[:, 'tags'])

result = docs.find({'$and': {'tags__x': {'$gte': 0.1}, 'tags__y': {'$gte': 0.5}}})
print(DocumentArray(result)[:, 'tags'])

result = docs.find({'text': {'$eq': '{tags__name}'}})
assert result[0].id == docs[0].id
```

TODO:
- [ ] documentation

### DIFF
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -159,7 +159,7 @@ class FindMixin:
     ) -> Tuple['np.ndarray', 'np.ndarray']:
         raise NotImplementedError
 
-    def filter(self, *args, **kwargs):
+    def _filter(self, *args, **kwargs):
         from ... import DocumentArray
         from ...lookup import Q, filter_items
 
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -159,9 +159,10 @@ class FindMixin:
     ) -> Tuple['np.ndarray', 'np.ndarray']:
         raise NotImplementedError
 
-    def _filter(self, *args, **kwargs):
+    def _filter(self, query: Dict, **kwargs):
         from ... import DocumentArray
-        from docarray.array.queryset.lookup import filter_items
+        from docarray.array.queryset.parser import QueryParser
 
-        result = filter_items(self, *args, **kwargs)
-        return DocumentArray(result)
+        parser = QueryParser(query)
+
+        return DocumentArray([d for d in self if parser.evaluate(d)])
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -158,3 +158,10 @@ class FindMixin:
         self, query: 'ArrayType', limit: int, **kwargs
     ) -> Tuple['np.ndarray', 'np.ndarray']:
         raise NotImplementedError
+
+    def filter(self, *args, **kwargs):
+        from ... import DocumentArray
+        from ...lookup import Q, filter_items
+
+        result = filter_items(self, *args, **kwargs)
+        return DocumentArray(result)
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -36,7 +36,7 @@ class FindMixin:
 
     def find(
         self: 'T',
-        query: Union['DocumentArray', 'Document', 'ArrayType'],
+        query: Union['DocumentArray', 'Document', 'ArrayType', Dict],
         metric: Union[
             str, Callable[['ArrayType', 'ArrayType'], 'np.ndarray']
         ] = 'cosine',
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -161,7 +161,7 @@ class FindMixin:
 
     def _filter(self, *args, **kwargs):
         from ... import DocumentArray
-        from ...lookup import Q, filter_items
+        from docarray.array.queryset.lookup import filter_items
 
         result = filter_items(self, *args, **kwargs)
         return DocumentArray(result)
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -68,8 +68,10 @@ class FindMixin:
             else:
                 limit = int(limit)
 
-        _limit = len(self) if limit is None else (limit + (1 if exclude_self else 0))
+        if isinstance(query, dict):
+            return self._filter(query, limit=limit, only_id=only_id)
 
+        _limit = len(self) if limit is None else (limit + (1 if exclude_self else 0))
         if isinstance(query, (DocumentArray, Document)):
 
             if isinstance(query, Document):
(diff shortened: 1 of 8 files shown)


docarray/array/mixins/find.py
```diff
@@ -158,3 +158,11 @@ def _find(
         self, query: 'ArrayType', limit: int, **kwargs
     ) -> Tuple['np.ndarray', 'np.ndarray']:
         raise NotImplementedError
+
+    def _filter(self, query: Dict, **kwargs):
```
JoanFM, line 162: This should be public right?

JoanFM (changes requested):

docarray/array/mixins/find.py
numb3r3: you can use `find` method now

docarray/array/queryset/lookup.py
```diff
@@ -0,0 +1,219 @@
+from typing import TYPE_CHECKING
```
JoanFM, line 1: keep the license agreement from original source on top of the file

JoanFM (changes requested):

docarray/array/mixins/find.py
```diff
@@ -36,7 +36,7 @@ def find(self: 'T', query: str, **kwargs):
 
     def find(
         self: 'T',
-        query: Union['DocumentArray', 'Document', 'ArrayType'],
+        query: Union['DocumentArray', 'Document', 'ArrayType', Dict],
```
JoanFM, line 39: I do not think this should be the query, we should at least prepare the interface fo be able to accept both arguments, query and filters

numb3r3: I got your point. But we can use another argument, e.g., `pre_filters`. For example, in annlite,
```
docs = DocumentArray(storage='annlite')

docs.find(query, pre_filters={'tags__x': {'$gte': 1.0}})
```
And the query language of `pre_filtering` only depends on the backend.

JoanFM: but isn't it weird to see that a `filter` can be put on both sides?

docarray/array/queryset/lookup.py
```diff
@@ -0,0 +1,248 @@
...
+        >>> lookup('text__exact', 'hello', doc)
+
+    The above will return True if doc.text == 'hello' else False
+
+    :param key  : the field name to find
+    :param val  : object to match the value in the document against
+    :param doc : the document to match
+    """
+    get_key, last = dunder_partition(key)
+    value = doc._get_attributes(get_key)
```
alaeddine-13, line 53:
```suggestion
    get_key, last = dunder_partition(key)
    value = doc._get_attributes(get_key)

    regex = re.compile(r'\$\{\{\s*([a-zA-Z0-9_]*)\s*}}')
    m = regex.match(val)
    if isinstance(val, str) and m:
        val_key = regex.sub(lambda matchobj: matchobj.group(1), val)
        val = doc._get_attributes(val_key)
```

should allow this syntax:
`da._filter({'tags__a': {'$gte': '${{tags__c}}'}})`

numb3r3: Like this idea, and will add it soon

numb3r3: Instead use `${{...}}` syntax, I choose to use pythonic-style placeholder syntax  `{...}` 
```
da._filter({'tags__a': {'$gte': '{tags__c}'}})
```

docarray/array/mixins/find.py
numb3r3: @JoanFM I think using `find` would be more reasonable. for your above question, i think this would not be a big deal with well-documented.

docarray/array/mixins/find.py
```diff
@@ -158,3 +160,22 @@ def _find(
         self, query: 'ArrayType', limit: int, **kwargs
     ) -> Tuple['np.ndarray', 'np.ndarray']:
         raise NotImplementedError
+
+    def _filter(
+        self,
+        query: Dict,
+        limit: Optional[Union[int, float]] = None,
```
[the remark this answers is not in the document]
numb3r3: in jina flow, the `limit` sometimes comes from `parameters` (e.g,. TagStructureView). Then, all of the integers in `parameters` will be enforced cast to float. But I agree that it is kinda weird here to use limit as float, I think it can be removed.

docarray/array/queryset/lookup.py
```diff
@@ -0,0 +1,265 @@
...
+    will return True if doc.text == doc.tags['name'] else False
+
+    :param key  : the field name to find
+    :param val  : object to match the value in the document against
+    :param doc : the document to match
+    """
+    get_key, last = dunder_partition(key)
+
+    if isinstance(val, str) and val.startswith('{'):
+        r = PLACEHOLDER_PATTERN.findall(val)
```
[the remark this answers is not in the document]
numb3r3: Can I get some clues about the reason behind to use `.match`

docarray/array/mixins/find.py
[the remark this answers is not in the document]
numb3r3: oh, I'm worried that this would break the executors in hub. I will keep it.

docarray/array/mixins/find.py
```diff
@@ -158,3 +160,22 @@ def _find(
...
+        **kwargs,
+    ):
+        from ... import DocumentArray
+        from ..queryset import QueryParser
+
+        parser = QueryParser(query)
+
+        result = DocumentArray()
+        for d in self:
+            if (limit is None or len(result) < limit) and parser.evaluate(d):
```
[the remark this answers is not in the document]
numb3r3: addressed

docarray/array/queryset/lookup.py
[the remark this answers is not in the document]
davidbp: I don't see any problem that `regex.match` would bring much benefit. LGTM.

PR #158, opened 2022-02-28
fix: support torch.Tensor and other embeddings in match

alaeddine-13:
fix: support torch.Tensor and other embeddings in match


davidbp (approved):
LGTM

PR #160, opened 2022-03-01
feat: add from to dict io

davidbp:
this PR implements https://github.com/jina-ai/docarray/issues/156

### DIFF
--- a/docarray/array/mixins/io/dict.py
+++ b/docarray/array/mixins/io/dict.py
@@ -24,7 +24,8 @@ class DictIOMixin:
         """
         from .... import Document, DocumentArray
 
-        da = DocumentArray(**kwargs).empty(len(input_dict))
+        da = cls(*args, **kwargs).empty(len(input_dict))
+        # da = DocumentArray(**kwargs).empty(len(input_dict))
 
         for offset, d in input_dict.items():
             da[offset] = Document(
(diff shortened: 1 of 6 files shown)


docarray/array/mixins/io/dict.py
```diff
@@ -0,0 +1,33 @@
...
+class DictIOMixin:
+    """Save/load a DocumentArray into a dict of the form `{offset_0: doc_0, offset_1: doc_1, ...}`"""
+
+    def to_dict(self, protocol: str = 'jsonschema', **kwargs) -> List:
+        """Convert the object into a Python dict of the form `{offset_0: doc_0, offset_1: doc_1, ...}`
+
+        :param protocol: `jsonschema` or `protobuf`
+        :return: a Python list
+        """
+        return {k: d.to_dict(protocol=protocol, **kwargs) for k, d in enumerate(self)}
```
hanxiao, line 16:
```suggestion
        return {d.id: d.to_dict(protocol=protocol, **kwargs) for d in self}
```

davidbp:
> dict should be indexed by id

If we index by id, we have the id in both the keys and the values (so we have redundant info). Besides, we loose the position in the original docarray, which means we can't restore it back, or we can't ensure the possitions will be correct.

Issue #157, opened 2022-02-28
only in `.match/.find` make `.embeddings` raise error if it contains None element


davidbp:

hanxiao:
decided to drop this feature

PR #163, opened 2022-03-01
feat: add none check to find

davidbp:
This PR solves https://github.com/jina-ai/docarray/issues/157,

Note this is only done for the in memory case, since cases where storage is on disk will make this unreasonably slow,. This needs to iterate over the data and check for each doc that embedding is not None.

```python
q = Document(embedding=np.array([1,2,3]))
da2 = DocumentArray([Document(embedding=np.array([1,2,3])), 
                    Document(),
                    Document(embedding=np.array([7,8,9]))])
q.match(da2)
```
```
TypeError: unsupported operation, one of the embeddings is None.
```

### DIFF
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -3,6 +3,7 @@ from typing import overload, Optional, Union, Dict, List, Tuple, Callable, TYPE_
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


hanxiao:
decide to drop this feature

PR #171, opened 2022-03-04
fix: correct file not found error

davidbp:
This PR raises `FileNotFoundError` when a file that is not valid is passed as input to `.load_binary`.

### DIFF
--- a/docarray/array/mixins/io/binary.py
+++ b/docarray/array/mixins/io/binary.py
@@ -54,11 +54,15 @@ class BinaryIOMixin:
             file_ctx = nullcontext(file)
         elif isinstance(file, bytes):
             file_ctx = nullcontext(file)
-        elif os.path.exists(file):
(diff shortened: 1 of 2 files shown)


docarray/array/mixins/io/binary.py
```diff
@@ -54,11 +54,15 @@ def load_binary(
...
-            )
-            file_ctx = open(file, 'rb')
+        elif isinstance(file, str):
+            if os.path.exists(file):
+                protocol, compress = protocol_and_compress_from_file_path(
+                    file, protocol, compress
+                )
+                file_ctx = open(file, 'rb')
+            else:
+                if not os.path.exists(file):
```
hanxiao, line 64: looks redundant

davidbp: totally redundant

hanxiao (approved):

PR #187, opened 2022-03-10
feat: collate function in embed function

numb3r3:
With this PR, users can create custom mini-batch inputs for different embeded model by passing `collate_fn`
```python
import functools

docs = DocumentArray([Document(text='hello world')] * 8)

from transformers import BertTokenizer, BertModel, TFBertModel

def collate_fn(da, tokenizer = None, return_tensors: str = 'pt'):
    texts = da.texts
    return tokenizer(texts, return_tensors= return_tensors)

tokenizer = BertTokenizer.from_pretrained('bert-base-uncased')

torch_model = BertModel.from_pretrained("bert-base-uncased")
docs.embed(torch_model, collate_fn=functools.partial(collate_fn, tokenizer=tokenizer, return_tensors='pt'))
print(docs.embeddings.shape)

tf_model = TFBertModel.from_pretrained("bert-base-uncased")
docs.embed(tf_model, collate_fn=functools.partial(collate_fn, tokenizer=tokenizer, return_tensors='tf'))
print(docs.embeddings.shape)
```

**Checklist**:
- [x] torch framework
- [x] keras framework
- [x] paddle framework
- [x] onnx framework
- [x] add unitest


docarray/array/mixins/embed.py
```diff
@@ -11,29 +17,40 @@ class EmbedMixin:
...
             `cpu` or `cuda`.
         :param batch_size: number of Documents in a batch for embedding
         :param to_numpy: if to store embeddings back to Document in ``numpy.ndarray`` or original framework format.
         :return: itself after modified.
         """
 
+        if collate_fn is None:
+
+            def default_collate_fn(da: 'DocumentArray'):
+                return da.tensors
```
bwanglzu, line 40: i think the default collate function should convert `tensors` property to `tensor` objects, such as pytorch tensor.

docarray/array/mixins/embed.py
```diff
@@ -11,29 +17,40 @@ class EmbedMixin:
     def embed(
         self: 'T',
         embed_model: 'AnyDNN',
+        collate_fn: Optional['CollateFnType'] = None,
```
bwanglzu, line 20: i would move it to the last variable

bwanglzu (changes requested):
left some comments, the implementation looks good, we need to add tests (to verify). You can refer to [this](https://github.com/jina-ai/finetuner.fit/blob/main/tests/unit/test_embedding.py#L40) one.

I know `collate_fn` is a syntax in Pytorch and hugging face transformers. I'm not sure about tensorflow and paddle, how their users process the tokens. I think if we support `transformers`, we are ready for 95% of this use case.

docarray/array/mixins/embed.py
numb3r3: any reason? because you think `collate_fn` should be not be considered as frequently used param atm?

numb3r3:
@bwanglzu some commends are addressed, please check it again, thx!

docarray/array/mixins/embed.py
```diff
@@ -93,8 +126,15 @@ def _set_embeddings_paddle(
         embed_model.to(device=device)
         embed_model.eval()
         for b_ids in self.batch_ids(batch_size):
-            batch_inputs = paddle.to_tensor(self[b_ids, 'tensor'], place=device)
-            r = embed_model(batch_inputs)
+            batch_inputs = collate_fn(self[b_ids])
+            if isinstance(batch_inputs, Mapping):
```
bwanglzu, line 130: this `batch_inputs` in Pytorch and Tensorflow are hugging face `transformer` `BatchEncoding` and `BatchFeature` classes, see [here](https://github.com/jina-ai/finetuner.fit/blob/main/finetuner/device.py#L21), I'm not sure how paddle resolve this.

bwanglzu (commented):
TFBert and Bert looks good to me, I'm not sure about paddle paddle.

tests/unit/array/mixins/test_embed.py
```diff
@@ -111,3 +117,52 @@ def test_embedding_on_random_network(
...
+
+def test_embeded_paddle_model(paddle_model):
+    def collate_fn(da):
+        return {'x': da.tensors, 'y': da.tensors}
+
+    docs = DocumentArray.empty(3)
+    docs.tensors = np.random.random([3, 5]).astype(np.float32)
+    docs.embed(paddle_model, collate_fn=collate_fn, to_numpy=True)
+    assert (docs.tensors == docs.embeddings).all()
+    # assert np.testing.assert_array_almost_equal(docs.tensors, docs.embeddings)
```
bwanglzu, line 141: this needs to be uncommented?

docarray/array/mixins/embed.py
```diff
@@ -22,18 +29,28 @@ def embed(
...
         getattr(self, f'_set_embeddings_{fm}')(
-            embed_model, device, batch_size, to_numpy
+            embed_model, collate_fn, device, batch_size, to_numpy
         )
         return self
 
     def _set_embeddings_keras(
         self: 'T',
         embed_model: 'AnyDNN',
+        collate_fn: 'CollateFnType',
```
bwanglzu, line 53: maybe we move all `collate_fn` to the last to be consistent? the reason is it's not frequently used as other arguments.

docarray/array/mixins/embed.py
numb3r3: same in paddle, this is not only about `transformers`. this is implemented for kwargs parameters in `forward` method

tests/unit/array/mixins/test_embed.py
numb3r3: `np.testing.assert_array_almost_equal` cannot work here. I have no idea what's the problem.

docarray/array/mixins/embed.py
numb3r3: in the `_set_embeddings_{fm}` function, the `collate_fn` is designed as positional parameter (as it is required always)

docarray/array/mixins/embed.py
numb3r3: and of course, it can't work at all cases. For example, the current implementation cannot work with `sentence_transformers` since its forward method accepts *dict* as the whole input.

bwanglzu (approved):
LGTM!

PR #188, opened 2022-03-10
feat: multi modal document

alaeddine-13:
feat: multi modal document

- [x] add type annotations
- [x] add from_dataclass method
- [x] store schema in root document meta tags (possible also move other attributes to meta tags)
- [x] build a sub attribute accessor using schema stored in root document
- [x] extend traversal path syntax
- [x] protobuf serialization/deserialization
- [x] modifications to traversal path grammar:
* [] for attributes **WARNING: this entails that the grammer becomes ambiguous**
* [] optional for slice
* tolerate white space
* accept offset index not just slice
- [x] test protobuf serialization/deserialization
- [x] extend dataclass decorator + add method from_document to decorated classes [optional]
- [ ] test coverage

Usage:
create schema:
```python
from docarray.types import dataclass, Text, Image, Audio
@dataclass
class MMDocument:
    title: Text
    image: Image
    version: int
```

create mm_doc:
```python
obj = MMDocument(title='hello world', image='img.png'), version=20)
```

translate to Document:
```python
doc = Document.from_dataclass(obj)
```

traverse a DocumentArray of mm_docs:
```python
da = DocumentArray([obj, ...])
title_docs = da['@r.[title]']
image_docs = da['@r.[image]']
image_title_docs = da['@r.[image,title]']
```
traverse with slice:
```python
da = DocumentArray([obj, ...])
first_3_title_docs = da['@r[:3].[title]']
first_image_docs = da['@r[1].[image]']
last_image_title_docs = da['@r[-1].[image,title]']
```

**Note**: the syntax of traversal paths is extended, there should be no breaking changes (e.g, old `r1:5c:3` syntax is still accepted)


docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,44 @@
...
+                    else:
+                        chunk = Document()
+                        for element in attribute:
+                            chunk.chunks.append(cls.from_dataclass(element))
+
+            elif field.type in [str, int, float]:
+                tags[key] = attribute
+
+        # TODO: may have to modify this?
+        root.tags = tags
```
numb3r3, line 42: we can use `tags` to store some global meta data, e.g., `tags['__global__meta__'] = {'image_field_loc': 0}`

winstonww: Agree, storing metadata to allow converting a `Document` back to a dataclass is a great idea

winstonww (commented):
I also wonder what's the best way to handle optional parameters? For instance, for `img: Optional[ImageDocument]`, should we append an empty chunk document or not in this case?

docarray/types.py
```diff
@@ -11,6 +11,10 @@
     Tuple,
 )
 
+import numpy as np
+import tensorflow
+import torch
```
numb3r3, line 16: move them under the `TYPE_CHECKING`? We don't want to include them as the dependencies of docarray.

numb3r3 (changes requested):

docarray/types.py
alaeddine-13: done

docarray/proto/docarray.proto
```diff
@@ -120,6 +120,9 @@ message DocumentProto {
   // Evaluations performed on the document, each element corresponds to a metric
   map<string, NamedScoreProto> evaluations = 19;
 
+  // system-defined meta attributes represented in a structured data value.
+  google.protobuf.Struct _metadata = 20;
```
numb3r3, line 124: Please also be aware that the integer values in the `google.protobufStruct` will be cast to float values automatically when serialize/deserialize to/from protobuf.  As shown in current document tags:

```
from docarray import Document

d = Document(tags={'x': 1})

pb = d.to_protobuf(d)
d = Document.from_protobuf(pb)
print(d.tags)
```
returns
```
{'x': 1.0}
```

alaeddine-13: thanks for pointing this out. Actually I didn't include tests for protobuf serialization/deserialization, will do so soon

docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,134 @@
...
+            raise ValueError('Object is not a dataclass instance')
+
+        from docarray import Document
+
+        root = Document()
+        tags = {}
+        multi_modal_schema = {}
+        for key, field in obj.__dataclass_fields__.items():
+            attribute = getattr(obj, key)
+            if field.type in [str, int, float]:
```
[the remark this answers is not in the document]
alaeddine-13: supported now

docarray/array/mixins/traverse.py
```diff
@@ -42,7 +105,8 @@ def traverse(
             - `r,c`: docs in this TraversableSequence and all child-documents at granularity 1
 
         """
-        for p in traversal_paths.split(','):
```
JoanFM, line 45: I guess there must be some update of the `docstrings`

docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,134 @@
...
+        for key, field in obj.__dataclass_fields__.items():
+            attribute = getattr(obj, key)
+            if field.type in [str, int, float, bool]:
+                tags[key] = attribute
+                multi_modal_schema[key] = {
+                    'attribute_type': AttributeType.PRIMITIVE,
+                    'type': field.type.__name__,
+                }
+
+            elif isinstance(field.type, typing._GenericAlias):
```
JoanFM, line 41: what does this mean?

docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,134 @@
...
+                if field.type._name in ['List', 'Iterable']:
```
JoanFM, line 42: isn't safer to check the `instance` and check if one can iterate?

docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,134 @@
...
+            AttributeType.ITERABLE_NESTED,
+        ]:
+            return self.chunks[position].chunks
+        else:
+            raise ValueError(
+                f'Invalid attribute {attribute}: must a Document attribute or nested dataclass'
+            )
+
+    @classmethod
+    def _from_obj(cls, obj, obj_type) -> typing.Tuple['Document', AttributeType]:
```
JoanFM, line 118: it is weird to need `obj` and `obj_type`, why not understand it here?

docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,134 @@
...
+
+        attribute_type = AttributeType.DOCUMENT
+
+        if is_dataclass(obj_type):
+            doc = cls.from_dataclass(obj)
+            attribute_type = AttributeType.NESTED
+        elif obj_type == ImageDocument:
+            doc = Document(tensor=obj, modality='image')
+        elif obj_type == BlobDocument:
+            doc = Document(blob=obj, modality='blob')
```
JoanFM, line 129: if we use `modality` like this, I would simply remove `modality`.

JoanFM (changes requested):

docarray/document/mixins/multimodal.py
alaeddine-13: it handles types like `List[type]`

docarray/document/mixins/multimodal.py
alaeddine-13: we're inferring the object type from the type annotation, not from the object instance.
For example, in the case of ImageDocument, the actual type is a numpy array, while the type annotation is ImageDocument which is a typevar that also covers np.array

docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,134 @@
...
+    PRIMITIVE = 2
+    ITERABLE_PRIMITIVE = 3
+    ITERABLE_DOCUMENT = 4
+    NESTED = 5
+    ITERABLE_NESTED = 6
+
+
+class MultiModalMixin:
+    @classmethod
+    def from_dataclass(cls, obj):
```
numb3r3, line 23: So the `from_dataclass` can also be applied to the `Document` class? Can we restrict this function to `MultiModalDocument` only ?

docarray/document/mixins/multimodal.py
numb3r3: Is `obj_type` parameter necessary? we can use `isinstance()` to validate the object type.

docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,134 @@
...
+        if is_dataclass(obj_type):
+            doc = cls.from_dataclass(obj)
+            attribute_type = AttributeType.NESTED
+        elif obj_type == ImageDocument:
+            doc = Document(tensor=obj, modality='image')
+        elif obj_type == BlobDocument:
+            doc = Document(blob=obj, modality='blob')
+        elif obj_type == TextDocument:
+            doc = Document(text=obj, modality='text')
+        else:
```
numb3r3, line 132: add `elif obj_type == Document`?

docarray/proto/io/__init__.py
```diff
@@ -22,6 +22,8 @@ def parse_proto(pb_msg: 'DocumentProto') -> 'Document':
             fields[f_name] = [Document.from_protobuf(d) for d in value]
         elif isinstance(value, NdArrayProto):
             fields[f_name] = read_ndarray(value)
+        elif isinstance(value, Struct) and f_name == '_metadata':
+            fields['_metadata'] = MessageToDict(value, preserving_proto_field_name=True)
```
numb3r3, line 26: This block is redundant with lines of 27-28, can be removed

alaeddine-13: fixed

docarray/document/mixins/multimodal.py
alaeddine-13: the typing module doesn't offer a `Document` type annotation.
If we're supposed to offer such an annotation, how do you suggest it should work ?

docarray/types.py
```diff
@@ -61,3 +64,106 @@
...
+    ForwardRef('torch.Tensor'),
+)
+
+BlobDocument = TypeVar(
+    'BlobDocument',
+    ForwardRef('np.ndarray'),
+    bytes,
+)
+
+TextDocument = TypeVar('TextDocument', bound=str)
```
hanxiao, line 81: as discussed in the last alignment, this needs to be changed

docarray/document/mixins/multimodal.py
numb3r3: I'm also not quite clear about it. Does it make sense to offer an annotation of `Document`?

docarray/array/mixins/traverse.py
alaeddine-13: updated

docarray/document/mixins/multimodal.py
alaeddine-13: I don't think so, I think the modality of the type should be clear.
We can also add a type annotation `Field` to allow users to provide their custom types but I'm not sure about the priority for it

docarray/types/multimodal.py
```diff
@@ -0,0 +1,166 @@
...
+    for key in TYPES_REGISTRY:
+        if key.__name__ in attribute_type:
+            _, deserializer = TYPES_REGISTRY[key]
+            return deserializer(attribute_doc)
+    else:
+        raise ValueError('Invalid attribute type')
+
+
+def _get_doc_nested_attribute(attribute_doc: 'Document', nested_cls):
+    if not is_dataclass(nested_cls):
```
alaeddine-13, line 91: weak check here

docarray/document/mixins/multimodal.py
```diff
@@ -0,0 +1,133 @@
...
+    ITERABLE_PRIMITIVE = 'iterable_primitive'
+    ITERABLE_DOCUMENT = 'iterable_document'
+    NESTED = 'nested'
+    ITERABLE_NESTED = 'iterable_nested'
+
+
+class MultiModalMixin:
+    @classmethod
+    def from_dataclass(cls, obj):
+        if not is_dataclass(obj):
```
alaeddine-13, line 25: weak check here

alaeddine-13 (commented):
also add test about support of default values in dataclass and see if it works

hanxiao (approved):

PR #192, opened 2022-03-11
feat: add `$exists` logic

numb3r3:
```python
# return True if `text` exists
result = docs.find({'text': {'$exists': True}})

# return True if `text` does not exist
result = docs.find({'text': {'$exists': False}})
```
**Note: ** empty string content is also considered as not exists.

### DIFF
--- a/docarray/array/queryset/lookup.py
+++ b/docarray/array/queryset/lookup.py
@@ -106,6 +106,8 @@ def lookup(key, val, doc: 'Document') -> bool:
         return iff_not_none(value, lambda y: re.search(val, y) is not None)
     elif last == 'size':
         return iff_not_none(value, lambda y: len(y) == val)
+    elif last == 'exists':
+        return (value is None or value == '' or value == b'') != val
     else:
         # return value == val
         raise ValueError(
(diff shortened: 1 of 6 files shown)


docarray/array/queryset/lookup.py
```diff
@@ -106,6 +106,8 @@ def lookup(key, val, doc: 'Document') -> bool:
         return iff_not_none(value, lambda y: re.search(val, y) is not None)
     elif last == 'size':
         return iff_not_none(value, lambda y: len(y) == val)
+    elif last == 'exists':
+        return (value is None or value == '' or value == b'') != val
```
JoanFM, line 110: shouldn't `bool(value)` do the trick?

JoanFM (changes requested):

docarray/array/queryset/lookup.py
numb3r3: in case `Document(tags={'x':0})`, then the query `{'tags_x': {'$exists': True}}` should return True

JoanFM (approved):

PR #199, opened 2022-03-14
fix: exist validate 

numb3r3:


JohannesMessner (approved):
Looks good, works for my use case

PR #207, opened 2022-03-15
Feat elastic storage

davidbp:


docarray/array/document.py
```diff
@@ -52,6 +52,16 @@ def __new__(
         """Create a AnnLite-powered DocumentArray object."""
         ...
 
+    @overload
+    def __new__(
+        cls,
+        _docs: Optional['DocumentArrayElastic'] = None,
+        storage: str = 'elastic',
+        config: Optional[Union['ElasticConfig', Dict]] = None,
+    ) -> 'DocumentArrayAnnlite':
```
numb3r3, line 61: should be `DocumentArrayElastic`

docarray/array/document.py
```diff
@@ -52,6 +52,16 @@ def __new__(
         """Create a AnnLite-powered DocumentArray object."""
         ...
 
+    @overload
+    def __new__(
+        cls,
+        _docs: Optional['DocumentArrayElastic'] = None,
```
numb3r3, line 58: `DocumentArraySourceType` as the type of `_docs`

docarray/array/storage/elastic/backend.py
```diff
@@ -108,5 +110,39 @@ def __setstate__(self, state):
 
     def _get_storage_infos(self) -> Dict:
         return {
-            'Backend': 'ElasticSearchConfig',
+            'Backend': 'ElasticConfig',
```
numb3r3, line 113: more info need to be included here

numb3r3 (changes requested):

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if not config:
+            raise ValueError('Empty config is not allowed for Elastic storage')
+        elif isinstance(config, dict):
+            config = dataclass_from_dict(ElasticConfig, config)
+
+        self._index_name_offset2id = 'index_offset2id'
+        self._config = config
+
+        self._client = self._build_client(self._config)
```
alaeddine-13, line 57: since it's an instance method, no need to accept config as a parameter I guess

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,154 @@
...
+        self._build_offset2id_index(self._index_name_offset2id)
+
+        super()._init_storage()  # CALLS _load_offset2ids > _get_offset2ids_meta
+
+    def _build_offset2id_index(self, index_name):
+        self._client.indices.delete(index=index_name, ignore=[404])
+        self._client.indices.create(index=index_name)
+
+    def _build_hosts(self, elastic_config):
```
alaeddine-13, line 66: just use self._config

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,154 @@
...
+        return elastic_config.host + ':' + str(elastic_config.port)
+
+    def _build_schema_from_elastic_config(self, elastic_config):
```
alaeddine-13, line 69: just use self._config

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,154 @@
...
+                        "dims": elastic_config.n_dim,
+                        "index": "true",
+                        "similarity": elastic_config.distance,
+                    },
+                },
+            }
+        }
+        return da_schema
+
+    def _build_client(self, elastic_config):
```
alaeddine-13, line 86: just use self._config

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,154 @@
...
+
+        client = Elasticsearch(
+            hosts=self._build_hosts(elastic_config),
```
alaeddine-13, line 89: same

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,154 @@
...
+            ca_certs=elastic_config.ca_certs,
+            basic_auth=elastic_config.basic_auth,
+        )
+
+        schema = self._build_schema_from_elastic_config(elastic_config)
+
+        if client.indices.exists(index=elastic_config.index_name):
+            client.indices.delete(index=elastic_config.index_name)
```
alaeddine-13, line 97: let's align this behaviour with other backends ?
if elastic_config.index_name is not specified (None), generate a random name and create index, otherwise, if it's specified, just load it, no need to alter the data (otherwise, there will be no persistence at all right ?)

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        if not self._client:
+            raise ValueError('Elastic client does not exist')
+
+        n_docs = self._client.count(index=self._index_name_offset2id)["count"]
+        if n_docs != 0:
+            offsets = [x for x in range(n_docs)]
+            resp = self._client.mget(index=self._index_name_offset2id, ids=offsets)
+            ids = [x['_source']['blob'] for x in resp['docs']]
+            return ids
```
alaeddine-13, line 155: can't we just store all offset2id in a single row within the index (use a fixed id, for instance '0') ?

docarray/array/storage/elastic/getsetdel.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def _set_doc_by_id(self, _id: str, value: 'Document'):
+        """Concrete implementation of base class' ``_set_doc_by_id``
+
+        :param _id: the id of doc to update
+        :param value: the document to update to
+        """
+        doc_copy = copy.deepcopy(value)
+        embedding = value.embedding
+        doc_copy.embedding = None
```
alaeddine-13: we don' do this in other storage backends as we have to keep a version of embeddings within the serialized documents.
This is because the indexed embeddings can be different from the actual ones (e.g adding noise).
what would happen if we call _get_doc_by_id(id).embedding right after _set_doc_by_id ?

docarray/array/storage/elastic/getsetdel.py
```diff
@@ -0,0 +1,85 @@
...
+            {
+                "_op_type": "index",
+                '_id': value.id,
+                '_index': elastic_config.index_name,
+                'embedding': embedding,
+                'blob': doc_copy.to_base64(),
+            }
+        ]
+
+        self._send_requests(request)
```
alaeddine-13, line 54: does this perform an upsert operation ?

docarray/array/storage/elastic/seqlike.py
```diff
@@ -0,0 +1,72 @@
...
+        :return: string representation of this object
+        """
+        return f'<{self.__class__.__name__} (length={len(self)}) at {id(self)}>'
+
+    def extend(self, values: Iterable['Document']) -> None:
+        """Extends the array with the given values
+
+        :param values: Documents to be added
+        """
+        with self._client.batch(batch_size=50) as _b:
```
alaeddine-13: if there's a batch support, let's also use it in _set_docs_by_ids

setup.py
```diff
@@ -66,6 +66,7 @@
             'annlite>=0.3.0',
             'qdrant-client~=0.6.0',
             'strawberry-graphql',
+            'elasticsearch>=8.0.1',
         ],
         'qdrant': [
             'qdrant-client~=0.6.0',
```
alaeddine-13, line 72: maybe let's add other tags for elastic, weaviate and pqlite ?

docarray/array/storage/elastic/seqlike.py
davidbp: there is none, updated

setup.py
davidbp: not for this PR I guess

docarray/array/document.py
davidbp: done

docarray/array/storage/elastic/backend.py
davidbp: done

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,176 @@
...
+        config: Optional[Union[ElasticConfig, Dict]] = None,
+        **kwargs,
+    ):
+
+        if not config:
+            raise ValueError('Empty config is not allowed for Elastic storage')
+        elif isinstance(config, dict):
+            config = dataclass_from_dict(ElasticConfig, config)
+
+        id = uuid.uuid4().hex
```
numb3r3, line 54: could use `from docarray.helper import random_identity` to generate random uuid

numb3r3 (changes requested):

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,217 @@
...
+        warnings.warn(f'The table name is changed to {ret} due to illegal characters')
+    return ret
+
+
+@dataclass
+class ElasticConfig:
+    n_dim: int  # dims  in elastic
+    basic_auth: Optional[Tuple[str, str]] = None
+    ca_certs: Optional[str] = None
+    distance: str = 'l2_norm'  # similarity in elastic
```
numb3r3, line 48: what about using `cosine` as the default distance metric

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,217 @@
...
+        self._client = self._build_client()
+        self._build_offset2id_index()
+
+        # Note super()._init_storage() calls _load_offset2ids which calls _get_offset2ids_meta
+        super()._init_storage()
+
+        if _docs is None:
+            return
+        elif isinstance(
+            _docs, (DocumentArray, Sequence, Generator, Iterator, itertools.chain)
```
numb3r3, line 90: `if isinstance(docs, Iterable) and self._ref_doc is not None:` is good enough here.

tests/unit/array/conftest.py
```diff
@@ -13,7 +13,7 @@ def start_storage():
         f"docker-compose -f {compose_yml} --project-directory . up  --build -d "
         f"--remove-orphans"
     )
-    time.sleep(5)
+    time.sleep(20)
```
numb3r3, line 16: I'm worried 20s is too large, and our CI workflow is too slow.

numb3r3 (changes requested):

tests/unit/array/conftest.py
davidbp: it takes much more time for elastic to be usable than other docker images, I've tried with 10 and 15 and didn't worked.

numb3r3: I see

docarray/array/storage/elastic/backend.py
davidbp: makes sense, done it

docarray/array/storage/elastic/backend.py
davidbp: This would allow other iterables that are not supported? I leave it as is, which is how it is done in other storage backends.

docarray/array/storage/elastic/backend.py
davidbp: True, done

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,154 @@
...
+        if not config:
+            raise ValueError('Empty config is not allowed for Elastic storage')
+        elif isinstance(config, dict):
+            config = dataclass_from_dict(ElasticConfig, config)
+
+        self._index_name_offset2id = 'index_offset2id'
+        self._config = config
+
+        self._client = self._build_client(self._config)
+        self._build_offset2id_index(self._index_name_offset2id)
```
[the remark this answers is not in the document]
davidbp: modified as suggested

docarray/array/storage/elastic/backend.py
davidbp: we could,  but this does the loading with a single mget as well

docarray/array/storage/elastic/backend.py
numb3r3: Please check the latest main, this is the current implementation in other storage

docarray/array/storage/elastic/backend.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def _refresh(self, index_name):
+        self._client.indices.refresh(index=index_name)
+
+    def _doc_id_exists(self, doc_id):
+        return self._client.exists(index=self._config.index_name, id=doc_id)
+
+    def _get_storage_infos(self) -> Dict:
+        return {
+            'Backend': 'ElasticConfig',
```
numb3r3, line 150: Typo, change to ElasticSearch here

numb3r3 (changes requested):

docarray/array/storage/elastic/backend.py
davidbp: done.

numb3r3 (approved):

Issue #221, opened 2022-03-23
Bug in to_pydantic_model/from_pydantic_model


JohannesMessner:
**Problem description:**

The class methods `to_pydantic_model()` / `from_pydantic_model()` do not appear to be inverses of each other.

**How to reproduce:**

``` python
>>> Document.from_pydantic_model(Document(blob=b'hello').to_pydantic_model())
binascii.Error: Invalid base64-encoded string: number of data characters (5) cannot be 1 more than a multiple of 4
```

Seems like this part of the conversion (in the `from_pydantic_model()` method) might be based on an assumption that does not always hold:

``` python
elif f_name == 'blob':
    # here is a dirty fishy itchy trick
    # the original bytes will be encoded two times:
    # first time is real during `to_dict/to_json`, it converts into base64 string
    # second time is at `from_dict/from_json`, it is unnecessary yet inevitable, the result string get
    # converted into a binary string and encoded again.
    # consequently, we need to decode two times here!
    fields[f_name] = base64.b64decode(base64.b64decode(value))
```

My guess is that the `blob` is actually only encoded once, but it tries to decode twice, giving the error above.

**Related Issue**

The `.dict()` method of `PydanticDocument` does not implement any decoding of `blob`, but I think it should:

```python
>>> d = Document(blob=b'hello')
>>> d.blob
b'hello'
>>> d_pyd = d.to_pydantic_model()  # encodes the blob
>>> d_pyd.blob
'aGVsbG8='
>>> d_pyd.dict()['blob']  # I think this should decode the blob and return b'hello'
'aGVsbG8='
```

@JoanFM I believe that if this behaviour of `.dict()` is fixed then the trick in the core of creating a `DocumentArray` becomes unnecessary.

PR #223, opened 2022-03-23
fix: encode and decode only once

alaeddine-13:
closes: https://github.com/jina-ai/docarray/issues/221

### DIFF
--- a/docarray/document/mixins/pydantic.py
+++ b/docarray/document/mixins/pydantic.py
@@ -68,12 +68,6 @@ class PydanticMixin:
             elif f_name == 'embedding' or f_name == 'tensor':
                 fields[f_name] = np.array(value)
             elif f_name == 'blob':
-                # here is a dirty fishy itchy trick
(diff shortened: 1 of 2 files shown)


docarray/document/mixins/pydantic.py
```diff
@@ -72,7 +74,7 @@ def from_pydantic_model(cls: Type['T'], model: 'BaseModel') -> 'T':
                 # second time is at `from_dict/from_json`, it is unnecessary yet inevitable, the result string get
                 # converted into a binary string and encoded again.
                 # consequently, we need to decode two times here!
```
numb3r3, line 76: The code comments above are still reasonable?

numb3r3 (changes requested):

numb3r3 (approved):
LGTM

docarray/document/mixins/pydantic.py
alaeddine-13: removed

PR #233, opened 2022-03-29
Fix elastic persistence

davidbp:
Ensure offsets2id are updated after adding Documents with `.extend`.
On main, the following script does not have offset updated.
This happens in other backends as well
```python
storage = 'elasticsearch'
index_name = 'index_name11'
da = DocumentArray(
    storage=storage,
    config={'index_name': index_name, 'n_dim': 128},
#    persist=False
)
print(f'len(da)={len(da)}')
print(f'len(da._offset2ids.ids)={len(da._offset2ids.ids)}')
da.extend([Document(embedding=np.random.random(128)) for _ in range(10)])
print(f'len(da)={len(da)}')
print(f'len(da._offset2ids.ids)={len(da._offset2ids.ids)}')

da2 = DocumentArray(
    storage=storage,
    config={'index_name': index_name, 'n_dim': 128},
)

print(f'len(da2)={len(da2)}')
print(f'len(da2._offset2ids.ids)={len(da2._offset2ids.ids)}')
```


docarray/array/storage/elastic/seqlike.py
```diff
@@ -79,3 +79,4 @@ def extend(self, values: Iterable['Document']) -> None:
         if len(request) > 0:
             self._send_requests(request)
             self._refresh(self._config.index_name)
+            self._update_offset2ids_meta()
```
winstonww, line 82: I see for other backends `_update_offset2ids_meta` is called only in `_save_offset2ids` on delete, I guess you encountered an issue when docs are extended but not properly saved?

davidbp:
closing this for a more general solution with a context manager

PR #234, opened 2022-03-29
fix: fix PIL dependency

alaeddine-13:
Currently, installing docarray and using it fails because PIL is not installed with the default tag.


azayz:
do not merge, the import should be `from PIL import Image`

azayz (approved):

alaeddine-13:
let's wait until I raise a PR for testing docarray import and confirm it fails on this

PR #235, opened 2022-03-29
ci: add basic import test

alaeddine-13:
ci: test basic import


hanxiao:
looks like our black script is not happy if no python code changes in a PR, could you fix it as well

PR #236, opened 2022-03-29
Feat context manager backends

davidbp:
Provide a context manager to ensure that modifications inside the "with statement" modify the underlying db at exit.

```python
import numpy as np

storage = 'sqlite'
table_name = 'Test02'
connection = './test.db'

da = DocumentArray(
    storage=storage,
    config={'table_name': table_name, 'connection':connection},
)

with da as da_open:
    da_open.append(Document(embedding=np.random.random(128)))
    da_open.append(Document(embedding=np.random.random(128)))

print(f'len(da)={len(da)}')
print(f'len(da._offset2ids.ids)={len(da._offset2ids.ids)}')

da2 = DocumentArray(
    storage=storage,
    config={'table_name': table_name, 'connection':connection},
)
print(f'len(da2)={len(da2)}')
print(f'len(da2._offset2ids.ids)={len(da2._offset2ids.ids)}')
```


docarray/array/annlite.py
```diff
@@ -7,3 +7,13 @@
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        """
+        Ensures that offset2ids are stored in the db after
+        operations in the DocumentArray are performed.
+        """
+        self._save_offset2ids()
```
hanxiao, line 19: 1. promote this to one of the base class, so that you don't need to repeatedly writing it
2. unify the name of `_save_offset2ids`, `_update_offset2ids_meta()` 
3. by doing 1 and 2, all storage backend should not need to implement them

hanxiao (changes requested):
need refactoring

docarray/array/annlite.py
davidbp: true, refactored

hanxiao (changes requested):
the only thing missing is we need to add a section in the docs to explain this a bit, so how about add it to here? https://docarray.jina.ai/advanced/document-store/#known-limitations

```
### Reference two DocumentArray to the same backend

Blah, blah, use context manager.

``code

``output

```

PR #238, opened 2022-03-29
feat(hubble): push/pull user authentication

delgermurun:
Related https://github.com/jina-ai/hubble/issues/353

### DIFF
--- a/docarray/array/mixins/io/pushpull.py
+++ b/docarray/array/mixins/io/pushpull.py
@@ -2,13 +2,32 @@ import json
 import os
 import warnings
 from functools import lru_cache
-from typing import Type, TYPE_CHECKING
+from pathlib import Path
+from typing import Dict, Type, TYPE_CHECKING
 from urllib.request import Request, urlopen
 
+from ....exceptions import ObjectNotFoundError
 from ....helper import get_request_header
 
 if TYPE_CHECKING:
     from ....typing import T
+
+JINA_CLOUD_CONFIG = 'config.json'
+
+
+@lru_cache()
+def _get_hub_config() -> Dict:
+    hub_root = Path(os.environ.get('JINA_HUB_ROOT', Path.home().joinpath('.jina')))
+
+    if not hub_root.exists():
+        hub_root.mkdir(parents=True, exist_ok=True)
+
+    config_file = hub_root.joinpath(JINA_CLOUD_CONFIG)
+    if config_file.exists():
+        with open(config_file) as f:
+            return json.load(f)
+
+    return {}
 
 
 @lru_cache()
@@ -41,17 +60,17 @@ class PushPullMixin:
 
     _max_bytes = 4 * 1024 * 1024 * 1024
 
-    def push(self, token: str, show_progress: bool = False) -> None:
+    def push(self, name: str, show_progress: bool = False) -> Dict:
         """Push this DocumentArray object to Jina Cloud which can be later retrieved via :meth:`.push`
 
         .. note::
-            - Push with the same ``token`` will override the existing content.
+            - Push with the same ``name`` will override the existing content.
             - Kinda like a public clipboard where everyone can override anyone's content.
-              So to make your content survive longer, you may want to use longer & more complicated token.
+              So to make your content survive longer, you may want to use longer & more complicated name.
             - The lifetime of the content is not promised atm, could be a day, could be a week. Do not use it for
               persistence. Only use this full temporary transmission/storage/clipboard.
 
-        :param token: a key that later can be used for retrieve this :class:`DocumentArray`.
+        :param name: a name that later can be used for retrieve this :class:`DocumentArray`.
         :param show_progress: if to show a progress bar on pulling
         """
         import requests
@@ -64,11 +83,16 @@ class PushPullMixin:
                     'DocumentArray',
                     delimiter,
                 ),
-                'token': token,
+                'name': name,
+                'type': 'documentArray',
             }
         )
 
         headers = {'Content-Type': ctype, **get_request_header()}
+
+        auth_token = _get_hub_config().get('auth_token')
+        if auth_token:
+            headers['Authorization'] = f'token {auth_token}'
 
         _head, _tail = data.split(delimiter)
         from rich import filesize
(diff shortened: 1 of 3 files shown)


docarray/array/mixins/io/pushpull.py
```diff
@@ -52,24 +71,35 @@ def push(self, token: str, show_progress: bool = False) -> None:
               persistence. Only use this full temporary transmission/storage/clipboard.
 
         :param token: a key that later can be used for retrieve this :class:`DocumentArray`.
+        :param id: an unique identifier that returned from previous :meth:`.push`
```
hanxiao, line 74: > returned from previous :meth:`.push`
what does this mean?

hanxiao: still confused, what is `id`? This variable is too general.

is it about filename? or a credential? to help me better understand it, could you copy paste one of your "id" so that I can see it?

delgermurun: `id` is the id of the database entry on the Hubble.

Here is a demo:
```python
da = DocumentArray([Document(text='hello')])
res = da.push() // authenticated users can push without token
```
`res` is dict which is response from Hubble API
`{'_id': '62432cd4665a8ed1a5f87d6a', 'visibility': 'private', ...}`

`62432cd4665a8ed1a5f87d6a` is the `id` I am talking about.
Later on, I can use this `id` to push or pull old DA.

```python
da = DocumentArray([Document(text='hello world')])
da.push(id='62432cd4665a8ed1a5f87d6a') // replacing old DA
// or
DocumentArray.pull(id='62432cd4665a8ed1a5f87d6a')
```

Yes, it is confusing because now we have 2 ways to do `push/pull`.
It is needed because logged-in users don't have to come up with tokens to protect their DA. They can just save `id` of it somewhere. No one can see or edit their DA by default.
Also, 3rd party programs (e.g, Finetuner) can manage user DAs if they have user authentication.

Sorry for the confusion, I hope it helps. Maybe we need to rename the parameter and/or improve documentation.

docarray/array/mixins/io/pushpull.py
```diff
@@ -41,7 +58,9 @@ class PushPullMixin:
 
     _max_bytes = 4 * 1024 * 1024 * 1024
 
-    def push(self, token: str, show_progress: bool = False) -> None:
+    def push(
+        self, token: str = None, id: str = None, show_progress: bool = False
+    ) -> None:
```
alaeddine-13, line 63:
```suggestion
    ) -> Dict:
```

docarray/array/mixins/io/pushpull.py
```diff
@@ -126,21 +159,34 @@ def pull(
...
         :return: a :class:`DocumentArray` object
         """
 
         import requests
 
-        url = f'{_get_cloud_api()}/v2/rpc/da.pull?token={token}'
-        response = requests.get(url)
+        headers = {}
+
+        auth_token = _get_hub_config().get('auth_token')
```
alaeddine-13, line 172: so it's either you provide an `auth_token` if you're logged in or a DA `token` if you're not ?

docarray/array/mixins/io/pushpull.py
```diff
@@ -126,21 +159,34 @@ def pull(
...
+        if id:
+            params['id'] = id
+
+        url = f'{_get_cloud_api()}/v2/rpc/da.pull?{parse.urlencode(params)}'
+        response = requests.get(url, headers=headers)
 
         if response.ok:
             url = response.json()['data']['download']
         else:
             raise FileNotFoundError(
```
alaeddine-13, line 188: The error here is misleading, maybe we should raise something like `BadRequestError` or so ?

docarray/array/mixins/io/pushpull.py
delgermurun: You can use DA `token` even when you're logged in.
User login status and DA `id`/`token` params are not related.
You can pull other users' DA by `id` if it is public.
API has an ability to set visibility (`public` or `private`) but is not added to this mixin yet.

tests/unit/array/mixins/test_pushpull.py
```diff
@@ -140,3 +154,44 @@ def _mock_get(url, stream=False, headers=None):
...
+
+
+def test_api_authorization_header(mocker, monkeypatch, tmpdir):
+    from docarray.array.mixins.io.pushpull import _get_hub_config
+
+    _get_hub_config.cache_clear()
+    os.environ['JINA_HUB_ROOT'] = str(tmpdir)
+
+    token = 'test-auth-token'
+    with open(tmpdir / 'config.json', 'w') as f:
```
alaeddine-13, line 166: let's store config.json as a global name `JINA_CLOUD_CONFIG`

hanxiao (changes requested):
user journey needs redesign, right now `token` and `id` is so confusing

docarray/array/mixins/io/pushpull.py
```diff
@@ -104,43 +127,55 @@ def gen():
             yield _tail
 
         res = requests.post(
-            f'{_get_cloud_api()}/v2/rpc/da.push', data=gen(), headers=headers
+            f'{_get_cloud_api()}/v2/rpc/artifact.upload', data=gen(), headers=headers
```
alaeddine-13, line 130: I'm just wondering here why the API endpoint was renamed from `da.push` to `artifact.upload`
Isn't artifact related to finetuner ?

delgermurun: Not specifically related to finetuner. Under the hood, every types of storage will use Hub's `artifact storage system`.
Doesn't matter docarray, finetuner model, or anything else in the future.

docarray/array/mixins/io/pushpull.py
```diff
@@ -104,43 +127,55 @@ def gen():
...
+        url = f'{_get_cloud_api()}/v2/rpc/artifact.getDownloadUrl?name={name}'
+        response = requests.get(url, headers=headers)
 
         if response.ok:
             url = response.json()['data']['download']
         else:
-            raise FileNotFoundError(
-                f'can not find `{token}` DocumentArray on the Cloud. Misspelled?'
+            json_res = response.json()
+            raise Exception(
```
alaeddine-13, line 174:
```suggestion
            raise ObjectNotFoundError(
```
and let's create a docarray/exceptions.py module and include such an error

delgermurun: addressed

PR #241, opened 2022-03-30
fix(sqlite): persist offset2id only in destructor and context manager

alaeddine-13:


hanxiao (changes requested):
let's hold it for now

JoanFM:
Hey @alaeddine-13 could you fill more details of the issue and what is the problem and why is this a good solution?

JoanFM (changes requested):
Please add details

JoanFM:
Closing until further notice

PR #247, opened 2022-03-31
fix: torch store embedding

davidbp:
@slettner showed this snippet

```python
import torch
import numpy as np

arr = np.ones((100, 512), dtype=np.float32)
tensor = torch.from_numpy(arr)


docsSmall = DocumentArray([Document() for _ in range(100)])
docsBig = DocumentArray([Document() for _ in range(100)])

docsBig.embeddings = tensor
docsSmall.embeddings = arr

print(f'Size docA {len(docsSmall.to_bytes())}')
print(f'Size docB {len(docsBig.to_bytes())}')
```

That resulted with 1500x memory usage for pytorch
```
Size docA 3444946
Size docB 4608772715
```

in this PR

```
Size docA 3321569
Size docB 3705623
```


hanxiao (changes requested):
please pull main before making pr, also, this breaks the Pythonic design of `.attr =`

docarray/array/storage/base/getsetdel.py
```diff
@@ -260,7 +262,7 @@ def _set_doc_attr_by_id(self, _id: str, attr: str, value: Any):
 
         d = self._get_doc_by_id(_id)
         if hasattr(d, attr):
-            setattr(d, attr, value)
+            setattr(d, attr, detach_tensor_if_present(value))
```
numb3r3, line 265: I would prefer to use if statement here, I believe it would be much safer and more efficient.
```
if attr in ['embedding', 'tensor']:
    value = detach_tensor(value)
    ...
```

numb3r3 (changes requested):

docarray/math/ndarray.py
```diff
@@ -121,7 +123,10 @@ def get_array_type(array: 'ArrayType') -> Tuple[str, bool]:
     if 'scipy' in module_tags and 'sparse' in module_tags:
         return 'scipy', True
 
-    raise TypeError(f'can not determine the array type: {module_tags}.{class_name}')
+    if raise_error_if_not_array:
+        raise TypeError(f'can not determine the array type: {module_tags}.{class_name}')
```
alaeddine-13, line 127: Why do we need to disable raising the error ?

alaeddine-13 (approved):

PR #281, opened 2022-04-17
test(image): add unit tests for _to_image_tensor

alessiamarcolini:
This PR is adding two unit tests for the function `_to_image_tensor` in `docarray/document/mixins/image.py`.

--- 
**Note**: I could not find a way to cover lines 340-341: https://github.com/jina-ai/docarray/blob/73f2aef9d3145f4d29982681fc8632230c37fdf4/docarray/document/mixins/image.py#L338-L341

By looking at the internals of `PIL.Image.convert`, it seems like there is no way to read the image correctly at
https://github.com/jina-ai/docarray/blob/73f2aef9d3145f4d29982681fc8632230c37fdf4/docarray/document/mixins/image.py#L333

and then make it fail during the conversion to `RGB`. Could it be a leftover/patch for something weird that happened in the past? If not, how can this be meaningfully tested?

### DIFF
--- a/tests/unit/document/test_image.py
+++ b/tests/unit/document/test_image.py
@@ -1,0 +1,42 @@
+import os
+
+import numpy as np
+import pytest
+from PIL import Image
+import io
+
+from docarray.document.mixins.image import _to_image_tensor
+
+
+@pytest.mark.parametrize(
+    "width, height, output_shape",
+    [
+        (None, None, (50, 10, 3)),
+        (8, None, (50, 8, 3)),
+        (None, 30, (30, 10, 3)),
+        (30, 8, (8, 30, 3)),
+    ],
+)
+def test_to_image_tensor_pil(rgb_image_path, width, height, output_shape):
+    tensor = _to_image_tensor(rgb_image_path, width, height)
+
+    assert tensor.shape == output_shape
+    assert isinstance(tensor, np.ndarray)
+
+
+def test_to_image_tensor_blob(rgb_image_path):
+    with open(rgb_image_path, "rb") as f:
+        blob = io.BytesIO(f.read())
+
+    tensor = _to_image_tensor(blob, None, None)
+
+    assert tensor.shape == (50, 10, 3)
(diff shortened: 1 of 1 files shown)


alaeddine-13:
Hey @alessiamarcolini , thanks for your contribution.
I don't think that private functions should be tested. Instead, we can try to cover those cases in public methods that use it.
For example, you can try covering the cases of width and height != None to `convert_blob_to_image_tensor` which should cover more lines in _to_image_tensor

alessiamarcolini:
Hi @alaeddine-13, thank you for your comment.
Is testing only public methods/functions part of the project guidelines? If so, I think it should be mentioned in the [CONTRIBUTING.md](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md) to inform future contributors.

Should I remove the tests then and write new ones for `convert_blob_to_image_tensor`? Or can I keep the existing ones?
Thank you.

hanxiao (changes requested):
thanks for the PR. @alaeddine-13 she does improve the coverage, so I think it's good

one remark, we use single quote for string in the codebase. So please adjust accordingly

tests/unit/document/test_image.py
```diff
@@ -0,0 +1,42 @@
...
+
+    assert tensor.shape == (50, 10, 3)
+    assert isinstance(tensor, np.ndarray)
+
+
+@pytest.fixture
+def rgb_image_path(tmpdir):
+    img_path = os.path.join(tmpdir, 'image.png')
+    RGB_COLOR_10X50_155_0_0 = Image.new('RGB', size=(10, 50), color=(0, 0, 0))
+    RGB_COLOR_10X50_155_0_0.save(img_path)
```
alaeddine-13, line 41:
```suggestion
    image = Image.new('RGB', size=(10, 50), color=(0, 0, 0))
    image.save(img_path)
```

alaeddine-13 (approved):

alaeddine-13:
Thanks @alessiamarcolini for your contribution!

PR #296, opened 2022-04-22
feat: add timeout parameter

### LINKED ISSUE #290: Expose Timeout Parameters in weaviate
Weaviate storage currently does not allow specifying the Timeout parameters.
Add the parameter to `WeaviateConfig` and used when creating the weaviate client

azayz:
fixes #290

### DIFF
--- a/docarray/array/storage/weaviate/backend.py
+++ b/docarray/array/storage/weaviate/backend.py
@@ -1,5 +1,5 @@
 import uuid
-from dataclasses import dataclass, field, asdict
+from dataclasses import dataclass, field
 from typing import (
     Iterable,
     Dict,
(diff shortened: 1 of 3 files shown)


docarray/array/storage/weaviate/backend.py
```diff
@@ -71,9 +72,15 @@ def _init_storage(
...
+        if config.timeout_config:
+            self._client = weaviate.Client(
+                f'{config.protocol}://{config.host}:{config.port}',
+                timeout_config=config.timeout_config,
+            )
+        else:
+            self._client = weaviate.Client(
+                f'{config.protocol}://{config.host}:{config.port}',
+            )
         self._config = config
```
alaeddine-13, line 84: here we have 2 options:
* give our own default parameters (with the burden of maintaining values in case client changes)
* follow the pattern here: https://github.com/jina-ai/docarray/blob/main/docarray/array/storage/annlite/backend.py#L70

@JoanFM  what's your opinion ?

JoanFM: I like the one where less attention from us is needed

azayz: done

hanxiao (approved):

PR #302, opened 2022-04-26
feat(helper): actively check version on start

hanxiao:


docarray/helper.py
```diff
@@ -443,3 +443,59 @@ def filter_dict(d: Dict) -> Dict:
...
+        pass
+
+
+def is_latest_version(package: str = None, github_repo: str = None) -> None:
+    """Check if there is a latest version from Pypi, set env `NO_VERSION_CHECK` to disable it.
+
+    :param package: package name if none auto-detected
+    :param github_repo: repo name that contains CHANGELOG if none then the same as package name
+    """
+    import threading
```
JoanFM, line 497: please do not use `threading`. This caused the deadlock on core, use multiprocessing

JoanFM (changes requested):

docarray/helper.py
hanxiao: deadlock in the core is caused by JinaLogger, not this.

PR #310, opened 2022-04-28
refactor: reuse code in both random and dataset benchmarking

alaeddine-13:
Goals:

- reuse code for both random dataset and real dataset benchmarking


JoanFM (changes requested):
Please rebase this PR so that it does not show unwanted changes

PR #312, opened 2022-04-29
perf: use less memory for apply

alaeddine-13:
Goals:

- Make `apply` memory efficient and suitable for storage backends

### DIFF
--- a/docarray/array/mixins/parallel.py
+++ b/docarray/array/mixins/parallel.py
@@ -154,13 +154,8 @@ class ParallelMixin:
         # noqa: DAR201
         :return: a new :class:`DocumentArray`
         """
-        from ... import DocumentArray
-
-        new_da = DocumentArray()
         for _b in self.map_batch(*args, **kwargs):
-            new_da.extend(_b)
-        self.clear()
-        self.extend(new_da)
+            self[[doc.id for doc in _b]] = _b
         return self
 
     def map_batch(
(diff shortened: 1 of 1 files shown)


docarray/array/mixins/parallel.py
```diff
@@ -60,11 +60,8 @@ def apply(self: 'T', *args, **kwargs) -> 'T':
         # noqa: DAR201
         :return: a new :class:`DocumentArray`
         """
-        from ... import DocumentArray
-
-        new_da = DocumentArray(self.map(*args, **kwargs))
-        self.clear()
-        self.extend(new_da)
+        for doc in self.map(*args, **kwargs):
```
numb3r3, line 63: use batch_map would be efficient for some backends?

numb3r3 (changes requested):

docarray/array/mixins/parallel.py
hanxiao: hmm, interesting thought, that could be true. especially when the afterward `da[d.id] = d` is IO bounded https://docarray.jina.ai/fundamentals/documentarray/parallelization/#use-map-batch-to-overlap-cpu-and-network-time

@alaeddine-13 you think?

alaeddine-13: then it's better to use it for `apply_batch` not `apply`.
I have applied the same logic for `apply_batch`, can you check ?

PR #315, opened 2022-04-29
fix: fix find by vector of n_dim=1

alaeddine-13:
fix: fix find by vector of n_dim=1


docarray/array/storage/qdrant/backend.py
```diff
@@ -187,6 +187,9 @@ def _map_embedding(self, embedding: 'ArrayType') -> List[float]:
         if embedding.ndim > 1:
             embedding = np.asarray(embedding).squeeze()
 
+        if embedding.ndim == 0:
```
hanxiao, line 190: when will this happen?

hanxiao, line 190:
```suggestion
        if embedding.ndim == 0:  # scalar
```

PR #317, opened 2022-05-01
update refer links to other docs with utm_source

Roshanjossey:
Goals:

- Track visits from docarray to other docs in jina eco-system

### DIFF
--- a/docs/_static/source-in-links.js
+++ b/docs/_static/source-in-links.js
@@ -1,4 +1,4 @@
-function addUTMSourceToLinks(href) {
+function addUTMSourceToLink(href) {
     try {
         var url = new URL(href);
         url.searchParams.set("utm_source", "docarray");
--- a/docs/_static/source-in-links.js
+++ b/docs/_static/source-in-links.js
@@ -4,17 +4,19 @@ function addUTMSourceToLink(href) {
         url.searchParams.set("utm_source", "docarray");
         return url.href
     }
-    catch{}
+    catch{
+        return href
+    }
 }
 
-function addSourceToAllLinks() {
+function addUTMSourceToLinks() {
     var anchors = document.getElementsByTagName("a");
 
     for (var i = 0; i < anchors.length; i++) {
-        anchors[i].href = addUTMSourceToLinks(anchors[i].href)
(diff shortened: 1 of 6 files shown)


alexcg1 (approved):
LGTM :+1:

alaeddine-13:
@Roshanjossey  I don't think this is a sustainable way to add the query string parameter to clicked links. We can easily forget to add it on existing or new links. In order to give reliable information, is it possible to modify the website so that it adds `utm_source=docarray` to all the clicked links ? for example https://blog.miguelbernard.com/how-to-dynamically-add-a-query-string-to-all-links-in-a-page

Roshanjossey:
@alaeddine-13, good point. I've made that change

docs/_static/source-in-links.js
```diff
@@ -0,0 +1,20 @@
+function addUTMSourceToLinks(href) {
+    try {
+        var url = new URL(href);
+        url.searchParams.set("utm_source", "docarray");
+        return url.href
+    }
+    catch{}
```
alaeddine-13, line 7: Maybe just return href in the catch clause ?

docs/_static/source-in-links.js
```diff
@@ -0,0 +1,20 @@
+function addUTMSourceToLinks(href) {
```
alaeddine-13, line 1: Rename to `addUTMSourceToLink`

alaeddine-13:
> @alaeddine-13, good point. I've made that change

Thanks, I just added small comments. Is it possible to apply this logic to other PRs that target other documentation websites if there are any ? Thanks

docs/_static/source-in-links.js
Roshanjossey: Catch block is executed only if there's an error in try block. If there are no errors, nothing will be returned from that function

Roshanjossey:
@alaeddine-13, I'm made changes you suggested. I'll do the same with other docs.

docs/_static/source-in-links.js
alaeddine-13: no the logic is, attempt to add utm source to link, otherwise if failed (catch block), keep it as it is (return href as it is)

PR #318, opened 2022-05-02
feat(hubble): add public parameter to da.push

delgermurun:
Goals:

- Added `public` parameter to `push` method for indicating docarray is public or private.

### DIFF
--- a/docarray/array/mixins/io/pushpull.py
+++ b/docarray/array/mixins/io/pushpull.py
@@ -57,7 +57,7 @@ class PushPullMixin:
 
     _max_bytes = 4 * 1024 * 1024 * 1024
 
-    def push(self, name: str, show_progress: bool = False) -> Dict:
+    def push(self, name: str, show_progress: bool = False, public: bool = True) -> Dict:
         """Push this DocumentArray object to Jina Cloud which can be later retrieved via :meth:`.push`
 
         .. note::
@@ -69,6 +69,7 @@ class PushPullMixin:
 
         :param name: a name that later can be used for retrieve this :class:`DocumentArray`.
         :param show_progress: if to show a progress bar on pulling
+        :param public: If True, the DocumentArray will be shared publicly. Otherwise, it will be private.
         """
         import requests
 
@@ -82,6 +83,7 @@ class PushPullMixin:
                 ),
                 'name': name,
                 'type': 'documentArray',
+                'public': public,
             }
         )
 
--- a/tests/unit/array/mixins/test_pushpull.py
+++ b/tests/unit/array/mixins/test_pushpull.py
@@ -1,7 +1,9 @@
+import cgi
 import json
 import os
 import pytest
 import requests
+from io import BytesIO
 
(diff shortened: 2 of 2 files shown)


docarray/array/mixins/io/pushpull.py
```diff
@@ -68,6 +70,7 @@ def push(self, name: str, show_progress: bool = False) -> Dict:
               persistence. Only use this full temporary transmission/storage/clipboard.
 
         :param name: a name that later can be used for retrieve this :class:`DocumentArray`.
+        :param public: it indicates whether docarray is public or not.
```
alaeddine-13, line 73:
```suggestion
        :param public: If True, the DocumentArray will be shared publicly. Otherwise, it will be private.
```

docarray/array/mixins/io/pushpull.py
```diff
@@ -57,7 +57,9 @@ class PushPullMixin:
 
     _max_bytes = 4 * 1024 * 1024 * 1024
 
-    def push(self, name: str, show_progress: bool = False) -> Dict:
+    def push(
+        self, name: str, *, public: bool = True, show_progress: bool = False
```
alaeddine-13, line 61: why did you add * ? are we expecting extra positional args in the future ?

delgermurun: Yeah, we might need to add new parameters in the future.

I added `public` before the `show_progress` param because it is more important than `show_progress`.
I think it is more intuitive.

Since `show_progress` is a boolean and was a 2nd positional param, I changed those to keyword params.
Otherwise, the old code will work incorrectly, if `show_progress` was true.

I think it's better to fail than to work incorrectly.
Also, it will be easier to add new parameters in the future. We will just add keyword parameters without worrying about the order.

docarray/array/mixins/io/pushpull.py
delgermurun: addressed

docarray/array/mixins/io/pushpull.py
alaeddine-13, line 61:
```suggestion
        self, name: str, show_progress: bool = False, public: bool = True
```
I think better not to change the order and to we don't need to provision extra positional arguments.
If an argument is is necessary it would have existed by now, if there will be one necessary argument in the future, then it means we're breaking the API and no longer supporting the current usage (in that case, we also break the interface). Otherwise, new parameters can be just keyword arguments

alaeddine-13, line 61:
```suggestion
        self, name: str, show_progress: bool = False, public: bool = True
```
I think better not to change the order and to we don't need to provision extra positional arguments.
If an argument is is necessary it would have existed by now, if there will be one necessary argument in the future, then it means we're breaking the API and no longer supporting the current usage (in that case, we also break the interface). Otherwise, new parameters can be just keyword arguments

alaeddine-13 (approved):
LGTM

PR #327, opened 2022-05-05
fix: use backend to determine annlite DocumentArray length

davidbp:


setup.py
```diff
@@ -63,14 +63,16 @@
             'fastapi',
             'uvicorn',
             'weaviate-client~=3.3.0',
-            'annlite>=0.3.0',
```
alaeddine-13, line 66: we keep both, for full and a tag specific to annlite

alaeddine-13 (approved):

PR #331, opened 2022-05-09
feat-add-annlite-filter

davidbp:
This PR adds the API needed to call `find` leveraging the underlaying query language from annlite


numb3r3 (approved):
LGTM 👍

tests/unit/array/mixins/test_find.py
```diff
@@ -192,3 +193,42 @@ def test_find_by_tag(storage, config, start_storage):
...
+
+    thresholds = [20, 50, 100, 400]
+
+    for threshold in thresholds:
+        da.find(
+            query_da, filter={'price': {operator: threshold}}, include_metadata=True
+        )
+        for query in query_da:
+            assert all(
+                [
```
JoanFM, line 230: you should first assert the length of matches, otherwise this could be empty and pass

JoanFM (changes requested):

davidbp:
@alaeddine-13  do you mean by  ` If filter is not None and and query is not None or vector, raise error.`  the following ?     `If filter is not None and ( (query is not None) or (query is not vector))`

tests/unit/array/mixins/test_find.py
davidbp: updated with your check

alaeddine-13:
> is

yes exactly, because applying a filter when the query is a condition or when the query is text does not really make sense

docarray/array/mixins/find.py
```diff
@@ -124,14 +127,10 @@ def find(
...
+        if isinstance(query, dict) and filter is None:
             return self._filter(query)
-        elif isinstance(query, (DocumentArray, Document)):
-
-            if isinstance(query, Document):
-                query = DocumentArray(query)
-
-            _query = query.embeddings
+        elif query is None and isinstance(filter, dict):
+            return self._filter(filter)
```
JoanFM, line 133: isn't it weird that giving query or filter as input works the same?

alaeddine-13: this aims to keep the compatibility

alaeddine-13, line 133: in filter tests, we need to also test the API:
`da.find(filter=filter)` not just `da.find(filter)`

docarray/array/mixins/find.py
```diff
@@ -87,13 +87,16 @@ def find(self: 'T', query: Dict, **kwargs) -> 'DocumentArray':
...
+        query: Union[
+            'DocumentArray', 'Document', 'ArrayType', Dict, str, List[str], None
+        ] = None,
         metric: Union[
             str, Callable[['ArrayType', 'ArrayType'], 'np.ndarray']
         ] = 'cosine',
         limit: Optional[Union[int, float]] = 20,
         metric_name: Optional[str] = None,
         exclude_self: bool = False,
+        filter: Optional[Dict] = None,
```
alaeddine-13, line 99: add to docstring

docarray/array/storage/sqlite/backend.py
```diff
@@ -33,6 +35,7 @@ class SqliteConfig:
     conn_config: Dict = field(default_factory=dict)
     journal_mode: str = 'DELETE'
     synchronous: str = 'OFF'
+    columns: Optional[List[Tuple[str, str]]] = None
```
alaeddine-13, line 38: why is this needed ?

docarray/array/storage/weaviate/find.py
```diff
@@ -64,7 +89,11 @@ def _find_similar_vectors(self, query: 'WeaviateArrayType', limit=10):
         return DocumentArray(docs)
 
     def _find(
-        self, query: 'WeaviateArrayType', limit: int = 10, **kwargs
+        self,
+        query: 'WeaviateArrayType',
+        limit: int = 10,
+        filter: Optional[Dict] = None,
```
alaeddine-13, line 95: add to docstring of private methods whenever other parameters are documented

docarray/array/storage/annlite/backend.py
```diff
@@ -25,6 +27,7 @@ class AnnliteConfig:
     ef_construction: Optional[int] = None
     ef_search: Optional[int] = None
     max_connection: Optional[int] = None
+    columns: Optional[List[Tuple[str, str]]] = None
```
alaeddine-13, line 30: this snippet needs to be added also (in the constructor, I believe right before `self._config = config`):
```python
        if self._config.columns is None:
            self._config.columns = []
```

docarray/array/storage/weaviate/find.py
```diff
@@ -27,26 +29,49 @@
...
-            .do()
-        )
         docs = []
 
+        found_results = (
+            results.get('data', {}).get('Get', {}).get(self._class_name, []) or []
+        )
+        if 'errors' in results:
+            errors = '\n'.join(map(lambda error: error['message'], results['errors']))
+            raise ValueError(
```
alaeddine-13, line 69: let's also add a custom check for this case

docarray/array/storage/annlite/backend.py
alaeddine-13, line 30: although there might not be actual type mapping for annlite, let's still add this type map to annlite backend mixin:
```python
TYPE_MAP = {'str': 'str', 'float': 'float', 'int': 'int'}
```
and use `_map_type ` to map configs. At least, in case we opt to change the definition of type mapping, we will be aware of it from docarray instead of sending to downstream annlite

docarray/array/storage/weaviate/find.py
```diff
@@ -27,26 +29,49 @@
...
-            .with_limit(limit)
-            .with_near_vector(query_dict)
-            .do()
-        )
         docs = []
 
+        found_results = (
+            results.get('data', {}).get('Get', {}).get(self._class_name, []) or []
+        )
+        if 'errors' in results:
```
JoanFM, line 67: it is weird to do this check after extracting from `found_results`, shouldn't it be done before?

JoanFM (changes requested):

docarray/array/mixins/find.py
```diff
@@ -140,6 +139,12 @@ def find(
...
             else:
                 return result
+        elif isinstance(query, (DocumentArray, Document)):
+
+            if isinstance(query, Document):
+                query = DocumentArray(query)
+
+            _query = query.embeddings
         else:
             _query = query
```
mapleeit, line 149: 1. After the long if-else, `_query` could be `Dict` (when query is Dict and filter is Dict as well) or `ArrayType`. Previous code will only be `ArrayType`. This could be a potensial bug, because it assert that the type of `_query` will only be `ArrayType` in `_find` function. 
2. I don't think this long if-else logic readable, at least to me. It's clear before because we only need to detect the type of one parameter but now two. Can we refactor somehow? For example, use nested if-else.

docarray/array/mixins/find.py
```diff
@@ -227,7 +233,7 @@ def find(
 
     @abc.abstractmethod
     def _find(
-        self, query: 'ArrayType', limit: int, **kwargs
+        self, query: 'ArrayType', limit: int, filter: Optional[Dict] = None, **kwargs
```
mapleeit, line 236: `query` could be `Dict`

mapleeit (changes requested):

docarray/array/storage/weaviate/find.py
```diff
@@ -27,26 +29,49 @@
...
+        else:
+            results = (
+                self._client.query.get(
+                    self._class_name,
+                    ['_serialized', '_additional {certainty}', '_additional {id}'],
+                )
+                .with_limit(limit)
+                .with_near_vector(query_dict)
+                .do()
             )
```
StefanBogdan, line 61: The `if 'filter'`/`else` here can be done in a more readable way:
```python
        query_builder = (
            self._client.query
            .get(self._class_name, '_serialized')
            .with_additional(['id', 'certainty'])
            .with_limit(limit)
            .with_near_vector(query_dict)

        )
        if filter:
            query_builder = query_builder.with_where(filter)
        results = query_builder.do()
```

alaeddine-13: @StefanBogdan does it always perform pre-filtering in this case, even though the `with_where` statement comes after `with_near_vector` ?

docarray/array/mixins/find.py
alaeddine-13: done in `aacf3fe6637582c3fc4a0749b60a934bb2256b07`

docarray/array/storage/weaviate/find.py
alaeddine-13: fixed

docarray/array/storage/weaviate/find.py
StefanBogdan: Yes, the order does not matter.

StefanBogdan:
I reviewed the changes for Weaviate storage and it seems fine.

docarray/array/storage/weaviate/find.py
davidbp: good catch, updated

docarray/array/mixins/find.py
alaeddine-13: no can't be, it is only used in https://github.com/jina-ai/docarray/blob/d0e18e860fcc693e27275e856b939e11da25d5cd/docarray/array/mixins/find.py#L180
and in case query or filter is dict, another method will be called

alaeddine-13: should be all good now

mapleeit (approved):

docs/advanced/document-store/weaviate.md
```diff
@@ -168,3 +168,93 @@ print(results[0].text)
 ```text
 Persist Documents with Weaviate.
 ```
+
+
+## Search with filter
+
+Search with `.find` can be restricted by user-defined filters. Such filters that can be constructed using the following operators:
+
+| Name              | Description            | Equivalent Python operator |
```
JoanFM, line 177: I would link to some `weaviate` documentation for their own syntax, so that we do not need to mantain these lists

JoanFM (changes requested):

docs/advanced/document-store/weaviate.md
davidbp: Note that this is their own operators , not a translation maintained by us

docs/advanced/document-store/annlite.md
```diff
@@ -48,3 +48,80 @@ The following configs can be set:
...
+| Name              | Description            | Equivalent Python operator |
+|-------------------|------------------------|----------------------------|
+| `$gte`            | Greater or equal to    | `>=`                       |
+| `$gt`             | Greater than           | `>`                        |
+| `$lte`            | Less or equal to       | `<=`                       |
+| `$lt`             | Less than )            | `<`                        |
+| `$eq`             | Equal to               | `==`                       |
+| `$neq`            | Not equal to           | `!=`                       |
+
+filters can be constructing following the guidelines provided in [the AnnLite source repository](https://github.com/jina-ai/annlite).
```
alaeddine-13, line 67:
```suggestion
Filters can be constructed following the guidelines provided in [the AnnLite source repository](https://github.com/jina-ai/annlite).
```

docs/advanced/document-store/annlite.md
```diff
@@ -48,3 +48,80 @@ The following configs can be set:
...
+    da.extend(
+        [
+            Document(id=f'r{i}', embedding=i * np.ones(n_dim), tags={'price': i})
+            for i in range(10)
+        ]
+    )
+```
+
+Consider we want the nearest vectors to the embedding `[8. 8. 8.]`, with the restriction that
+prices must follow a filter. As an example, let us consider that retrieved documents must have `price` value lower
```
alaeddine-13, line 97:
```suggestion
prices must follow a filter. As an example, le's consider that retrieved documents must have `price` value lower
```

docs/advanced/document-store/annlite.md
```diff
@@ -48,3 +48,80 @@ The following configs can be set:
...
+or equal than `max_price`. We can encode this information in annlite using `filter = {'price': {'$lte': max_price}}`.
+
+Then the search with the proposed filter can implemented and used with the following code:
+
+```python
+max_price = 7
+n_limit = 4
+
+np_query = np.ones(n_dim) * 8
```
alaeddine-13, line 106:
```suggestion
np_query = np.array([8, 8, 8])
```
maybe more readable like this

docs/advanced/document-store/annlite.md
```diff
@@ -48,3 +48,80 @@ The following configs can be set:
...
+print(f'\nQuery vector: \t{np_query}')
+
+filter = {'price': {'$lte': max_price}}
+results = da.find(np_query, filter=filter, limit=n_limit)
+print('\nEmbeddings Nearest Neighbours with "price" at most 7:\n')
+for embedding, price in zip(results.embeddings, results[:, 'tags__price']):
+    print(f'\tembedding={embedding},\t price={price}')
+```
+
+This would print
```
alaeddine-13, line 116:
```suggestion
This would print:
```

alaeddine-13 (approved):

JoanFM (approved):

PR #339, opened 2022-05-12
feat: added support for all Weaviate HNSW configuration parameters

jjlatval:
The biggest reason for this PR is to enable DocArray to import large datasets (2M+ objects) to Weaviate backend (see https://weaviate.io/developers/weaviate/current/architecture/resources.html#imports-slowed-down-after-crossing-2m-objects---what-can-i-do). The current implementation does not allow passing relevant configuration parameters for Weaviate.

This PR adds support for all currently listed Weaviate HNSW parameters based on the documentation available here: https://weaviate.io/developers/weaviate/current/vector-index-plugins/hnsw.html#how-to-use-hnsw-and-parameters

### DIFF
--- a/docarray/array/storage/weaviate/backend.py
+++ b/docarray/array/storage/weaviate/backend.py
@@ -37,13 +37,13 @@ class WeaviateConfig:
     ef_construction: Optional[int] = None
     timeout_config: Optional[Tuple[int, int]] = None
     max_connections: Optional[int] = None
-    dynamic_ef_min: Optional[int] = field(default=100)
-    dynamic_ef_max: Optional[int] = field(default=500)
-    dynamic_ef_factor: Optional[int] = field(default=8)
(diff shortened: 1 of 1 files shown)


docarray/array/storage/weaviate/backend.py
```diff
@@ -32,11 +32,18 @@ class WeaviateConfig:
     protocol: Optional[str] = field(default='http')
     name: Optional[str] = None
     serialize_config: Dict = field(default_factory=dict)
-    n_dim: Optional[int] = None  # deprecated, not used anymore since weaviate 1.10
+    # vectorIndexConfig parameters
```
JoanFM, line 35: we may need to keep n_dim to guarantee backward compatibility, maybe it is clear to which parameter it should correspond?

JoanFM (changes requested):

jjlatval:
No point in fixing commit message linting with excessive commits, will open a new PR

PR #367, opened 2022-05-24
fix: update qdrant to support python 3.7

alaeddine-13:
Revert "feat: make distance metric configurable in Weaviate backend"


davidbp (approved):
LGTM

alaeddine-13:
> LGTM, I suspect you will add in the config the distance right?

no it is not in relation to the distance parameter. I was just trying to revert to check why CI is breaking

Issue #351, opened 2022-05-19
Add warning if users try to set a vector with ndim > 1 to embeddings


alaeddine-13:
If a user tries to add a vector with ndim >1 to embeddings, finding vectors in storage backends will fail for them without knowing the extact reason.
It would be great to warn them about this behavior.

alaeddine-13:
Closing because it's not possible to achieve without a performance overhead

PR #369, opened 2022-05-24
feat: add warning for non vector embeddings

samsja:
Goals : warning user when instantiating a Document with an embedding which is not a vector (i.e one dimension ndarray)

fix : https://github.com/jina-ai/docarray/issues/351

### DIFF
--- a/docarray/document/mixins/property.py
+++ b/docarray/document/mixins/property.py
@@ -1,7 +1,9 @@
 import mimetypes
+import warnings
 from typing import TYPE_CHECKING, Optional
 
 from ._property import _PropertyMixin
+from ...math.ndarray import is_vector
 
 if TYPE_CHECKING:
     from ...typing import DocumentContentType, ArrayType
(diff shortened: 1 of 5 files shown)


docarray/document/mixins/property.py
```diff
@@ -87,6 +89,16 @@ def matches(self, value: 'DocumentArray'):
 
         self._data.matches = value
 
+    @_PropertyMixin.embedding.setter
+    def embedding(self, value: 'ArrayType'):
+
+        if is_vector(value):
```
JoanFM, line 95: this can be a little overkill for every document.

JoanFM (changes requested):
I am thinking maybe we should not do this

JoanFM (commented):
Can we please benchmark the code?

docarray/document/mixins/property.py
samsja: Yes I agree

samsja:
> 
Here is some benchmark, the feature is not fully implemented yet (still a bug when creating an empty array) but before I go further and fix all the test we can decide if this overhead ( + 35%) is affordable.

Screenshot from 2022-05-25 16-19-49

@alaeddine-13 @JoanFM @davidbp

I would say that the difference is too big and we should not warn here

samsja:
Conclusion : the overhead is not acceptable ( + 35%) on Document creation, therefore we drop the warning

PR #371, opened 2022-05-26
feat: bump protobuf

JoanFM:
closes: https://github.com/jina-ai/docarray/issues/376


hanxiao:
@alaeddine-13 please take care of this PR thanks, we have to move forward. It doesn't make sense to be blocked by paddle because of the lack of maintenance on their side. Refactor the tests or in the worst case, remove paddle tests (or even functions). Move this PR forward at any cost.

alaeddine-13:
we're proceeding like so:

1. create another protoc gen image with new version
2. build protos in 2 versions (old and new)
3. in docarray and jina, dynamically find the version of protobuf and accordingly, use the appropriate protobuf file

docarray/proto/__init__.py
```diff
@@ -0,0 +1,12 @@
+from google.protobuf import __version__ as __pb__version__
+
+if __pb__version__.startswith('4'):
```
hanxiao, line 3: aha, smart

hanxiao (approved):

hanxiao (changes requested):
could you also change `__version__ = 0.16.0` so 

after this merge,
1. the corresponding PR in core has to be merged, with version restriction of `docarray >=0.16.0` and its own `__version=3.18.0`
2. docarray release `0.16.0`
3. jina release `3.8.0`

alaeddine-13:
> 



> could you also change `__version__ = 0.16.0` so
> 
> after this merge,
> 
> 1. the corresponding PR in core has to be merged, with version restriction of `docarray >=0.16.0` and its own `__version=3.18.0`
> 2. docarray release `0.16.0`
> 3. jina release `3.8.0`

I agree on this
old jina versions will no more be compatible with new docarray and installing old versions of jina becomes buggy
the dependency here should be fixed to use docarray~=0.16.0: https://github.com/jina-ai/jina/blob/master/extra-requirements.txt#L35

hanxiao:
i haven't looked at the protobuf docs generation, is it affected?

PR #378, opened 2022-05-30
feat: add filter to match method

samsja:
This PR add the ability to do pre-filtering when calling `match`. Under the hood it passed the filter to the `find` function.

```python
 doc.match(da, filter=filter)
```

- [ ] code
- [ ] tests
- [ ] docs

### DIFF
--- a/docarray/array/mixins/match.py
+++ b/docarray/array/mixins/match.py
@@ -1,4 +1,4 @@
-from typing import Optional, Union, Callable, Tuple, TYPE_CHECKING
+from typing import Optional, Union, Callable, Tuple, TYPE_CHECKING, Dict
 
 if TYPE_CHECKING:
     import numpy as np
@@ -20,6 +20,7 @@ class MatchMixin:
         metric_name: Optional[str] = None,
         batch_size: Optional[int] = None,
         exclude_self: bool = False,
+        filter: Optional[Dict] = None,
         only_id: bool = False,
         use_scipy: bool = False,
         device: str = 'cpu',
@@ -50,6 +51,7 @@ class MatchMixin:
             elements. When `darray` is big, this can significantly speedup the computation.
         :param exclude_self: if set, Documents in ``darray`` with same ``id`` as the left-hand values will not be
                         considered as matches.
+        :param filter: filter query used for pre-filtering or filtering
         :param only_id: if set, then returning matches will only contain ``id``
         :param use_scipy: if set, use ``scipy`` as the computation backend. Note, ``scipy`` does not support distance
             on sparse matrix.
@@ -76,6 +78,7 @@ class MatchMixin:
             metric_name=metric_name,
             batch_size=batch_size,
             exclude_self=exclude_self,
+            filter=filter,
             only_id=only_id,
             use_scipy=use_scipy,
             device=device,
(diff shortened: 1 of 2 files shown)


docarray/array/mixins/match.py
```diff
@@ -50,6 +51,7 @@ def match(
             elements. When `darray` is big, this can significantly speedup the computation.
         :param exclude_self: if set, Documents in ``darray`` with same ``id`` as the left-hand values will not be
                         considered as matches.
+        :param filter: filter query used for pre-filtering or filtering
```
alaeddine-13, line 54:
```suggestion
        :param filter: filter query used for pre-filtering
```
In match we only use filter for pre-filtering right ?

tests/unit/array/mixins/test_match.py
```diff
@@ -577,3 +577,120 @@ def test_match_ensure_scores_unique():
...
+                    lambda operator, threshold: {'price': {operator: threshold}},
+                    numeric_operators_annlite,
+                    operator,
+                ]
+            )
+            for operator in numeric_operators_annlite.keys()
+        ],
+    ],
+)
+def test_search_match_pre_filtering(
```
alaeddine-13, line 668:
```suggestion
def test_match_pre_filtering(
```

alaeddine-13 (commented):
a few suggestions and it would be great if we update the docs

alaeddine-13:
Let's put docs in this section:
https://docarray.jina.ai/fundamentals/documentarray/matching/
basicly you can say that match and find support pre-filtering but it is specific to some backends and that you can check in the Document store section whether the backend supports pre-filtering or not. (also warn users that this is not supported in the in-memory backend)

docs/fundamentals/documentarray/matching.md
```diff
@@ -155,6 +155,26 @@ Note that framework is auto-chosen based on the type of `.embeddings`. For examp
 
 By default `A.match(B)` will copy the top-K matched Documents from B to `A.matches`. When these matches are big, copying them can be time-consuming. In this case, one can leverage `.match(..., only_id=True)` to keep only {attr}`~docarray.Document.id`.
 
+### Pre filtering
+
+Both `match` and `find` support pre-filtering by passing the `filter` argument to the function.
```
JohannesMessner, line 160:
```suggestion
Both `match` and `find` support pre-filtering by passing a `filter` argument to the method.
```

docs/fundamentals/documentarray/matching.md
```diff
@@ -155,6 +155,26 @@ Note that framework is auto-chosen based on the type of `.embeddings`. For examp
...
+
+Pre-filtering is an advance approximate nearest neighbors feature  that allow to efficiently retrieve the nearest vector
+which respect the filtering condition. In contrast, post-filtering in the naive approach where you first retrieve the 
+nearest neighbors and then discard all the candidates that does not respect the filter condition.
```
JohannesMessner, line 164:
```suggestion
Pre-filtering is an advanced approximate nearest neighbors feature that allows to efficiently retrieve the nearest vectors
that respect the filtering condition.

In contrast, post-filtering in the naive approach where you first retrieve the 
nearest neighbors and then discard all the candidates that do not respect the filter condition.
```

alaeddine-13 (approved):

JoanFM (approved):

davidbp (approved):
LGTM

Issue #387, opened 2022-06-01
filter by equal not working with Qdrant


samsja:
## The issue

Pre-filtering with the operator equal is not working with qdrant backend.

For instance using the filter `{'must': [{'key': 'price', 'value': {'eq': 7}}]}` will not match any docs even though one has the price 7. See the following code snippet to reproduce it.

```python
from docarray import Document, DocumentArray
import numpy as np

n_dim = 3
distance = 'euclidean'

da = DocumentArray(
    storage='qdrant',
    config={'n_dim': n_dim, 'columns': [('price', 'int')], 'distance': distance},
)

with da:
    da.extend(
        [
            Document(id=f'r{i}', embedding=i * np.ones(n_dim), tags={'price': i})
            for i in range(10)
        ]
    )

max_price = 7
n_limit = 4

np_query = np.ones(n_dim) * 8

filter = {'must': [{'key': 'price', 'value': {'eq': 7}}]}
results = da.find(np_query, filter=filter, limit=n_limit)

print(len(results))
>>> 0
```

The tests passed but are actually failing. Indeed the assert [here](https://github.com/jina-ai/docarray/blob/1a3fb27ec9b3eacb02d5fa764d64778c5bb88236/tests/unit/array/mixins/test_find.py#L308) passed only because the results are empty therefore doing `assert all([])` passed.

CC: @davidbp 

## Possible solution:

Maybe the filter is not written properly

JoanFM:
Hey @samsja let's start by opening a draft by changing the test to check the size of the results

samsja:
> Hey @samsja let's start by opening a draft by changing the test to check the size of the results

https://github.com/jina-ai/docarray/pull/388

PR #388, opened 2022-06-01
fix(tests): eq and neq operator for qdrant filtering 

samsja:
# Context
Pre-filtering with the operator equal is not working with qdrant backend. See full issue for details https://github.com/jina-ai/docarray/issues/387

# Solution
To do filtering with with `eq` and `neq` one should use the `match` keyword.

for `eq`:

`filter = {'must': [{'key': 'price', 'match': {'value': 7}}]}`

for `neq`:

filter = {'must_not': [{'key': 'price', 'match': {'value': 7}}]}

# What this pr do
This pr fix the test to use this notation for the filter


davidbp (approved):
LGTM

Issue #391, opened 2022-06-07
Docoarray protobuff convert int to float


samsja:
# Context

When a DA is serialize then deserialize in protobuf, if the tags are `int` then there are converted to `float`.

This is an issue when using docarray with Jina. Especially for VectorDB indexer. For instance with the WeaviateIndexer:

```python
from jina import Flow
from docarray import Document, DocumentArray

f = Flow().add(
    uses=WeaviateIndexer,
    uses_with={
        'name': 'Test',
        'n_dim': 3,
        'columns': [('price', 'int')],
    },
)

docs = DocumentArray(
    [
        Document(id=f'r{i}', embedding=np.random.rand(3), tags={'price': int(i)})
        for i in range(50)
    ]
)
with f:
        f.index(docs)
```

This will fail because the Weaviate backend expect the docs to have the tags price as `int` but the Executor will receive them as `float` because of the protobuff de/serialization.

## code to reproduce:
see the following code to reproduce 

```python
from docarray import DocumentArray, Document


def test_type_protobuf():

    da_int = DocumentArray([Document(tags={'price': i}) for i in range(10)])
    da_receive = DocumentArray.from_protobuf(da_int.to_protobuf())

    for doc in da_receive:
        assert type(doc.tags['price']) == int
```
Output:

```shell
FAILED scratch_1.py::test_type_protobuf - AssertionError: assert <class 'float'> == int
```

JoanFM:
Can you please ellaborate? this is a common protobuff thing that has no solution I'd say

PR #392, opened 2022-06-07
fix: cast column to its correct type before insertion in database

alaeddine-13:
closes: https://github.com/jina-ai/docarray/issues/391

### DIFF
--- a/docarray/array/storage/annlite/backend.py
+++ b/docarray/array/storage/annlite/backend.py
@@ -34,9 +34,9 @@ class BackendMixin(BaseBackendMixin):
     """Provide necessary functions to enable this storage backend."""
 
     TYPE_MAP = {
-        'str': TypeMap(type='string', converter=str),
-        'float': TypeMap(type='number', converter=float),
-        'int': TypeMap(type='int', converter=_safe_cast_int),
+        'str': TypeMap(type='TEXT', converter=str),
+        'float': TypeMap(type='float', converter=float),
+        'int': TypeMap(type='integer', converter=_safe_cast_int),
     }
 
     def _map_embedding(self, embedding: 'ArrayType') -> 'ArrayType':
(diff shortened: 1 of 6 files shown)


JoanFM (changes requested):
some test should be added/modified

docarray/array/storage/annlite/backend.py
```diff
@@ -33,7 +33,11 @@ class AnnliteConfig:
 class BackendMixin(BaseBackendMixin):
     """Provide necessary functions to enable this storage backend."""
 
-    TYPE_MAP = {'str': 'TEXT', 'float': 'float', 'int': 'integer'}
+    TYPE_MAP = {
+        'str': ('TEXT', str),
```
samsja, line 37: Could we make it a name tuple ? Otherwise it is not straightforward to understand the rest of the code

docarray/array/storage/base/backend.py
```diff
@@ -25,10 +25,13 @@ def _get_storage_infos(self) -> Optional[Dict]:
     def _map_id(self, _id: str) -> str:
         return _id
 
+    def _map_column(self, value, col_type) -> str:
+        return self.TYPE_MAP[col_type][1](value)
```
samsja, line 29: If it was a namedtuple we could have something like 
```suggestion
        return self.TYPE_MAP[col_type].type(value)
```

docarray/array/storage/base/backend.py
```diff
@@ -25,10 +25,13 @@ def _get_storage_infos(self) -> Optional[Dict]:
...
+
     def _map_embedding(self, embedding: 'ArrayType') -> 'ArrayType':
         from ....math.ndarray import to_numpy_array
 
         return to_numpy_array(embedding)
 
     def _map_type(self, col_type: str) -> str:
-        return self.TYPE_MAP[col_type]
+        return self.TYPE_MAP[col_type][0]
```
samsja, line 37: If it was a namedtuple we could have something like
```suggestion
        return self.TYPE_MAP[col_type].type_identifier
```

docarray/array/storage/annlite/backend.py
alaeddine-13: done

JoanFM (approved):

PR #403, opened 2022-06-16
feat: draft secondary index

davidbp:
This PR is a draft for storing secondary indices as a dict with different docarrays

**ToDo**:

- [x] .find()
- [x] .match()
- [x] add
- [x] delete
- [x] update
- [x] bug: multi-modal access path. Fix here: #448
- [ ] Decide on syntax and argument names

Design doc: https://docs.google.com/presentation/d/1rntTa1Ur2WmdAvOUEI01l2WeNSWgtqk2xjth_gdIr50/edit#slide=id.g13443d1ddb2_1_23


JohannesMessner:
I considering the following changes to parameter names:
`secondary_index` -> `subindex`
`secondary_index_configs` -> `subindex_configs`
`da.find(query=..., secondary_index='@.[image]')` -> `da.find(query=..., find_on='@.[image]')`

Technically, `secondary_index` is probably more correct than `subindex`, but I much prefer the brevity

JohannesMessner:
Also related to this: with this nomenclature on `find()` we could enable the same syntax on in-memory-DA without causing confusion. It would unify the experience between in-memory and document store.

The implementation could be as simple as this, since no actual secondary index is needed:

```python
def find(self, ..., find_on=None):
    if find_on:
        return self[find_on].find(...)
```

JohannesMessner:
Continued here: https://github.com/jina-ai/docarray/pull/456

PR #404, opened 2022-06-17
fix: convert mesh tensor to nparray

ZiniuYu:
Summary:

This PR convert mesh tensor get from `load_uri_to_point_cloud_tensor` to numpy array.

Description:

We use `Trimesh` APIs in this method and it returns a datatype of `<class 'trimesh.caching.TrackedArray'>` which is a subclass of numpy array ([reference](https://trimsh.org/trimesh.caching.html#trimesh.caching.tracked_array)). This may cause some data type problems in downstream applications.

### DIFF
--- a/docarray/document/mixins/mesh.py
+++ b/docarray/document/mixins/mesh.py
@@ -1,5 +1,7 @@
 from typing import TYPE_CHECKING
 
+import numpy as np
+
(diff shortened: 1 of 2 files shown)


numb3r3 (approved):
LGTM

hanxiao (approved):

PR #411, opened 2022-06-22
fix: elastic search offset management on doc deletion

alphinside:


alphinside:
> Thanks for working on this @alphinside ! Could you add a test that shows that currently this is not working on master and your PR solves it?

sure, on it

docarray/array/storage/elastic/backend.py
```diff
@@ -179,6 +179,23 @@ def _update_offset2ids_meta(self):
             r = bulk(self._client, requests)
             self._client.indices.refresh(index=self._index_name_offset2id)
 
+            # Clean trailing unused offsets
+            offset_count = self._client.count(index=self._index_name_offset2id)
+            unused_offsets = range(len(self._offset2ids.ids), offset_count["count"])
```
JoanFM, line 184:
```suggestion
            unused_offsets = range(len(self._offset2ids.ids), offset_count["count"])
```
```suggestion
            unused_offsets = range(len(self._offset2ids.ids), offset_count['count'])
```

JoanFM (changes requested):
Hey @alphinside, thank you very much for the contribuition, may you add a test showcasing the case that was failing?

alphinside:
added the tests, please review ya @JoanFM @davidbp

tests/unit/array/storage/elastic/test_del.py
```diff
@@ -0,0 +1,102 @@
+from unittest.mock import MagicMock
```
JoanFM, line 1: we do not use `unittest` mock, we prefer to use `monkeypatch` from `pytest` since is our own testing framework

setup.py
```diff
@@ -102,6 +102,8 @@
             'weaviate-client~=3.3.0',
             'annlite>=0.3.2',
             'elasticsearch>=8.2.0',
+            'qdrant-client~=0.7.3',
+            'strawberry-graphql',
```
JoanFM, line 106: why is this dependency needed?

tests/unit/array/storage/elastic/test_del.py
```diff
@@ -0,0 +1,102 @@
...
+class MockElasticClient:
+    def __init__(self, **kwargs):
+        self.indices = MockElasticIndices()
+
+    def count(self, **kwargs):
+        return {"count": len(MOCK_ES_OFFSET_INDEX)}
+
+
+def mock_bulk_index(*args, **kwargs):
+    global ES_BULK_COUNT
```
JoanFM, line 62: is there a way to avoid `global` variables?

JoanFM (changes requested):

tests/unit/array/storage/elastic/test_del.py
alphinside: okay will try to update this

alphinside: done

tests/unit/array/storage/elastic/test_del.py
alphinside: done

setup.py
JoanFM: For now yes, If this is true, it will be solved in a different scope

tests/unit/array/storage/elastic/test_del.py
```diff
@@ -0,0 +1,98 @@
...
+        for request in requests:
+            if request["_op_type"] == "index":
+                MOCK_ES_OFFSET_INDEX[request["_id"]] = {
+                    "_id": request["_id"],
+                    "_source": {"blob": request["blob"]},
+                }
+            elif request["_op_type"] == "delete":
+                del MOCK_ES_OFFSET_INDEX[request["_id"]]
+
+        if ES_BULK_COUNT == 1:  # 1st bulk api call, update offset index
```
JoanFM, line 75: this does not seem to make sense, this ES_BULK_COUNT`seems not to be used

JoanFM (changes requested):
Please use single quote instead of double quote in the code and in tests

JoanFM (changes requested):
Also, please add some "integration test" (We will have to refactor tests) where u use the `elastic` server to prove the error that lead to the creation of this PR

tests/unit/array/storage/elastic/test_del.py
alphinside: sorry forget to update this, already updated

setup.py
alphinside: done

JoanFM:
Hey @alphinside ,

Please fix the `commit-lint` error by following our [CONTRIBUTING](https://github.com/jina-ai/docarray/blob/main/CONTRIBUTING.md) guidelines.

As for the integration tests. In our tests, there is a`docker-compose` that gets started before every module is tested. This docker compose will start an elasticsearch service that u can use to test the issue that you raised and validate that is fixed.

alphinside:
> Hey @alphinside ,
> 
> Please fix the `commit-lint` error by following our [CONTRIBUTING](https://github.com/jina-ai/docarray/blob/main/CONTRIBUTING.md) guidelines.
> 
> As for the integration tests. In our tests, there is a`docker-compose` that gets started before every module is tested. This docker compose will start an elasticsearch service that u can use to test the issue that you raised and validate that is fixed.

I see so in my local, I can test it by running the docker compose and run the pytest, is that correct?

JoanFM:
> > Hey @alphinside ,
> > Please fix the `commit-lint` error by following our [CONTRIBUTING](https://github.com/jina-ai/docarray/blob/main/CONTRIBUTING.md) guidelines.
> > As for the integration tests. In our tests, there is a`docker-compose` that gets started before every module is tested. This docker compose will start an elasticsearch service that u can use to test the issue that you raised and validate that is fixed.
> 
> I see so in my local, I can test it by running the docker compose and run the pytest, is that correct?

Yes you can

tests/unit/array/storage/elastic/test_add.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def __getitem__(self):
+        pass
+
+    def __setitem__(self):
+        pass
+
+
+def test_add_ids_that_already_exist():
+    es_seq_mixin = MockESMixin()
```
JoanFM, line 18: I think a best solution is to come with `monkeypatch` here and path the `ESBackend` Mixin or functions and wori wht the `da.add()` interface so that is clear what is being tested and what not. Here we are doing a lot of private magic to make the test work, this should be done inside the Mocking class itself

tests/unit/array/storage/elastic/test_del.py
```diff
@@ -0,0 +1,96 @@
...
+            if request['_op_type'] == 'index':
+                MOCK_ES_OFFSET_INDEX[request['_id']] = {
+                    '_id': request['_id'],
+                    '_source': {'blob': request['blob']},
+                }
+            elif request['_op_type'] == 'delete':
+                del MOCK_ES_OFFSET_INDEX[request['_id']]
+
+        if cls.bulk_call_count == 1:  # 1st bulk api call, update offset index
+            assert MOCK_ES_OFFSET_INDEX == BUG_OFFSET_INDEX_AFTER_DELETION
```
JoanFM, line 74: I think this assert is not important at all and only provides noise

tests/unit/array/storage/elastic/test_del.py
```diff
@@ -0,0 +1,96 @@
...
+    2: {'_id': 2, '_source': {'blob': 'r2'}},
+    3: {'_id': 3, '_source': {'blob': 'r3'}},
+    4: {'_id': 4, '_source': {'blob': 'r4'}},
+    5: {'_id': 5, '_source': {'blob': 'r5'}},
+    6: {'_id': 6, '_source': {'blob': 'r6'}},
+    7: {'_id': 7, '_source': {'blob': 'r7'}},
+    8: {'_id': 8, '_source': {'blob': 'r8'}},
+}
+
+BUG_OFFSET_INDEX_AFTER_DELETION = {
```
JoanFM, line 16: I think this should not be considered.

JoanFM (changes requested):

tests/unit/array/storage/elastic/test_add.py
alphinside: yep. instead I will change this using integration test, so that the main issue shows directly

tests/unit/array/storage/elastic/test_del.py
alphinside: noted

tests/unit/array/storage/elastic/test_del.py
alphinside: noted

alphinside:
I'll submit new PR using your standard from your contributing guideline

Issue #412, opened 2022-06-22
Inconsistency when add/extend/delete documents with same ID using elastic search backend


alphinside:
Hi, this issue is partially related to https://github.com/jina-ai/docarray/issues/407

So I try to execute the following code twice 

```python
from docarray import Document, DocumentArray
import numpy as np

n_dim = 3

da = DocumentArray(
    storage="elasticsearch",
    config={
        "n_dim": n_dim,
        "columns": [("price", "int")],
        "distance": "l2_norm",
        "index_name": "old_stuff",
    },
)

with da:
    da.extend(
        [
            Document(id=f"r{i}", embedding=i * np.ones(n_dim), tags={"price": i})
            for i in range(10)
        ]
    )

print("\nIndexed Prices:\n")
for embedding, price in zip(da.embeddings, da[:, "tags__price"]):
    print(f"\tembedding={embedding},\t price={price}")

print(f"\nIndexed documents:{len(da)}")
```

The first time execution, output is 

```shell
Indexed Prices:

        embedding=[0. 0. 0.],    price=0
        embedding=[1. 1. 1.],    price=1
        embedding=[2. 2. 2.],    price=2
        embedding=[3. 3. 3.],    price=3
        embedding=[4. 4. 4.],    price=4
        embedding=[5. 5. 5.],    price=5
        embedding=[6. 6. 6.],    price=6
        embedding=[7. 7. 7.],    price=7
        embedding=[8. 8. 8.],    price=8
        embedding=[9. 9. 9.],    price=9

Indexed documents:10
```

The second time execution, output is

```shell
Indexed Prices:

        embedding=[0. 0. 0.],    price=0
        embedding=[1. 1. 1.],    price=1
        embedding=[2. 2. 2.],    price=2
        embedding=[3. 3. 3.],    price=3
        embedding=[4. 4. 4.],    price=4
        embedding=[5. 5. 5.],    price=5
        embedding=[6. 6. 6.],    price=6
        embedding=[7. 7. 7.],    price=7
        embedding=[8. 8. 8.],    price=8
        embedding=[9. 9. 9.],    price=9
        embedding=[0. 0. 0.],    price=0
        embedding=[1. 1. 1.],    price=1
        embedding=[2. 2. 2.],    price=2
        embedding=[3. 3. 3.],    price=3
        embedding=[4. 4. 4.],    price=4
        embedding=[5. 5. 5.],    price=5
        embedding=[6. 6. 6.],    price=6
[...]

Issue #407, opened 2022-06-21
How to delete a document inside DocArray by id correctly?


alphinside:
Hi, I'm trying to delete several documents using this syntax and elastic backend.

```python
from docarray import Document, DocumentArray
import numpy as np

n_dim = 3

da = DocumentArray(
    storage="elasticsearch",
    config={
        "n_dim": n_dim,
        "columns": [("price", "int")],
        "distance": "l2_norm",
        "index_name": "old_stuff",
    },
)

with da:
    da.extend(
        [
            Document(id=f"r{i}", embedding=i * np.ones(n_dim), tags={"price": i})
            for i in range(10)
        ]
    )

print("\nIndexed Prices:\n")
for embedding, price in zip(da.embeddings, da[:, "tags__price"]):
    print(f"\tembedding={embedding},\t price={price}")

del da[["r0","r1"]]

print("\nIndexed Prices:\n")
for embedding, price in zip(da.embeddings, da[:, "tags__price"]):
    print(f"\tembedding={embedding},\t price={price}")
```

when I run those code, It print the correct output ( 2 doc deleted )

```shell
Indexed Prices:

        embedding=[0. 0. 0.],    price=0
        embedding=[1. 1. 1.],    price=1
        embedding=[2. 2. 2.],    price=2
        embedding=[3. 3. 3.],    price=3
        embedding=[4. 4. 4.],    price=4
        embedding=[5. 5. 5.],    price=5
        embedding=[6. 6. 6.],    price=6
        embedding=[7. 7. 7.],    price=7
        embedding=[8. 8. 8.],    price=8
        embedding=[9. 9. 9.],    price=9

Indexed Prices:

        embedding=[2. 2. 2.],    price=2
        embedding=[3. 3. 3.],    price=3
        embedding=[4. 4. 4.],    price=4
        embedding=[5. 5. 5.],    price=5
        embedding=[6. 6. 6.],    price=6
        embedding=[7. 7. 7.],    price=7
        embedding=[8. 8. 8.],    price=8
        embedding=[9. 9. 9.],    price=9
```

but when I execute the following code( similar with above but without adding and delete data, using the already initialized index ), the output become like this ( duplicate doc ).

The code 

```python
from docarray import Document, DocumentArray
import numpy as np

n_dim = 3

da = DocumentArray(
    storage="elasticsearch",
    config={
        "n_dim": n_dim,
        "columns": [("price", "int")],
        "distance": "l2_norm",
[...]

PR #416, opened 2022-06-23
fix(elastic): update offset2ids management

alphinside:
Goals:

Solving inconsistent offset issue on deletion https://github.com/jina-ai/docarray/issues/407.

I've found that after index deletion, update offset persistence has missed on truncating the trailing offset if the len of ids is less than initial length. So in this PR I submit logic to reevaluate offset length compared to in memory offset ids list and bulk delete offset which is more than the offset ids len

Also solving inconsitent offset issue on extend/add https://github.com/jina-ai/docarray/issues/412

Elasticsearch handle document indexing as upsert operation when indexing using same id, however the extend logic didn't using this logic, so I update the offset indexing to not extend id that already exist

- ...
- ...

### DIFF
--- a/tests/unit/array/storage/elastic/test_add.py
+++ b/tests/unit/array/storage/elastic/test_add.py
@@ -1,7 +1,7 @@
 from docarray import Document, DocumentArray
 
 
-def test_add_ignore_existing_doc_id():
+def test_add_ignore_existing_doc_id(start_storage):
     elastic_doc = DocumentArray(
         storage='elasticsearch',
         config={
--- a/tests/unit/array/storage/elastic/test_add.py
+++ b/tests/unit/array/storage/elastic/test_add.py
@@ -40,10 +40,3 @@ def test_add_ignore_existing_doc_id(start_storage):
 
     assert len(elastic_doc) == len(elastic_doc[:, 'embedding'])
     assert len(elastic_doc) == indexed_offset_count
-
-    elastic_doc._client.indices.delete(
-        index=elastic_doc._index_name_offset2id, ignore=[400, 404]
-    )
-    elastic_doc._client.indices.delete(
-        index=elastic_doc._config.index_name, ignore=[400, 404]
-    )
--- a/tests/unit/array/storage/elastic/test_add.py
+++ b/tests/unit/array/storage/elastic/test_add.py
@@ -1,0 +1,49 @@
+from docarray import Document, DocumentArray
+
+
+def test_add_ignore_existing_doc_id():
+    elastic_doc = DocumentArray(
+        storage='elasticsearch',
+        config={
+            'n_dim': 3,
+            'columns': [('price', 'int')],
+            'distance': 'l2_norm',
+            'index_name': 'test_delete',
+        },
+    )
+
+    with elastic_doc:
+        elastic_doc.extend(
+            [
+                Document(id=f'r0', embedding=[0, 0, 0]),
+                Document(id=f'r1', embedding=[1, 1, 1]),
+                Document(id=f'r2', embedding=[2, 2, 2]),
+                Document(id=f'r3', embedding=[3, 3, 3]),
+                Document(id=f'r4', embedding=[4, 4, 4]),
+            ]
+        )
+
+    with elastic_doc:
+        elastic_doc.extend(
+            [
+                Document(id=f'r0', embedding=[0, 0, 0]),
+                Document(id=f'r2', embedding=[2, 2, 2]),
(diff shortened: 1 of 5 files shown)


alphinside:
please re-review @JoanFM

tests/unit/array/storage/elastic/test_add.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+
+    indexed_offset_count = elastic_doc._client.count(
+        index=elastic_doc._index_name_offset2id
+    )['count']
+
+    assert len(elastic_doc) == len(elastic_doc[:, 'embedding'])
+    assert len(elastic_doc) == indexed_offset_count
+
+    elastic_doc._client.indices.delete(
```
JoanFM, line 44: do not erase here, just create a random name for the index at the beginning so u do not need to care about this. In any case if test fails the index will be polluted and other tests may fail

tests/unit/array/storage/elastic/test_del.py
```diff
@@ -0,0 +1,54 @@
...
+    assert len(elastic_doc._offset2ids.ids) == indexed_offset_count
+
+    for id in expected_offset_after_del:
+        expected_offset = str(expected_offset_after_del.index(id))
+        actual_offset_index = elastic_doc._client.search(
+            index=elastic_doc._index_name_offset2id, query={'match': {'blob': id}}
+        )['hits']['hits'][0]['_id']
+        assert actual_offset_index == expected_offset
+
+    elastic_doc._client.indices.delete(
```
JoanFM, line 49: same here, nothing after asserts should be needed

JoanFM (changes requested):

tests/unit/array/storage/elastic/test_add.py
alphinside: noted

tests/unit/array/storage/elastic/test_del.py
alphinside: noted

tests/unit/array/storage/elastic/test_add.py
```diff
@@ -0,0 +1,42 @@
...
+                Document(id='r5', embedding=[2, 2, 2]),
+                Document(id='r6', embedding=[4, 4, 4]),
+            ]
+        )
+
+    indexed_offset_count = elastic_doc._client.count(
+        index=elastic_doc._index_name_offset2id
+    )['count']
+
+    assert len(elastic_doc) == len(elastic_doc[:, 'embedding'])
```
JoanFM, line 41: I would assert that `len(elastic_doc) == 7` also for extra security, this test otherwise would pass even with wrong behavior

tests/unit/array/storage/elastic/test_del.py
```diff
@@ -0,0 +1,47 @@
...
+    expected_offset_after_del = ['r2', 'r3', 'r4', 'r5', 'r6', 'r7']
+
+    with elastic_doc:
+        del elastic_doc[deleted_elmnts]
+
+    indexed_offset_count = elastic_doc._client.count(
+        index=elastic_doc._index_name_offset2id
+    )['count']
+
+    assert len(elastic_doc._offset2ids.ids) == indexed_offset_count
```
JoanFM, line 40: same here, what should be the length here?

JoanFM (changes requested):

tests/unit/array/storage/elastic/test_add.py
alphinside: Okay noted

tests/unit/array/storage/elastic/test_del.py
alphinside: yep will be updated

tests/unit/array/storage/elastic/test_add.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def test_add_ignore_existing_doc_id(start_storage):
+    elastic_doc = DocumentArray(
+        storage='elasticsearch',
+        config={
+            'n_dim': 3,
+            'columns': [('price', 'int')],
+            'distance': 'l2_norm',
+            'index_name': 'test_add',
```
JoanFM, line 11: can you perhaps randomize the index name or give the index name the exact same name as the test function name

JoanFM (changes requested):

tests/unit/array/storage/elastic/test_add.py
alphinside: oh sure will do

PR #425, opened 2022-06-30
feat: allow modalities of multimodal docs to be accessed

JohannesMessner:
This allows Documents that come from a multimodal dataclass to expose their multimodal attributes like any other 'native' attribute (check last line in the code snippet):

```python
MyText = TypeVar('MyText', bound=str)


def my_setter(value) -> 'Document':
    return Document(text=value + ' but custom!')


def my_getter(doc: 'Document'):
    return doc.text


@dataclass
class MyMultiModalDoc:
    avatar: Image
    description: Text
    heading: MyText = field(setter=my_setter, getter=my_getter, default='')


m = MyMultiModalDoc(avatar='testflow.jpg', description='hello, world', heading='hello, world')
d = Document(m)
print(d.heading)  # prints a Document with text='hello, world but custom!'
print(d.heading.text)  # prints 'hello, world but custom!'
```

This also extends to use inside of an Executor:

```python
class MyExec(Executor):
    @requests
    def foo(self, docs, **kwargs):
        model = ...  # embedding model
        for d in docs:
            d.heading.embed(model)
```

**Advanced usage example**

This also handles list-types, nested dataclasses, and list-types of nested data classes:

<details>
  <summary>Click to expand: Advanced example </summary>

```python
MyText = TypeVar('MyText', bound=str)


def my_setter(value) -> 'Document':
    return Document(text=value + ' but custom!')


def my_getter(doc: 'Document'):
    return doc.text


@dataclass
class InnerDoc:
    avatar: Image
    description: Text
    heading: MyText = field(setter=my_setter, getter=my_getter, default='')


@dataclass
class MyMultiModalDoc:
    avatar: Image
    description: Text
    heading_list: List[Text]
    other_doc: InnerDoc
    other_doc_list: List[InnerDoc]
    heading: MyText = field(setter=my_setter, getter=my_getter, default='')


inner_doc_list = [InnerDoc(avatar='testflow.jpg', description='hello, world', heading=f'{i} hello, world') for i in range(3)]

m = MyMultiModalDoc(avatar='testflow.jpg', description='hello, world', heading='hello, world',
                    heading_list=['hello', 'world'], other_doc=InnerDoc(avatar='testflow.jpg',
                                                                        description='inner hello, world',
                                                                        heading='inner hello, world'),
                    other_doc_list=inner_doc_list)
d = Document(m)
print(d.heading.text)  # returns 'hello, world but custom!'
print(d.heading_list.texts)  # returns ['hello', 'world']
print(d.other_doc)  # returns the inner doc
print(d.other_doc.heading.text)  # returns 'inner hello, world but custom!'
print(d.other_doc_list)  # returns DocArray of inner docs
print(d.other_doc_list[1].heading.text)  # returns '1 inner hello, world but custom!'
```
</details>

This should solve the issue of multimodal features not being easily usable inside of Jina.

**Still TODO:**
- [x] support for list-like multimodal types
- [x] support for nested dataclasse
- [x]~~make sure that `content` is always the desired return value~~
- [x] tests
- [ ]~~support custom getters~~ getters are only called when performng `MMDoc(d)`, not when doing `d.attr`. so not relevant here
- [x] support json and dict (de)serialization with this feature **Edit**: This is done separately here: jina-ai/docarray/issues/429]
- [ ] documentation

### DIFF
--- a/docarray/document/mixins/multimodal.py
+++ b/docarray/document/mixins/multimodal.py
@@ -141,10 +141,13 @@ class MultiModalMixin:
         return doc, attribute_type
 
     def __getattr__(self, attr):
-        has_metadata = (
-            super().__getattribute__('_data') is not None
-            and getattr(self._data, '_metadata') is not None
-        )
+        try:
+            data = super().__getattribute__('_data')
+            has_data = bool(data)
+        except AttributeError:
+            has_data = False
+
+        has_metadata = has_data and getattr(self._data, '_metadata') is not None
         if (
             has_metadata
             and self.is_multimodal
--- a/docarray/document/mixins/multimodal.py
+++ b/docarray/document/mixins/multimodal.py
@@ -139,3 +139,10 @@ class MultiModalMixin:
         else:
             raise ValueError(f'Unsupported type annotation')
         return doc, attribute_type
+
+    def __getattr__(self, attr):
+        if self.is_multimodal and attr in self._metadata['multi_modal_schema']:
+            content = self.get_multi_modal_attribute(attr)[:, 'content']
+            return content[0] if len(content) == 1 else content
+        else:
+            raise AttributeError(f'{self.__class__.__name__} has no attribute {attr}')
--- a/docarray/document/mixins/multimodal.py
+++ b/docarray/document/mixins/multimodal.py
@@ -116,7 +116,8 @@ class MultiModalMixin:
         position = self._metadata['multi_modal_schema'][attribute].get('position')
 
         if attribute_type in [AttributeType.DOCUMENT, AttributeType.NESTED]:
-            return DocumentArray([self.chunks[int(position)]])
+            d = DocumentArray([self.chunks[int(position)]])
+            return d
         elif attribute_type in [
             AttributeType.ITERABLE_DOCUMENT,
             AttributeType.ITERABLE_NESTED,
@@ -142,7 +143,17 @@ class MultiModalMixin:
 
     def __getattr__(self, attr):
         if self.is_multimodal and attr in self._metadata['multi_modal_schema']:
-            content = self.get_multi_modal_attribute(attr)[:, 'content']
-            return content[0] if len(content) == 1 else content
+            mm_attr_da = self.get_multi_modal_attribute(attr)
+            collected_attributes = []
+            for d in mm_attr_da:
+                if d.is_multimodal:  # comes from nested dataclass, add Document as is
+                    collected_attributes.append(d)
+                else:  # comes from a non-nested attribute, add the value
+                    collected_attributes.append(d.content)
+            return (
+                collected_attributes
+                if len(collected_attributes) > 1
+                else collected_attributes[0]
+            )
         else:
             raise AttributeError(f'{self.__class__.__name__} has no attribute {attr}')
--- a/docarray/document/mixins/multimodal.py
+++ b/docarray/document/mixins/multimodal.py
@@ -142,7 +142,15 @@ class MultiModalMixin:
         return doc, attribute_type
 
     def __getattr__(self, attr):
-        if self.is_multimodal and attr in self._metadata['multi_modal_schema']:
+        has_metadata = (
+            super().__getattribute__('_data') is not None
+            and getattr(self._data, '_metadata') is not None
+        )
+        if (
+            has_metadata
+            and self.is_multimodal
+            and attr in self._metadata['multi_modal_schema']
+        ):
             mm_attr_da = self.get_multi_modal_attribute(attr)
             collected_attributes = []
             for d in mm_attr_da:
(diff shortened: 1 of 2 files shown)


docarray/document/mixins/multimodal.py
```diff
@@ -139,3 +139,10 @@ def _from_obj(cls, obj, obj_type, field) -> typing.Tuple['Document', AttributeTy
         else:
             raise ValueError(f'Unsupported type annotation')
         return doc, attribute_type
+
+    def __getattr__(self, attr):
+        if self.is_multimodal and attr in self._metadata['multi_modal_schema']:
+            content = self.get_multi_modal_attribute(attr)[:, 'content']
+            return content[0] if len(content) == 1 else content
+        else:
```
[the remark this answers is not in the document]
JohannesMessner: no, `__getattr__` only gets called when 'normal' attribute access already didn't work. So no need to take car of the normal access. That's what makes it different form `__get_attribute__`, and apparently it is better practice to implement `__getattr__`

docarray/document/mixins/multimodal.py
```diff
@@ -139,3 +139,10 @@ def _from_obj(cls, obj, obj_type, field) -> typing.Tuple['Document', AttributeTy
         else:
             raise ValueError(f'Unsupported type annotation')
         return doc, attribute_type
+
+    def __getattr__(self, attr):
+        if self.is_multimodal and attr in self._metadata['multi_modal_schema']:
+            content = self.get_multi_modal_attribute(attr)[:, 'content']
```
JohannesMessner, line 145: I am not sure that `content` is in all cases what we want, this is something i still have to check

docarray/document/mixins/multimodal.py
```diff
@@ -139,3 +139,28 @@ def _from_obj(cls, obj, obj_type, field) -> typing.Tuple['Document', AttributeTy
...
+        ):
+            mm_attr_da = self.get_multi_modal_attribute(attr)
+            collected_attributes = []
+            for d in mm_attr_da:
+                if d.is_multimodal:  # comes from nested dataclass, add Document as is
+                    collected_attributes.append(d)
+                else:  # comes from a non-nested attribute, add the value
+                    collected_attributes.append(d.content)
+            return (
+                collected_attributes
```
JoanFM, line 161: can you point me to an example where the List would apply and when a single element would apply?

tests/unit/document/test_multi_modal.py
```diff
@@ -629,3 +629,74 @@ class MMDoc:
...
+        heading_list: List[Text]
+        heading: MyText = field(setter=my_setter, getter=my_getter, default='')
+
+    m = MyMultiModalDoc(
+        avatar='toydata/test.png',
+        description='hello, world',
+        heading='hello, world',
+        heading_list=['hello', 'world'],
+    )
+    d = Document(m)
```
JoanFM, line 656: can u add a test after serialization/deserialization in pickle and protobuf and see this still holds.

JoanFM (changes requested):

docarray/document/mixins/multimodal.py
JohannesMessner: You can see it in the test [here](https://github.com/jina-ai/docarray/pull/425/files#diff-daea3d13c630b543a87c4e19c4607effce9121a41c629bb4297b892070c56692R644-R648), if the type is `List[T]` then it will be a list, otherwise a single element. Same applies for nested dataclasses, in that case the return will be a (list of) document

JoanFM (changes requested):
I miss tests to validate and see how `Documents` are changed from the `Document` and not the `DataClass`.

Saying:

How do I set `d.header.text`? or `d.header`? If `Document` has `header` as subdocument

docarray/document/mixins/multimodal.py
```diff
@@ -139,3 +139,21 @@ def _from_obj(cls, obj, obj_type, field) -> typing.Tuple['Document', AttributeTy
...
+        has_metadata = has_data and getattr(self._data, '_metadata') is not None
+        if (
+            has_metadata
+            and self.is_multimodal
+            and attr in self._metadata['multi_modal_schema']
+        ):
+            mm_attr_da = self.get_multi_modal_attribute(attr)
+            return mm_attr_da if len(mm_attr_da) > 1 else mm_attr_da[0]
+        else:
+            raise AttributeError(f'{self.__class__.__name__} has no attribute {attr}')
```
samsja, line 159: I don't think we should refer to the class here but rather to the object. An Python object can have an attribute without it being define in the class ( we can add attribute on the fly after object instantiation)

samsja (commented):
small comment

docarray/document/mixins/multimodal.py
JohannesMessner: Are you referring to the error message? i don't think that the object itself has a `__name__`, could you write a code change suggestion showing what you would do?

JoanFM (approved):

hanxiao:
Please make sure to bump **minor** version before the release, this is a significant feature and worth a **minor** not **patch** release @JoanFM

JoanFM:
@hanxiao we know, we will work on the blog post and documentation before though

PR #441, opened 2022-07-14
fix: post should handle version properly

bwanglzu:
# Context

Not a big fan of the solution, but the current implementation of `post` is not usable:

1. versions are not handled. e.g. executor versions such as `v0.9.1-gpu` is commonly used while using an executor. The current implementation only allow us to use default, which is `latest`. This result in `host` parsing not correctly parse the executor versions.
2. keyword arguments are not parsed into `Executor`, such as `uses_with`, `volumes`. 

This could be better handled by unifying `jina.hubble.helper.parse_hub_uri`, or even better to create an extra `on` argument to the `post` method, but it will introduce breaking change, so i did not implement it.

It is already working. But i would do a bit more refactoring, move the logic of urlparsing out of `post`, to make it unit testable.

Why this is needed?

```python

da = DocumentArray([Document(text='hello')])
da = da.post(
    'jinahub+docker://FinetunerExecutor/v0.9.2/encode',
    uses_with={'artifact': '/mnt/finetune-quora-19.zip'},
    volumes=['/Users/bo/Documents/work2/test/my-model/:/mnt'],
)
assert da[0].embedding.shape==(768,)
```

### DIFF
--- a/docs/fundamentals/documentarray/post-external.md
+++ b/docs/fundamentals/documentarray/post-external.md
@@ -68,6 +68,7 @@ scheme://netloc[:port][/path]
 |           | 2. One of `jinahub`, `jinahub+docker`, `jinhub+sandbox` | Jina hub executor in source code, Docker container, sandbox                                                                          |
 | `netloc`  | 1. Host address                                         | `host` of the connected Flow                                                                                                         |
 |   | 2. Hub Executor name                                    | Any Executor [listed here](https://hub.jina.ai)                                                                                      |
+|   | 3. Executor version(optional)                           | Such as v0.1.1, v0.1.1-gpu, by default latest                                                                                        |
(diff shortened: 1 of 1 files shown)


JoanFM (changes requested):
Why not instead of pattern matching `version` just check if there are more `/` than expected, and u just extract the last part as the endpoint, and everything before as the host where u apply some extra logic?

bwanglzu:
@JoanFM do you think check extra `/` is even more fragile?

JoanFM:
> @JoanFM do you think check extra `/` is even more fragile?

it is much easier to mantain

JoanFM (changes requested):
LGTM, please make sure it is clear in documentation

JoanFM (approved):

Issue #429, opened 2022-07-06
bug: serializing and then deserializing to/from json or dict does not preserve dataclass properties


JohannesMessner:
**How to reproduce:**

```python
from docarray import Document, dataclass
from docarray.typing import Text


@dataclass
class MyMMDoc:
    t: Text
    

m = MyMMDoc(t='hi')
d = Document(m)
d.is_multimodal
Out[0]: True
d_proto = Document.from_protobuf(d.to_protobuf())
d_proto.is_multimodal
Out[1]: True
d_dict = Document.from_dict(d.to_dict())
d_dict.is_multimodal
Out[2]: False
d_json = Document.from_json(d.to_json())
d_json.is_multimodal
Out[2]: False
```

**Probable Cause:**

When calling `d.to_dict()` or `d.to_json()`, `d` gets converted through `to_pydantic_model()` to a `PydanticDocument`.
The `PydanticDocument` contains `_metadata` (which is responsible for enabling multimodal features) as a private attribute, and **private attributes are not kept by pydantic** (more info (here)[https://pydantic-docs.helpmanual.io/usage/models/#private-model-attributes]).

**Possible Solution:**

One idea to solve this coild be with a small special logic that converts `_metadata` to `metadata` when converting to `PydanticDocument`.

Something like this:
```python
    def to_pydantic_model(self) -> 'PydanticDocument':
        """Convert a Document object into a Pydantic model."""
        from ..pydantic_model import PydanticDocument as DP

        _p_dict = {}
        for f in self.non_empty_fields:
            v = getattr(self, f)
            ...
            elif f == '_metadata':
                _p_dict['metadata'] = v
            else:
                _p_dict[f] = v
        return DP(**_p_dict)
```

And the equivalent in `from_pydantic_model()`.

PR #443, opened 2022-07-18
fix: keep _metadata when serializing to dict and json

JohannesMessner:
This bug is the root cause of multi-modal documents not working after being (de)serilaized from/to json or dict.
Details here: https://github.com/jina-ai/docarray/issues/429

Better multi-modal support (https://github.com/jina-ai/docarray/pull/425) partially depends on this.

There seems to be no super elegant way of solving this with pydantic, based on [this](https://pydantic-docs.helpmanual.io/usage/models/#private-model-attributes) and [this](https://github.com/samuelcolvin/pydantic/issues/655).

I see two options:
1. make metadata public, so rename `_metadata` -> `metadata`
2. set and validate `_metadata` manually

In order not to change the pydantic model of `Document`, this PR implements option 2.

### DIFF
--- a/docarray/document/pydantic_model.py
+++ b/docarray/document/pydantic_model.py
@@ -65,9 +65,7 @@ class PydanticDocument(BaseModel):
         _metadata = data.get('_metadata', None)
         if _metadata is not None:
             _MetadataModel(metadata=_metadata)  # validate _metadata
-            object.__setattr__(
-                self, '_metadata', _metadata
-            )  # if no exception, assign _metadata
+            object.__setattr__(self, '_metadata', _metadata)
 
 
 PydanticDocument.update_forward_refs()
--- a/docarray/document/pydantic_model.py
+++ b/docarray/document/pydantic_model.py
@@ -1,6 +1,6 @@
 from typing import Optional, List, Dict, Any, TYPE_CHECKING, Union
 
-from pydantic import BaseModel, validator
+from pydantic import BaseModel, validator, PrivateAttr
 
 from ..math.ndarray import to_list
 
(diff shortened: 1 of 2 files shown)


tests/unit/document/test_multi_modal.py
```diff
@@ -629,3 +629,37 @@ class MMDoc:
...
+    @dataclass
+    class MMDocument:
+        text: Text
+
+    mm = MMDocument(
+        text='hello world',
+    )
+
+    doc = Document(mm)
+    assert doc._metadata
```
JoanFM, line 661: can u assert the value of this `metadata`?

JoanFM (changes requested):

docarray/document/pydantic_model.py
```diff
@@ -54,6 +59,16 @@ class PydanticDocument(BaseModel):
     class Config:
         smart_union = True
 
+    def __init__(self, **data):
+        super().__init__(**data)
+        # underscore attributes need to be set and validated manually
+        _metadata = data.get('_metadata', None)
+        if _metadata is not None:
+            _MetadataModel(metadata=_metadata)  # validate _metadata
+            object.__setattr__(
```
JoanFM, line 68: what is this `object`? Is `object` as the Python object base class? why not call `self`?

JohannesMessner: yes, it's the Python base object; calling on `self` does not work, because that would invoke `BaseModel`s implementation of `__setattr__`, which fails (I assume for the same reasons that we can't pass underscore attributes in the normal way in the first place).
The pydantic creator also recommends this way of doing it [here](https://github.com/samuelcolvin/pydantic/issues/655#issuecomment-570310120).

JoanFM (approved):

Issue #445, opened 2022-07-21
Implement Redis storage backend


AnneYang720:
Now Redis[ supports vector similarity search ](https://redis.io/docs/stack/search/reference/vectors/)algorithm based on HNSW.
 #452 

This issue will declare the overall steps on supporting a new vector storage backend: Redis into docarray.

This is part of Summer Of Open Source project work with Jina from students, together with @bwanglzu. To achieve the goal, we break down our tasks in to the following sub-tasks:

- [ ] implement backend.py
- [ ] implement getsetdel.py
- [ ] implement seqlike.py
- [ ] implement find.py
- [ ] documentation on the usage of RedisStorage in `integrations/document store` page.
- [ ] benchmarking and compare with other storage backends.
- [ ] document benchmarking results.

PR #452, opened 2022-07-25
feat(redis): implement Redis storage backend and unit tests

AnneYang720:
This PR will implement Redis storage backend and its unit test according to #445


docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,183 @@
...
+    host: Optional[str] = field(default='localhost')
+    port: Optional[int] = field(default=6379)
+    flush: Optional[bool] = field(default=False)
+    update_schema: Optional[bool] = field(default=True)
+    distance: Optional[str] = field(default='COSINE')
+    redis_config: Dict[str, Any] = field(default_factory=dict)
+    index_text: Optional[bool] = field(default=False)
+    tag_indices: List[str] = field(default_factory=list)
+    batch_size: Optional[int] = field(default=64)
+    method: Optional[str] = field(default='HNSW')
```
bwanglzu, line 38: why these fields are `Optional`? do they accept `None`?

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,183 @@
...
+    ):
+        if not config:
+            raise RedisConfig()
+        elif isinstance(config, dict):
+            config = dataclass_from_dict(RedisConfig, config)
+
+        if not config.distance in ['L2', 'IP', 'COSINE']:
+            raise ValueError(f'Distance metric {config.distance} not supported')
+        if not config.method in ['HNSW', 'FLAT']:
+            raise ValueError(f'Method {config.method} not supported')
```
bwanglzu, line 74: i guess a more pythonic way is to use `if something not in [your, list]`

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,183 @@
...
+
+        if self._config.flush:
+            client.flushdb()
+
+        # redis client will throw error if index does not exist
+        if self._config.update_schema:
+            try:
+                client.ft().dropindex('idx')
+            except ResponseError:
+                pass
```
bwanglzu, line 106: throw a proper error, not handle it sliently

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,183 @@
...
+
+            embedding = to_numpy_array(embedding)
+
+            if embedding.ndim > 1:
+                embedding = np.asarray(embedding).squeeze()
+        else:
+            embedding = np.zeros(self.n_dim)
+        return embedding.astype(np.float32).tobytes()
+
+    def _get_offset2ids_meta(self) -> List:
```
bwanglzu, line 164: the return type hints is a `List[something]`

bwanglzu (changes requested):
left some comments

JoanFM (changes requested):
I think these unit tests are excessive.You are only testing private methods. In general, we should test the public interface. These tests may make changing underlying implementation very hard

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,183 @@
...
+        """Return the offset2ids stored in redis
+
+        :return: a list containing ids
+
+        :raises ValueError: error is raised if index _client is not found or no offsets are found
+        """
+        if not self._client:
+            raise ValueError('Redis client does not exist')
```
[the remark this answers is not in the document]
JoanFM: Why do not u have `_client` as a property that creates _client if not exists? This way you do not need this check all over the place

JoanFM (changes requested):

bwanglzu:
please also fix the commit messages, see error in `commit-link`, use `git amend`

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,173 @@
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Tuple, Union
+
+import numpy as np
+
+from redis import Redis
+from redis.commands.search.field import NumericField, TextField, VectorField
+from redis.exceptions import ResponseError
+
+from .... import Document
```
JoanFM, line 10: please switch to absolute import as it has changed recently

tests/unit/array/storage/redis/test_backend.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+@pytest.mark.parametrize(
+    'id',
+    [
+        ('abc'),
+        ('123'),
+    ],
+)
+def test_doc_id_exists(id, da_redis, start_storage):
```
JoanFM, line 150: remove this test, as it is testing only private things we do not care

JoanFM (changes requested):
Of all these tests for now I would keep only `test_init_storage`

bwanglzu:
@AnneYang720 let's commit everything on this branch without creating seperate PRs, we'll make it ready-to-review until everything is here

JohannesMessner:
Hey @AnneYang720, just a heads up that we made some changes to Document Stores on master, in [this](https://github.com/jina-ai/docarray/pull/456) PR.
There are two small ways in which this will affect you:
- you need to change `extend()` (and `append()`, in case you also have it) to `_extend()` (and `_append()`) in `seqlike.py`.
- You need to implement `_ensure_unique_config()` in `backend.py`, more info [here](https://github.com/jina-ai/docarray/pull/456/files#diff-8871377f56f807d22dbe21ee1186e1c5bf9f3b744b286b0f42b365f0eb5e7274R210).

You can also do this at the end of your project, and we can collaborate on it if you need any help. Keep up the good work!

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,171 @@
...
+    """Provide necessary functions to enable this storage backend."""
+
+    TYPE_MAP = {
+        'str': TypeMap(type='text', converter=TextField),
+        'bytes': TypeMap(type='text', converter=TextField),
+        'int': TypeMap(type='integer', converter=NumericField),
+        'float': TypeMap(type='float', converter=NumericField),
+        'double': TypeMap(type='double', converter=NumericField),
+        'long': TypeMap(type='long', converter=NumericField),
+        # TODO add bool
```
bwanglzu, line 47: I forgot, why we leave a todo here?

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,171 @@
...
+        config: Optional[Union[RedisConfig, Dict]] = None,
+        **kwargs,
+    ):
+        if not config:
+            raise ValueError('Empty config is not allowed for Redis storage')
+        elif isinstance(config, dict):
+            config = dataclass_from_dict(RedisConfig, config)
+
+        if config.distance not in ['L2', 'IP', 'COSINE']:
+            raise ValueError(f'Distance metric {config.distance} not supported')
```
bwanglzu, line 62: something like this?
f'Expecting distance metric one of COSINE, L2 OR IP, got {config.distance} instead'

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,171 @@
...
+        if config.method not in ['HNSW', 'FLAT']:
+            raise ValueError(f'Method {config.method} not supported')
```
bwanglzu, line 64: same, the error message could be more clear

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,171 @@
...
+
+        if (
+            'decode_responses' in config.redis_config
+            and config.redis_config['decode_responses']
```
bwanglzu, line 68: i don't understand this if check, arent' they the same? why not:

```python
if config.redis_config.get('decode_responses')
```

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,171 @@
...
+            host=self._config.host,
+            port=self._config.port,
+            **self._config.redis_config,
+        )
+
+        if self._config.flush:
+            client.flushdb()
+
+        if self._config.update_schema:
+            if len(client.execute_command('FT._LIST')) > 0:
```
bwanglzu, line 98: what does FT._LIST means?

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,171 @@
...
+                client.ft().dropindex()
+
+        if self._config.flush or self._config.update_schema:
+            schema = self._build_schema_from_redis_config()
+            client.ft().create_index(schema)
```
bwanglzu, line 103: same question, what is `client.ft()`? can we create index directly from client with `client.create_index`? similarly, client.dropindex()

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,171 @@
...
+            port=self._config.port,
+            **self._config.redis_config,
+        )
+
+        if self._config.flush:
+            client.flushdb()
+
+        if self._config.update_schema:
+            if len(client.execute_command('FT._LIST')) > 0:
+                client.ft().dropindex()
```
bwanglzu, line 99: question, why `dropindex` here while in redis documentation i saw `drop_index`?

DA55FF4F-8826-4C84-980A-F7B6DAA2E576

bwanglzu (commented):
left some comments

docarray/array/storage/redis/find.py
```diff
@@ -0,0 +1,119 @@
...
+    def _find_similar_vectors(
+        self, query: 'RedisArrayType', filter: Optional[Dict] = None, limit=10
+    ):
+        q = (
+            Query("*=>[KNN " + str(limit) + " @embedding $vec AS vector_score]")
+            .sort_by('vector_score')
+            .dialect(2)
+        )
+
+        query_params = {"vec": to_numpy_array(query).astype(np.float32).tobytes()}
```
alaeddine-13, line 36:
```suggestion
        q = (
            Query('*=>[KNN $limit @embedding $vec AS vector_score]')
            .sort_by('vector_score')
            .dialect(2)
        )

        query_params = {'vec': to_numpy_array(query).astype(np.float32).tobytes(), 'limit': str(limit)}
```

docarray/array/storage/redis/find.py
```diff
@@ -0,0 +1,119 @@
...
+        if n_dim != 2:
+            query = query.reshape((num_rows, -1))
+
+        return [
+            self._find_similar_vectors(q, filter=filter, limit=limit) for q in query
+        ]
+
+    def _find_with_filter(self, filter: Dict, limit: Optional[Union[int, float]] = 20):
+
+        if filter:
```
alaeddine-13, line 68: why do we need this check ?

docarray/array/storage/redis/find.py
```diff
@@ -0,0 +1,119 @@
...
+        elif filter['operator'] == 'gte':
+            f = NumericFilter(filter['key'], filter['value'], INF)
+        elif filter['operator'] == 'lt':
+            f = NumericFilter(
+                filter['key'], NEG_INF, filter['value'], maxExclusive=True
+            )
+        elif filter['operator'] == 'lte':
+            f = NumericFilter(filter['key'], NEG_INF, filter['value'])
+
+        return f
```
alaeddine-13, line 104: Since redis does not provide a query language, let's try to apply mongoDB-like query language, which is used by default in docarray: https://docarray.jina.ai/fundamentals/documentarray/find/

docarray/array/storage/redis/find.py
```diff
@@ -0,0 +1,119 @@
...
+        if filter['operator'] == 'gt':
+            s = "@{}:[({} {}]".format(filter['key'], filter['value'], INF)
+        elif filter['operator'] == 'gte':
+            s = "@{}:[{} {}]".format(filter['key'], filter['value'], INF)
+        elif filter['operator'] == 'lt':
+            s = "@{}:[{} ({}]".format(filter['key'], NEG_INF, filter['value'])
+        elif filter['operator'] == 'lte':
+            s = "@{}:[{} {}]".format(filter['key'], NEG_INF, filter['value'])
+
+        return s
```
alaeddine-13, line 119: Same as above: filter objects input by users ideally should follow syntax here instead of inventing a new one: https://docarray.jina.ai/fundamentals/documentarray/find/

docarray/array/storage/redis/seqlike.py
```diff
@@ -0,0 +1,80 @@
...
+        """Return the length of :class:`DocumentArray` that uses Redis as storage
+
+        :return: the length of this :class:`DocumentArrayRedis` object
+        """
+        try:
+            keys = self._client.keys()
+            if b'offset2id' in keys:
+                return len(keys) - 1
+            else:
+                return len(keys)
```
alaeddine-13, line 34: This will return all keys in the database (image we have millions of those)
let's instead use `dbsize()` and check `b'offset2id' in keys` differently (for example client.exists or so)

docarray/array/storage/redis/seqlike.py
```diff
@@ -0,0 +1,80 @@
...
+    #     self._offset2ids.clear()
+
+    def __repr__(self):
+        """Return the string representation of :class:`DocumentArrayRedis` object
+        :return: string representation of this object
+        """
+        return f'<DocumentArray[Redis] (length={len(self)}) at {id(self)}>'
+
+    def _upload_batch(self, docs: Iterable['Document']):
+        print('uploading batch @', __import__('time').time_ns())
```
alaeddine-13, line 64: let's avoid this print

tests/unit/array/mixins/test_text.py
```diff
@@ -140,12 +145,13 @@ def test_da_text_to_tensor_max_len_1(docs, da_cls, config, start_storage):
...
-        (DocumentArrayWeaviate, WeaviateConfig(n_dim=128)),
-        (DocumentArrayQdrant, QdrantConfig(n_dim=128)),
-        (DocumentArrayElastic, ElasticConfig(n_dim=128)),
+        # (DocumentArray, None),
+        # (DocumentArraySqlite, None),
+        # (DocumentArrayAnnlite, AnnliteConfig(n_dim=128)),
+        # (DocumentArrayWeaviate, WeaviateConfig(n_dim=128)),
+        # (DocumentArrayQdrant, QdrantConfig(n_dim=128)),
+        # (DocumentArrayElastic, ElasticConfig(n_dim=128)),
+        (DocumentArrayRedis, RedisConfig(n_dim=128, flush=True)),
```
alaeddine-13, line 154:
```suggestion
        (DocumentArray, None),
        (DocumentArraySqlite, None),
        (DocumentArrayAnnlite, AnnliteConfig(n_dim=128)),
        (DocumentArrayWeaviate, WeaviateConfig(n_dim=128)),
        (DocumentArrayQdrant, QdrantConfig(n_dim=128)),
        (DocumentArrayElastic, ElasticConfig(n_dim=128)),
        (DocumentArrayRedis, RedisConfig(n_dim=128, flush=True)),
```

docarray/array/storage/redis/backend.py
AnneYang720: I wasn't sure whether to add boolean type to TYPE_MAP.

docarray/array/storage/redis/backend.py
AnneYang720: FT._LIST returns a list of all existing indexes and this command is not wrapped in redis-py. Before dropping current index, we need to check if there is existing index, otherwise Redis will throw exception.

docarray/array/storage/redis/backend.py
AnneYang720: This example is using redisearch.client package. Here I'm using redis-py. `client.ft() ` accesses the search namespace, providing support for redis search. 
https://redis.readthedocs.io/en/stable/examples/search_vector_similarity_examples.html

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,180 @@
...
+    host: str = field(default='localhost')
+    port: int = field(default=6379)
+    flush: bool = field(default=False)
+    update_schema: bool = field(default=True)
+    distance: str = field(default='COSINE')
+    redis_config: Dict[str, Any] = field(default_factory=dict)
+    index_text: bool = field(default=False)
+    tag_indices: List[str] = field(default_factory=list)
+    batch_size: int = field(default=64)
+    method: str = field(default='HNSW')
```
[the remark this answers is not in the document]
AnneYang720: FLAT is a brute-force searching method and can provide accurate results. I think some users may want to use it and supporting FLAT doesn't make the code more complex.

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,180 @@
...
+        return client
+
+    def _ensure_unique_config(
+        self,
+        config_root: dict,
+        config_subindex: dict,
+        config_joined: dict,
+        subindex_name: str,
+    ) -> dict:
+        return config_joined
```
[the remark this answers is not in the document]
AnneYang720: We planned to add sub_index for redis in next PR.

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,180 @@
...
+            index_options = {
+                'M': self._config.m or 16,
+                'EF_CONSTRUCTION': self._config.ef_construction or 200,
+                'EF_RUNTIME': self._config.ef_runtime or 10,
+            }
+            index_param.update(index_options)
+
+        if self._config.method == 'FLAT' and self._config.block_size:
+            index_options = {'BLOCK_SIZE': self._config.block_size}
+            index_param.update(index_options)
```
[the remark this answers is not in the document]
AnneYang720: That's true. I'll discuss whether to support FLAT with @bwanglzu

docarray/array/storage/redis/seqlike.py
```diff
@@ -0,0 +1,91 @@
...
+            # method 2
+            # this way, extend(), insert() funcs have to call self._save_offset2ids()
+            # if self._client.exists(self._offset2id_key.encode()):
+            #     print('offset2id exists')
+            #     return self._client.llen(self._offset2id_key.encode())
+            # else:
+            #     return 0
+
+            # method 3
+            return len(self._offset2ids)
```
AnneYang720, line 47: Here I have three possible ways to implement the function __len__ and I cannot decide which one is better.
@alaeddine-13 @JohannesMessner

docarray/array/storage/redis/getsetdel.py
```diff
@@ -0,0 +1,74 @@
...
+        """Concrete implementation of base class' ``_get_doc_by_id``
+
+        :param _id: the id of the document
+        :return: the retrieved document from redis
+        """
+        try:
+            result = self._client.hgetall(self._doc_prefix + _id)
+            doc = Document.from_base64(result[b'blob'])
+            return doc
+        except Exception as ex:
```
bwanglzu, line 22: what exception do we get from Redis, can we make it more specific?

docarray/array/storage/redis/seqlike.py
bwanglzu, line 47: any thoughts around these different ways of implementing this function? @alaeddine-13

docarray/array/storage/redis/seqlike.py
```diff
@@ -0,0 +1,91 @@
...
+            return self._doc_id_exists(x)
+        elif isinstance(x, Document):
+            return self._doc_id_exists(x.id)
+        else:
+            return False
+
+    # TODO this del is unreachable, del will call __del__ in base/getsetdel
+    # def __del__(self):
+    #     """Delete this :class:`DocumentArrayRedis` object"""
+    #     self._offset2ids.clear()
```
bwanglzu, line 67: what is it?

docarray/array/storage/redis/seqlike.py
```diff
@@ -0,0 +1,91 @@
...
+        batch = 0
+        for doc in docs:
+            payload = self._document_to_redis(doc)
+            pipe.hset(self._doc_prefix + doc.id, mapping=payload)
+            batch += 1
+            if batch >= self._config.batch_size:
+                pipe.execute()
+                batch = 0
+        if batch > 0:
+            pipe.execute()
```
bwanglzu, line 86: now you mainted a `batch` of counter. why not leverage docarray [batch function](https://docarray.jina.ai/api/docarray.array.document/?highlight=batch#docarray.array.document.DocumentArray.batch) to do the batch inserting:

```python
for batch_of_docs in docs.batch(batch_size):
    # .. do something
    pipe.execute()
```

bwanglzu: note in this case, if `docs` is not a `DocumentArray`, first convert the `Iterable` into a `DocumentArray`.

docarray/array/storage/redis/seqlike.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _extend(self, docs: Iterable['Document']):
+        docs = list(docs)
+        self._upload_batch(docs)
+        self._offset2ids.extend([doc.id for doc in docs])
```
bwanglzu, line 91: if you can take the `batch` above, then you can move `offset2ids` into `_upload_batch` and use `self._offset2ids.extend(batch_of_docs.batch_ids)`

bwanglzu (commented):
lefrt some new comments, please take a look

docarray/array/storage/redis/seqlike.py
alaeddine-13: @AnneYang720 not possible to do something like `self._client.keys() - 1` (or minus whatever is the number of non-document array keys) ?
If possible, I'd go with it since it reflects most up-to-date number, otherwise, len(self._offset2ids)

AnneYang720: I didn't find any clean commands that can return only the number of all matching keys.
There is one possible solution, using sever-side lua scipt, from [here](https://stackoverflow.com/a/26863445)

bwanglzu:
last thing: decide how we create index for root and chunks

docarray/array/storage/redis/seqlike.py
bwanglzu: let's keep it like this and create a follow up ticket to research more.

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        if self._config.method == 'HNSW' and (
+            self._config.m or self._config.ef_construction or self._config.ef_runtime
+        ):
+            index_options = {
+                'M': self._config.m or 16,
+                'EF_CONSTRUCTION': self._config.ef_construction or 200,
+                'EF_RUNTIME': self._config.ef_runtime or 10,
+            }
```
bwanglzu, line 142: why not set 16, 200 and 10 as default values in `_init_`?

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,192 @@
...
+            index_options = {'BLOCK_SIZE': self._config.block_size}
+            index_param.update(index_options)
+
+        if self._config.initial_cap:
+            index_param['INITIAL_CAP'] = self._config.initial_cap
+        schema = [VectorField('embedding', self._config.method, index_param)]
+
+        if self._config.tag_indices:
+            for index in self._config.tag_indices:
+                # TODO TextField or TagField
```
bwanglzu, line 155: what is this TODO?

bwanglzu (commented):
some minor comments

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,192 @@
...
+                schema.append(TextField(index))
+
+        # TODO whether to add schema to column (elastic does but qdrant doesn't)
```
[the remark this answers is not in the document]
AnneYang720: That was something I wasn't sure at the beginng. 
We need to add schema to column.

bwanglzu (approved):
LGTM!

bwanglzu:
let's create documentation after this one

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,188 @@
...
+                f'Expecting search method one of HNSW OR FLAT, got {config.method} instead'
+            )
+
+        if config.redis_config.get('decode_responses'):
+            config.redis_config['decode_responses'] = False
+
+        self._offset2id_key = 'offset2id__' + config.index_name
+        self._config = config
+        self.n_dim = self._config.n_dim
+        self._doc_prefix = "doc__" + config.index_name + ":"
```
JohannesMessner, line 78: will we remove this `"doc__"` as discussed in the call? Also, please use `'` instead of `"`,  just to maintain a consistent style.

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,188 @@
...
+        config_root: dict,
+        config_subindex: dict,
+        config_joined: dict,
+        subindex_name: str,
+    ) -> dict:
+        if 'index_name' not in config_subindex:
+            config_joined['index_name'] = (
+                config_joined['index_name'] + '_subindex_' + subindex_name
+            )
+        config_joined['flush'] = False
```
JohannesMessner, line 125: What does this do? Why do we need it here?

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,188 @@
...
+        return config_joined
+
+    def _build_schema_from_redis_config(self):
+        index_param = {
+            'TYPE': 'FLOAT32',
```
JohannesMessner, line 130: Should this be configurable?

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,188 @@
...
+        if self._config.index_text:
+            schema.append(TextField('text'))
+        return schema
+
+    def _doc_id_exists(self, doc_id):
+        return self._client.exists(self._doc_prefix + doc_id)
+
+    def _map_embedding(self, embedding: 'ArrayType') -> bytes:
+        if embedding is not None:
+            from ....math.ndarray import to_numpy_array
```
JohannesMessner, line 167: I think we try to use abolsute import instead of `...`, so like `from docarray.whatever.math.ndarray import to_numpy_array`, just for readability

docarray/array/storage/redis/getsetdel.py
JohannesMessner: I think since we raise it again it is fine like this

tests/unit/array/storage/redis/test_backend.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+type_convert = {
+    'int': b'NUMERIC',
+    'float': b'NUMERIC',
+    'double': b'NUMERIC',
+    'long': b'NUMERIC',
+    'str': b'TEXT',
+    'bytes': b'TEXT',
+}
```
JohannesMessner, line 33: Since we also have `bool` now, should it be here as well? Additional test case needed?

JohannesMessner (changes requested):
I can't comment on the Redis specific things, but looks like we are getting very close! Nice!
I left some small things to consider.

docarray/array/storage/redis/backend.py
AnneYang720: The parameter flush only exists in RedisConfig, flush=True will clear the redis database. We don't want to clear database when there is subindex docarray, so I make sure flush is set to False here.

docarray/array/storage/redis/backend.py
AnneYang720: TYPE is a mandantory parameter. Current supported type is only FLOAT32. see [here](https://redis.io/docs/stack/search/reference/vectors/)

hanxiao (changes requested):
LGTM, however docs is not in this PR, so I will wait until it is clear and straightforward to me how the usage, caveat is before merging. Please refer other store backend docs yet don't copy from them. Highlight the basics, features, the differences. 

Rule of thumb is, that when you write a doc and you start to copy-paste sentences, then those sentences serve no use besides a placeholder

tests/unit/array/mixins/test_getset.py
```diff
@@ -493,9 +515,13 @@ def test_getset_subindex(storage, config):
         ('qdrant', {'n_dim': 3, 'distance': 'euclidean'}),
         ('elasticsearch', {'n_dim': 3, 'distance': 'l2_norm'}),
         ('sqlite', dict()),
+        ('redis', {'n_dim': 3, 'distance': 'L2', 'flush': True}),
     ],
 )
 def test_init_subindex(storage, config):
+    if storage == 'redis':
+        ensure_gc
```
alaeddine-13, line 523: I don't get how this works
why don't we make it a function instead of fixture and just call it ?

tests/unit/array/storage/redis/test_backend.py
AnneYang720: I add test case test_redis_category_filter in /docarray/tests/unit/array/mixins/test_find.py

tests/unit/array/test_advance_indexing.py
```diff
@@ -658,6 +683,9 @@ def test_edge_case_two_strings(storage, config_gen, start_storage):
     with pytest.raises(IndexError):
         da['1', 'hellohello'] = 'hello'
 
+    if storage == 'redis':
+        ensure_gc
```
alaeddine-13, line 687: same as my comment in tests/unit/array/mixins/test_getset.py

tests/unit/array/test_sequence.py
```diff
@@ -81,12 +89,16 @@ def update_config_inplace(config, tmpdir, tmpfile):
...
         ('elasticsearch', {'n_dim': 3, 'index_name': 'elasticsearch'}),
+        ('redis', {'n_dim': 3, 'flush': True}),
     ],
 )
 def test_context_manager_from_disk(storage, config, start_storage, tmpdir, tmpfile):
     config = config
     update_config_inplace(config, tmpdir, tmpfile)
 
+    if storage == 'redis':
+        ensure_gc
```
alaeddine-13, line 100: same as in tests/unit/array/mixins/test_getset.py

docarray/array/storage/redis/backend.py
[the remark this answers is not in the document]
AnneYang720: We decide to remove part '__doc'.

docarray/array/storage/redis/backend.py
```diff
@@ -0,0 +1,188 @@
...
+
+        if self._config.method == 'HNSW':
+            index_options = {
+                'M': self._config.m,
+                'EF_CONSTRUCTION': self._config.ef_construction,
+                'EF_RUNTIME': self._config.ef_runtime,
+            }
+            index_param.update(index_options)
+
+        if self._config.method == 'FLAT' and self._config.block_size:
```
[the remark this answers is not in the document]
AnneYang720: block_size is optional and has default value. so i changed it to the similar as HNSW

docarray/array/storage/redis/backend.py
JohannesMessner: I am not sure about this, why would a subindex behave differently from the main index? Shouldn't wee let the user decide what they want to do? @alaeddine-13 @bwanglzu ?

bwanglzu: didn't get it, i think we already have a[ flush option ](https://github.com/jina-ai/docarray/pull/452/files#diff-4cf8967d09d437c784b4b867ed657bdbbb4be5e9df90dad272717d063b0a64b9R23) which user can control in the config?

docarray/array/storage/redis/getsetdel.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _load_offset2ids(self):
+        ids = self._get_offset2ids_meta()
+        self._offset2ids = Offset2ID(ids)
+
+    def _save_offset2ids(self):
+        self._update_offset2ids_meta()
+
+    def _clear_storage(self):
+        self._client.flushdb()
```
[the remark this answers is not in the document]
AnneYang720: Currently we want it to be 1 docarray per redis. But with current prefix design added, it seems that we cannot stop users from storing multiple docarrays.

alaeddine-13 (approved):
Besides the documentation, LGTM, well done !

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,334 @@
...
+```
+
+
+
+Other functions behave the same as in-memory DocumentArray.
+
+
+### Vector search with filter query
+
+One can perform Vector Similarity Search based on FLAT or HNSW algorithm and pre-filter results using a filter query that is based on [MongoDB's Query](https://www.mongodb.com/docs/manual/reference/operator/query/). We currently support a subset of those selectors:
```
bwanglzu, line 122: i wouldn't mention MongoDB here. BTW we can refer it as `domain specific language` in the future or [DSL](https://en.wikipedia.org/wiki/Domain-specific_language)

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,334 @@
...
+| `block_size`      | Optional parameter for Redis FLAT algorithm                                                       | `1048576`                                         |
+| `initial_cap`     | Optional parameter for Redis HNSW and FLAT algorithm                                              | `None`, defaults to the default value in Redis    |
+| `columns`         | Other fields to store in Document and build schema                                                | `None`                                            |
+
+You can check the default values in [the docarray source code](https://github.com/jina-ai/docarray/blob/main/docarray/array/storage/redis/backend.py)
+
+
+```{note}
+This Document Store will support storing multiple DocumentArrays and full-text search soon.
+```
```
bwanglzu, line 334: don't forget the geo filters and additional features are not supported yet

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,334 @@
...
+| `n_dim`           | Dimensionality of the embeddings                                                                  | `None`                                            |
+| `flush`           | Boolean flag indicating whether to clear previous DocumentArray in Redis                          | `False`                                           |
+| `update_schema`   | Boolean flag indicating whether to update Redis Search schema                                     | `True`                                            |
+| `distance`        | Similarity distance metric in Redis                                                               | `'COSINE'`                                        |
+| `batch_size`      | Batch size used to handle storage updates                                                         | `64`                                              |
+| `method`          | Vector similarity index algorithm in Redis                                                        | `'HNSW'`                                          |
```
bwanglzu, line 321: can we say `Vector similarity index algorithm in Redis` -> `Vector similarity index algorithm in Redis, either `FLAT` or `HNSW`  `

bwanglzu (commented):
great work on the docs! left some minor comments

docs/advanced/document-store/redis.md
hanxiao: wait, `This Document Store will support storing multiple DocumentArrays` is against the design principle of a DocumentArray. Note that `da = DocumentArray(storage='redis')` is after all **one** DocumentArray. By concept-substitution, you see the logic problem here: it doesn't make sense to say "This DocumentArray will support storing multiple DocumentArrays"

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,339 @@
...
+
+### Start Redis service
+
+To use Redis as the storage backend, it is required to have the Redis service started. Create `docker-compose.yml` as follows:
+
+```yaml
+version: "3.3"
+services:
+  redis:
+    image: redislabs/redisearch:2.6.0
```
alaeddine-13, line 20: why it's different from the other docker image in the tests ?

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,339 @@
...
+```python
+from docarray import DocumentArray
+
+da = DocumentArray(storage='redis', config={'n_dim': 128})
+```
+
+The usage would be the same as the ordinary DocumentArray, but the dimension of an embedding for a Document must be provided at creation time.
+
+```{caution}
+Currently, one Redis server instance can only support to store a single DocumentArray.
```
alaeddine-13, line 44:
```suggestion
Currently, one Redis server instance can only store a single DocumentArray.
```

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,339 @@
...
+
+To store a new DocumentArray in current Redis server, one can set `flush` to `True` so that previous DocumentArray will be cleared:
+
+```python
+from docarray import DocumentArray
+
+da = DocumentArray(storage='redis', config={'n_dim': 128, 'flush': True})
+```
+
+To access a previously stored DocumentArray, one can specify `host` and `port`.
```
alaeddine-13, line 55: I think this is not accurate, host a port are mostly about the target instance. You can still not specify host and port and have access to a previously stored document array. Maybe you're referring to flush = False ?

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,339 @@
...
+This would print:
+
+```console
+Embeddings Approximate Nearest Neighbours with "price" at most 7 and "color" red:
+
+ embedding=[3. 3. 3.],   price=3,        color=red,      score=0
+ embedding=[6. 6. 6.],   price=6,        color=red,      score=0
+ embedding=[1. 1. 1.],   price=1,        color=red,      score=5.96046447754e-08
+ embedding=[2. 2. 2.],   price=2,        color=red,      score=5.96046447754e-08
+ embedding=[4. 4. 4.],   price=4,        color=red,      score=5.96046447754e-08
```
alaeddine-13, line 214: something is wrong, shouldn't we find those closes to (8, 8, 8) ? (or maybe we're not putting the right values as example)

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,339 @@
...
+This would print:
+
+```console
+Embeddings Approximate Nearest Neighbours:
+
+ embedding=[3. 3. 3.],   score=0
+ embedding=[6. 6. 6.],   score=0
+ embedding=[1. 1. 1.],   score=5.96046447754e-08
+ embedding=[2. 2. 2.],   score=5.96046447754e-08
+ embedding=[4. 4. 4.],   score=5.96046447754e-08
```
alaeddine-13, line 273: same I feel something is wrong with this example

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+Then you can use a different search indexing schema on current DocumentArray as follows:
```
JohannesMessner, line 276:
```suggestion
Then you can use a different search indexing schema on the current DocumentArray as follows:
```

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,339 @@
...
+
+print('\nEmbeddings Approximate Nearest Neighbours:\n')
+for embedding, score in zip(
+    results.embeddings,
+    results[:, 'scores'],
+):
+    print(f' embedding={embedding},\t score={score["score"].value}')
+```
+
+This would print:
```
JohannesMessner, line 297:
```suggestion
This will print:
```

docs/advanced/document-store/redis.md
```diff
@@ -0,0 +1,339 @@
...
+
+ embedding=[8. 8. 8.],   score=0
+ embedding=[9. 9. 9.],   score=3
+ embedding=[7. 7. 7.],   score=3
+ embedding=[6. 6. 6.],   score=12
+ embedding=[5. 5. 5.],   score=27
+```
+
+
+## Config
```
JohannesMessner, line 310:
```suggestion
## Configuration
```

JohannesMessner (changes requested):
some rephrasings

docs/advanced/document-store/redis.md
AnneYang720: redis-stack includes something we won't need here, like graph data models
and the size redissearch docker image is half of redis-stack

PR #456, opened 2022-07-28
feat: subindex for all document stores

JohannesMessner:
This PR is a draft for storing secondary indices as a dict with different docarrays

**ToDo**:

- [x] .find()
- [x] .match()
  - [x] tests
- [x] add
- [x] delete
- [x] update
- [x] bug: multi-modal access path. Fix here: #448
- [x] tests for multimodal (added in find and del)
- [x] Decide on syntax and argument names
  - [x] Implement new nomenclature
- [ ] ~~Implement InMemory as syntax sugar~~
- [x] Validate/test other backends
  - [x] ES
  - [x] Qdrant
  - [x] Weaviate
  - [x] SQLite
- [x] set subindices correctly when chunks* are given at init
- [x] return subindex when corresponding access path is passed
- [x] Documentation
  - [x] for end user
  - [x] for backend devs  

**Example usage**:

This feature lets you store a separate index for a given nesting level, and then search through it.

This feature works with access paths like `'@c'` and `'@cc'`, but also with multi-modal access paths like `'@.[image]`. Example:

```python
from docarray import dataclass, Document, DocumentArray
from docarray.typing import Image, Text


@dataclass
class Page:
    main_text: Text
    image: Image
    description: Text


query_page = Page(
    main_text='Hello world',
    image='testflow.jpg',
    description='This is the image of an apple',
)

query = Document(query_page)  # our query Document

da = DocumentArray(
    [
        Document(
            Page(
                main_text='First page',
                image='apple.png',
                description='This is the image of an apple',
            )
        ),
        Document(
            Page(
                main_text='Second page',
                image='testflow.jpg',
                description='This is an image of a pear',
            )
        ),
    ],
    subindex_configs = {'@.[image]': None})  # specify subindices at docarray creation time


from torchvision.models import resnet50

img_model = resnet50(pretrained=True)

# embed query
query.image.set_image_tensor_shape(shape=(224, 224)).set_image_tensor_channel_axis(
    original_channel_axis=-1, new_channel_axis=0
).set_image_tensor_normalization(channel_axis=0).embed(img_model)

# embed dataset
da['@.[image]'].apply(
    lambda d: d.set_image_tensor_shape(shape=(224, 224))
    .set_image_tensor_channel_axis(original_channel_axis=-1, new_channel_axis=0)
    .set_image_tensor_normalization(channel_axis=0)
).embed(img_model)


closest_match_img = da.find(query.image, on='@.[image]')[0][0]  # search through subindex using `on=`
print('CLOSEST IMAGE:')
closest_match_img.summary()
print('PAGE WITH THE CLOSEST IMAGE:')
closest_match_page = da[closest_match_img.parent_id]
closest_match_page.summary()
```

**Design doc:**

Design doc: https://docs.google.com/presentation/d/1rntTa1Ur2WmdAvOUEI01l2WeNSWgtqk2xjth_gdIr50/edit#slide=id.g13443d1ddb2_1_23

Continuation of this: https://github.com/jina-ai/docarray/pull/403


docarray/array/storage/annlite/getsetdel.py
```diff
@@ -19,15 +19,29 @@ def _get_doc_by_id(self, _id: str) -> 'Document':
         return doc
 
     def _set_doc_by_id(self, _id: str, value: 'Document'):
+
+        if self._secondary_indices:
+            for selector, da in self._secondary_indices.items():
+                old_ids = DocumentArray(self[_id])[
+                    selector, 'id'
+                ]  # hack to get the Document['@c'] without having to do Document.chunks
+                print(da.embeddings)
```
JoanFM, line 28: remember to remove prints

docarray/array/storage/annlite/getsetdel.py
```diff
@@ -19,15 +19,29 @@ def _get_doc_by_id(self, _id: str) -> 'Document':
...
+                with da:
+                    del da[old_ids]
+                    da.extend(DocumentArray(value)[selector])  # same hack here
+                print(da.embeddings)
```
JoanFM, line 32: same here

JoanFM (changes requested):

docarray/array/mixins/match.py
```diff
@@ -70,20 +71,34 @@ def match(
...
+                exclude_self=exclude_self,
+                filter=filter,
+                only_id=only_id,
+                use_scipy=use_scipy,
+                device=device,
+                num_worker=num_worker,
+                **find_kwargs,
+            )
+        except TypeError as e:
+            if 'on' in str(e):
```
JoanFM, line 96: I do not like this logic, the ValueError should already be sent from inside.

Also `on` is too common of a word

JoanFM (changes requested):
I think the way to handle other backends not supporting `on` should be improved

docarray/array/mixins/match.py
JohannesMessner: I agree that this is not great. But if ValueError should be sent from the inside we would have to implement that for every backend. Alternatively we can just remove this check and it will also fail if, but in a less informative way.

docarray/array/mixins/find.py
```diff
@@ -125,11 +126,15 @@ def find(
...
                       otherwise the tag field specified by `index` will be used. You can only use this parameter if the
                       storage backend supports searching by text.
+        :param on: specifies a subindex to search on. If set, then the returned DocumentArray will be retrieved from the given subindex. Needs to be supported by a given Document Store.
         :param kwargs: other kwargs.
 
         :return: a list of DocumentArrays containing the closest Document objects for each of the queries in `query`.
         """
 
+        if on is not None:
+            kwargs.update({'on': on})
```
JoanFM, line 136: I prefer to pass `on=on` wherever is needed, or just avoid having `on` as a keyword parameter here

JoanFM (changes requested):

docarray/array/storage/weaviate/backend.py
```diff
@@ -115,6 +115,30 @@ def _init_storage(
             if isinstance(_docs, Document):
                 self.append(_docs)
 
+    def _init_subindices(self, *args, **kwargs):
+        from docarray import DocumentArray
+
+        self._subindices = {}
+        subindex_configs = kwargs.get('subindex_configs', None)
```
JoanFM, line 122: why do not u pass it as an argument? so that u can pass proper `docstrings` and all of that?

JoanFM (changes requested):

docarray/array/storage/weaviate/backend.py
JohannesMessner: It is because when such a da gets created through the syntax sugar `DocumentArray(storage='weaviate', ...)`, all the *args and **kwargs get passed to the specialized backend class. This is the code snippet:
```python
class BaseDocumentArray(MutableSequence[Document]):
    def __init__(self, *args, storage: str = 'memory', **kwargs):
        super().__init__()
        self._init_storage(*args, **kwargs)
        self._init_subindices(*args, **kwargs)
```
So when `_init_subindices` takes `subindex_configs` as a normal argument, it will mistake the first argument that the user passes as `subindex_configs`. Usually that is the list of Documents that make up the DocumentArray, and `_init_subindices` will think that this list of Documents is `subindex_configs`.

What I am doing here is enforcing `subindex_configs` to be passes with the keyword, not as a positional argument, so that no confusion can happen between args.

JohannesMessner:
This is currently blocked by a bug in the SQLite backend: https://github.com/jina-ai/docarray/issues/470

docarray/array/mixins/match.py
```diff
@@ -70,6 +71,11 @@ def match(
         for d in self:
             d.matches.clear()
 
+        if on is not None:
+            find_kwargs = {'on': on}
```
JoanFM, line 75: why {'on': on} will not directly work regardless of it being None?

docarray/array/mixins/match.py
```diff
@@ -60,7 +61,7 @@ def match(
 
                 .. note::
                     This argument is only effective when ``batch_size`` is set.
-
+        :param on: specifies a subindex to search on. If set, then the returned DocumentArray will be retrieved from the given subindex. Needs to be supported by a given Document Store.
```
JoanFM, line 64: It is supported by default right, last comment is not needed anymore?

docarray/array/storage/annlite/backend.py
```diff
@@ -60,35 +60,50 @@ def _normalize_columns(self, columns):
...
         return columns
 
+    def _ensure_subindex_is_unique(
+        self,
+        config_root: dict,
+        config_subindex: dict,
+        config_joined: dict,
+        subindex_name: str,
+    ) -> dict:
+        import os
```
JoanFM, line 70: _ensure_unique_config? or something like this?

docarray/array/storage/base/getsetdel.py
```diff
@@ -115,6 +115,7 @@ def _del_docs_by_mask(self, mask: Sequence[bool]):
 
     def _del_all_docs(self):
         self._clear_storage()
+        self._clear_subindices()
```
JoanFM, line 118: it seems to me that `subindices` should be cleared before?

JoanFM (changes requested):

docarray/array/mixins/match.py
JohannesMessner: This is a relic from when not all backends had an `on` parameter. Will remove.

docarray/array/storage/base/getsetdel.py
JohannesMessner: I don't think it matters. Why would it make a difference?

JoanFM: semantically speaking

docarray/array/storage/elastic/find.py
```diff
@@ -127,6 +127,10 @@
         :return: DocumentArray containing the closest documents to the query if it is a single query, otherwise a list of DocumentArrays containing
            the closest Document objects for each of the queries in `query`.
         """
+        if 'on' in kwargs.keys():
+            raise ValueError(
+                f'subindices (`on=...`) are not available for this Document Store ({self.__class__.__name__}).'
+            )
```
samsja, line 133: Why is it not available here ?

docarray/array/storage/memory/find.py
```diff
@@ -53,6 +53,11 @@ def _find(
 
         :return: a list of DocumentArrays containing the closest Document objects for each of the queries in `query`.
         """
+        if 'on' in kwargs.keys():
+            raise ValueError(
+                f'subindices (`on=...`) are not available for this Document Store ({self.__class__.__name__}).'
+            )
```
samsja, line 59: Same, Why is it not available here ? I thought we said it could still be usefull for in memory

docarray/array/storage/qdrant/find.py
```diff
@@ -93,6 +93,11 @@ def _find(
         :return: a list of DocumentArrays containing the closest Document objects for each of the queries in `query`.
         """
 
+        if 'on' in kwargs.keys():
+            raise ValueError(
+                f'subindices (`on=...`) are not available for this Document Store ({self.__class__.__name__}).'
+            )
+
```
samsja, line 100: Same, Why is it not available here ?

docarray/array/storage/weaviate/find.py
```diff
@@ -191,6 +191,11 @@ def _find(
                   https://weaviate.io/developers/weaviate/current/more-resources/faq.html#q-how-do-i-get-the-cosine-similarity-from-weaviates-certainty
         """
 
+        if 'on' in kwargs.keys():
+            raise ValueError(
+                f'subindices (`on=...`) are not available for this Document Store ({self.__class__.__name__}).'
+            )
```
samsja, line 197: Same , Why is it not available here ?

samsja (commented):
Really great job ! The abstraction is super nice it is almost seamless to support subindex for a new backend.

I added some question regarding the raise of not available on the find method

docarray/array/storage/weaviate/backend.py
```diff
@@ -115,6 +115,28 @@ def _init_storage(
             if isinstance(_docs, Document):
                 self.append(_docs)
 
+    def _sanitize_class_name(self, name):
+        banned_chars = [
```
JoanFM, line 119: put this at the top of rhe file. No need to instantiate every call

docarray/array/storage/elastic/backend.py
```diff
@@ -99,6 +100,13 @@ def _init_storage(
             if isinstance(_docs, Document):
                 self.append(_docs)
 
+    def _sanitize_index_name(self, name):
+        banned_chars = ['[', ' ', '"', '*', '\\', '<', '|', ',', '>', '/', '?', ']']
```
JoanFM, line 104: same here

docarray/array/storage/elastic/backend.py
```diff
@@ -1,4 +1,5 @@
 import copy
+import re
```
JoanFM, line 2: is this needed?

docarray/array/mixins/match.py
```diff
@@ -70,6 +71,11 @@ def match(
         for d in self:
             d.matches.clear()
 
+        if on is not None:
```
JoanFM, line 74: I do not think this `if` else is needed.

docarray/array/storage/base/getsetdel.py
JoanFM: or should `_clear_storage` actually call `_clear_subindices`?

docs/datatypes/text/index.md
```diff
@@ -190,4 +202,163 @@ print(q.matches[:, ('text', 'scores__jaccard')])
 {'value': 0.6666666666666666, 'ref_id': 'f47f7448709811ec960a1e008a366d49'}, 
 {'value': 0.6666666666666666, 'ref_id': 'f47f7448709811ec960a1e008a366d49'}, 
 {'value': 0.7142857142857143, 'ref_id': 'f47f7448709811ec960a1e008a366d49'}]]
-```
\ No newline at end of file
+```
+
+
+## Searching at chunk level with subindex
```
JoanFM, line 208: We should somewhere connect this to `multimodal` access.

JoanFM (commented):
Small comments.

Main comment is about having example linking to the `multimodal` feature

docarray/array/storage/base/getsetdel.py
JohannesMessner: >or should _clear_storage actually call _clear_subindices?
It probably should, conceptually speaking, but `_clear_storage` can be implemented by each backend, so let's keep `_clear_subindices` outside of their responsability.
I'll switch the order though.

docs/datatypes/text/index.md
JohannesMessner: Yes, I think is is still from the very beginning. I will look at this again

docarray/array/storage/elastic/backend.py
JohannesMessner: nope, will remove

docs/advanced/document-store/subindex.md
```diff
@@ -0,0 +1,159 @@
+(subindex)=
+# Subindices
+
+Subindices offer to possibility to efficiently search through multimodal or nested documents.
```
samsja, line 4:
```suggestion
Subindices offer the possibility to efficiently search through multimodal or nested documents.
```

docs/advanced/document-store/subindex.md
```diff
@@ -0,0 +1,159 @@
+(subindex)=
+# Subindices
+
+Subindices offer the possibility to efficiently search through multimodal or nested documents.
```
JohannesMessner, line 4:
```suggestion
Subindices allow you to efficiently search through multimodal or nested Documents.
```

docs/advanced/document-store/subindex.md
```diff
@@ -0,0 +1,159 @@
+(subindex)=
+# Subindices
+
+Subindices allow you to efficiently search through multimodal or nested Documents.
```
samsja, line 4: I think that we are missing a phrase to explain how the subindices allow you to bet efficient.
We need to state that subindices is a different da that the main one and that's why it is more efficient. Maybe a schema in needed.

docs/advanced/document-store/subindex.md
```diff
@@ -0,0 +1,159 @@
...
+
```
samsja, line 5: should we add a link here to the example in text/index.md ? So that people who end up on this page can understand why the feature is needed in some case

JohannesMessner: or the multi-modal one, yep, I agree

samsja (approved):
LGTM. it is a great PR !

tests/unit/array/mixins/test_getset.py
```diff
@@ -392,3 +393,171 @@ def test_zero_embeddings(da_cls, config, start_storage):
...
+
+
+@pytest.mark.parametrize(
+    'storage, config',
+    [
+        ('memory', None),
+        # ('weaviate', {'n_dim': 3, 'distance': 'l2-squared'}),
+        ('annlite', {'n_dim': 3, 'metric': 'Euclidean'}),
+        # ('qdrant', {'n_dim': 3, 'distance': 'euclidean'}),
+        # ('elasticsearch', {'n_dim': 3, 'distance': 'l2_norm'}),
```
nan-wang, line 495: May I ask about the reason for skipping these tests? Please at least add a reminder in the comments for the maintenance purpose.

JohannesMessner: It's a mistake, I enabled them now

docs/advanced/document-store/subindex.md
```diff
@@ -0,0 +1,164 @@
...
+top_image_matches = da.find(query=np.random.rand(512), on='@.[image]')
+# find best matching paragraphs using .match()
+Document(embedding=np.random.rand(128)).match(da, on='@.[paragraph]')
+```
+````
+````{tab} Subindex with chunks
+```python
+# find best matching images using .find()
+top_image_matches = da.find(query=np.random.rand(128), on='@c')
+# find best matching images using .match()
```
nan-wang, line 145:
```suggestion
# find best matching paragraphs using .match()
```

docs/advanced/document-store/subindex.md
```diff
@@ -0,0 +1,164 @@
...
+# find best matching images using .find()
+top_image_matches = da.find(query=np.random.rand(512), on='@.[image]')
+# find best matching paragraphs using .match()
+Document(embedding=np.random.rand(128)).match(da, on='@.[paragraph]')
+```
+````
+````{tab} Subindex with chunks
+```python
+# find best matching images using .find()
+top_image_matches = da.find(query=np.random.rand(128), on='@c')
```
nan-wang, line 144:
```suggestion
top_image_matches = da.find(query=np.random.rand(512), on='@c')
```

docs/advanced/document-store/subindex.md
```diff
@@ -0,0 +1,164 @@
...
+    da.extend(_docs)
+```
+````
+````{tab} Subindex with chunks
+```python
+import numpy as np
+
+# construct DocumentArry with subindices
+da = DocumentArray(
+    config={'n_dim': 256},
```
nan-wang, line 109: 512?

tests/unit/array/mixins/test_getset.py
```diff
@@ -4,6 +4,7 @@
 import tensorflow as tf
 import torch
 from scipy.sparse import csr_matrix
+from torch import nn
```
nan-wang, line 7: I think it is not used.

nan-wang (changes requested):
Minor comments

docs/advanced/document-store/subindex.md
JohannesMessner: The images are on the chunks level so I think it is correct the way it is. Without the use of dataclass API it is not easy to have a nesting with both images and paragraphs, so for simplicity I do everything on the images in the chunk example.

nan-wang: ok, i see.

nan-wang (approved):
LGTM👍

docarray/array/storage/base/getsetdel.py
```diff
@@ -126,6 +127,12 @@ def _del_docs_by_ids(self, ids):
         for _id in ids:
             self._del_doc_by_id(_id)
 
+    def _update_subindices_del(self, ids):
+        if getattr(self, '_subindices', None):
+            for selector, da in self._subindices.items():
+                ids_subindex = DocumentArray(self[ids])[selector, 'id']
+                da._del_docs_by_ids(ids_subindex)
```
AnneYang720, line 134: I think here should be `da._del_docs(ids_subindex)`, to also delete ids in self._offset2ids

PR #460, opened 2022-07-29
feat: add root_id

JohannesMessner:
*Goals:**

This adds `.root_id` to `Document`.
`.root_id` is similar to `.parent_id`, but points _all the way to the top_, instead of just one chunk-level above.

This is particularly useful in combination with [secondary indices](https://github.com/jina-ai/docarray/pull/456), to tie a search back to the top-level document:

```python
best_match = da.find(query=..., secondary_index='@ccc'][0]  # best_match is a chunk of a chunk of a chunk
match_top_level = da[best_match.root_id]  # get top level document
``` 

**Caution**
this is a breaking change, since it changed the protobuf definition

### DIFF
--- a/docarray/document/data.py
+++ b/docarray/document/data.py
@@ -30,7 +30,7 @@ default_values = dict(
     chunks='ChunkArray',
     matches='MatchArray',
     timestamps=dict,
-    root_id='self.id',
+    root_id='__self_id__',
 )
 
 _all_mime_types = set(mimetypes.types_map.values())
@@ -115,7 +115,7 @@ class DocumentData:
                     from docarray.score import NamedScore
 
                     setattr(self, key, defaultdict(NamedScore))
-                elif v == 'self.id':
+                elif v == '__self_id__':
                     setattr(self, key, self.id)
                 else:
                     setattr(self, key, v() if callable(v) else v)
--- a/docarray/document/data.py
+++ b/docarray/document/data.py
@@ -30,6 +30,7 @@ default_values = dict(
     chunks='ChunkArray',
     matches='MatchArray',
     timestamps=dict,
+    root_id='self.id',
 )
 
 _all_mime_types = set(mimetypes.types_map.values())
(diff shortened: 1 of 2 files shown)


docarray/document/data.py
```diff
@@ -30,6 +30,7 @@
     chunks='ChunkArray',
     matches='MatchArray',
     timestamps=dict,
+    root_id='self.id',
```
JoanFM, line 33: put `__self__id__`

JohannesMessner:
**This PR is put on hold** for now, until the secondary index features is done at least for ANNLite and InMemory.

There is the following challenge with `root_id`:
When constructing a nested Document (might have chunks of chunks of chunks etc.), this graph has to be built bottom-up. This means that the leaf Document cannot know the `id` of the root Document yet, since it doesn't even exist at that point. Therefore, `root_id` would have to be computed and propagated down once the graph is fully built.

Now the trouble is this: If a Document graph (doc with chunks of chunks etc.) is inserted into another Document graph, `root_id` needs to be computed. This is, however, indistinguishable from building such a graph for the first time; this build process is made up of one subgraph insertion per granularity level. Therefore, `root_id` would be recomputed for every sub-tree at every granularity, leading to blow-up in the computation.

The potential solution is to set `root_id` not at creation / insertion time, but inside of `find()/match()`. There the entire graph structure is already known, and `root_id` can be efficiently propagated top-down.
This would mean that 1) docs have `root_id` only after having been involved in a search operation, and 2) changing the document graph after a search would invalidate `root_id` until another search is performed.

JohannesMessner:
We decided that the overhead is no worth the benefit in this case, since most scenarios do not require deep nesting.
However, there is one exception: list-types in multi-modal documents look like a single nesting to the user, but are actually two levels of nesting under the hood. This means that there is no straightforward way to get the root doc from a list doc.

So the new plan is the following:

- We make `root_id` as multimodal-only feature
- This will be a `@property` so we don't have to change the proto. Under the hood it will access a field in `_metadata`
- The information in `_metadata` is set only once, when the dataclass is converted to Document. We make no guarantees about any behaviour if the user changes the document tree structure afterwards

PR #465, opened 2022-08-02
fix(elastic): handle bulk operation error info for another op type

alphinside:
Previous Behavior

When delete operation performed  and error occured it will throw these error 

```bash
2022-08-01T19:48:11.223520457Z self._update_offset2ids_meta()
2022-08-01T19:48:11.223527216Z File "/opt/pysetup/.venv/lib/python3.8/site-packages/docarray/array/storage/elastic/backend.py", line 203, in _update_offset2ids_meta
2022-08-01T19:48:11.223553779Z self._send_requests(requests)
2022-08-01T19:48:11.223567841Z File "/opt/pysetup/.venv/lib/python3.8/site-packages/docarray/array/storage/elastic/backend.py", line 165, in _send_requests
2022-08-01T19:48:11.223575146Z failed_index.append(info['index'])
2022-08-01T19:48:11.223581640Z KeyError: 'index'
```

This because it directly parse `'index'` key from the bulk info inside `_send_requests` function

Goals:

- handle the error info from elastic search bulk requests if the operation return error from `_op_type` other than `index` e.g. `delete`

### DIFF
--- a/docarray/array/storage/elastic/seqlike.py
+++ b/docarray/array/storage/elastic/seqlike.py
@@ -69,7 +69,7 @@ class SequenceLikeMixin(BaseSequenceLikeMixin):
 
         return failed_index_ids
 
-    def _upload_batch(self, docs: Iterable['Document']):
+    def _upload_batch(self, docs: Iterable['Document']) -> List[int]:
         batch = []
         accumulated_info = []
         for doc in docs:
--- a/docarray/array/storage/elastic/seqlike.py
+++ b/docarray/array/storage/elastic/seqlike.py
@@ -1,4 +1,4 @@
-from typing import Union, Iterable, Dict
+from typing import Union, Iterable, Dict, List
 import warnings
 
 from docarray.array.storage.base.seqlike import BaseSequenceLikeMixin
@@ -58,34 +58,47 @@ class SequenceLikeMixin(BaseSequenceLikeMixin):
         """
         return f'<{self.__class__.__name__} (length={len(self)}) at {id(self)}>'
 
+    def _parse_failed_index_ids_from_bulk_info(
+        self, accumulated_info: List[Dict]
+    ) -> List[int]:
+        failed_index_ids = []
+
+        for info in accumulated_info:
+            if 'index' in info:
+                failed_index_ids.append(info['index']['_id'])
+
+        return failed_index_ids
+
     def _upload_batch(self, docs: Iterable['Document']):
         batch = []
-        failed_index = []
+        accumulated_info = []
         for doc in docs:
             batch.append(self._document_to_elastic(doc))
             if len(batch) > self._config.batch_size:
-                failed_index.extend(self._send_requests(batch))
+                accumulated_info.extend(self._send_requests(batch))
                 self._refresh(self._config.index_name)
                 batch = []
         if len(batch) > 0:
-            failed_index.extend(self._send_requests(batch))
+            accumulated_info.extend(self._send_requests(batch))
             self._refresh(self._config.index_name)
 
-        return failed_index
+        failed_index_ids = self._parse_failed_index_ids_from_bulk_info(accumulated_info)
+
+        return failed_index_ids
 
     def extend(self, docs: Iterable['Document']):
         docs = list(docs)
-        failed_index = self._upload_batch(docs)
-        failed_ids = [index['_id'] for index in failed_index]
+        failed_index_ids = self._upload_batch(docs)
         self._offset2ids.extend(
             [
                 doc.id
                 for doc in docs
-                if (doc.id not in self._offset2ids.ids) and (doc.id not in failed_ids)
+                if (doc.id not in self._offset2ids.ids)
+                and (doc.id not in failed_index_ids)
             ]
         )
 
-        if len(failed_ids) > 0:
-            err_msg = f'fail to add Documents with ids: {failed_ids}'
+        if len(failed_index_ids) > 0:
+            err_msg = f'fail to add Documents with ids: {failed_index_ids}'
             warnings.warn(err_msg)
             raise IndexError(err_msg)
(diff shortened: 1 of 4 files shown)


docarray/array/storage/elastic/seqlike.py
```diff
@@ -58,34 +58,47 @@ def __repr__(self):
         """
         return f'<{self.__class__.__name__} (length={len(self)}) at {id(self)}>'
 
+    def _parse_failed_index_ids_from_bulk_info(
```
samsja, line 61:
```suggestion
    def _parse_index_ids_from_bulk_info(
```

Could you rename it to _parse_index_ids_from_bulk_info ? The function is more generic than failed index. And could you make it a staticmethod ?

docarray/array/storage/elastic/seqlike.py
```diff
@@ -58,34 +58,47 @@ def __repr__(self):
...
+    ) -> List[int]:
+        failed_index_ids = []
+
+        for info in accumulated_info:
+            if 'index' in info:
+                failed_index_ids.append(info['index']['_id'])
+
+        return failed_index_ids
+
     def _upload_batch(self, docs: Iterable['Document']):
```
samsja, line 72:
```suggestion
    def _upload_batch(self, docs: Iterable['Document']) -> List[int]:
```

samsja (commented):
Thank you for the PR ! I added few minors comments on the code.

 Could you update your PR initial message and show the behavior before and after your PR when you try to delete and except to fail like in your test ?

docarray/array/storage/elastic/seqlike.py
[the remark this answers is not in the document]
alphinside: on it

docarray/array/storage/elastic/seqlike.py
```diff
@@ -58,34 +58,47 @@ def __repr__(self):
...
+        failed_index_ids = self._parse_failed_index_ids_from_bulk_info(accumulated_info)
+
+        return failed_index_ids
 
     def extend(self, docs: Iterable['Document']):
         docs = list(docs)
-        failed_index = self._upload_batch(docs)
-        failed_ids = [index['_id'] for index in failed_index]
+        failed_index_ids = self._upload_batch(docs)
         self._offset2ids.extend(
```
JoanFM, line 92: now seeing the implementation, wouldn't it be more performant to get the successfull IDs?

alphinside: that is make sense, also more readable, I'll try to refactor it

alphinside: done, pls rereview

docarray/array/storage/elastic/seqlike.py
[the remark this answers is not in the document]
alphinside: done pls rereview

alphinside: done pls rereview

docarray/array/storage/elastic/seqlike.py
```diff
@@ -58,34 +58,58 @@ def __repr__(self):
...
                 self._refresh(self._config.index_name)
                 batch = []
         if len(batch) > 0:
-            failed_index.extend(self._send_requests(batch))
+            accumulated_info.extend(self._send_requests(batch))
             self._refresh(self._config.index_name)
 
-        return failed_index
+        successful_ids = self._parse_index_ids_from_bulk_info(accumulated_info)
+        if 'index' not in successful_ids:
```
JoanFM, line 97: what else is in the `successful_ids`?

JoanFM: Why not extracting directly a list from `parse_index_ids_from_bulk_info`?

JoanFM (changes requested):

docarray/array/storage/elastic/seqlike.py
alphinside: as @samsja mention in previous thread that the function `_parse_index_ids_from_bulk_info` could be utilized for a more general purpose, I take this as we could extract any ids from any `_op_type` provided from info ( currently we only extract `_op_type` = `index` in this function but if we maybe need to extract `delete` we can utilize this as welll. So I make that the function return like this 

```python
{
    "index" : [1,2,3,4,5] # successfull indexed ids,
    "delete" : [6,7] # successfull deleted index ids
}
```

tests/unit/array/storage/elastic/test_del.py
```diff
@@ -47,3 +48,43 @@ def test_delete_offset_success_sync_es_offset_index(deleted_elmnts, start_storag
             index=elastic_doc._index_name_offset2id, query={'match': {'blob': id}}
         )['hits']['hits'][0]['_id']
         assert actual_offset_index == expected_offset
+
+
+@pytest.mark.filterwarnings('ignore::UserWarning')
+def test_success_handle_bulk_delete_not_found(start_storage):
```
JoanFM, line 54: I would prefer not to test the private methods. Why can't this be tested with the public API?

JoanFM (changes requested):

tests/unit/array/storage/elastic/test_del.py
alphinside: because when I want try to delete non existed ids, the docarray data structure already handle this and raise error before sending bulk request to elastic search. Thus I test using private method. 

The actual error I get on delete operation is actually not on deleting non existed ids (docarray already guard it), but the following error 

```python
{'delete': {'_index': 'offset2id__docarray_unit_apartement__v0_1', '_id': '52', '_version': 4, 'result': 'not_found', '_shards': {'total': 2, 'successful': 1, 'failed': 0}, '_seq_no': 832, '_primary_term': 2, 'status': 404}}
```

this error is because we are using multiple shards, but data is only found on 1 shard. However I dont really know how to reflect this exact detail on your test. Thus I change it to similar scope ( on delete ops, but change the error on not found error ) with the same handling logic

JoanFM (approved):

samsja (approved):
lgtm

PR #469, opened 2022-08-03
refactor: remove lru_cache decorator from _get_auth_token

guenthermi:
Goals:
When a user logout and login again, the `_get_auth_token` function still returns the old authentification token which is not valid anymore due to the lru_cache.  To prevent this unwanted behavior, I suggest to remove the lru_cache decorator. Moreover, I think the cache is not very useful for the `_get_auth_token` function, since sending a request to hubble is usually much more time-consuming than reading the authentification token.
 
### DIFF
--- a/docarray/array/mixins/io/pushpull.py
+++ b/docarray/array/mixins/io/pushpull.py
@@ -27,7 +27,6 @@ def _get_hub_config() -> Optional[Dict]:
             return json.load(f)
 
 
-@lru_cache()
 def _get_auth_token() -> Optional[str]:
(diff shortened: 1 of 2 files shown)


samsja:
I agree that it is not good to to have the cache on the token since we can't control how long it is valid. 
The pr seems good from my side but I would like someone from hubble to validate as well. @delgermurun @mapleeit

mapleeit (approved):
It makes sense. I'm for it.

delgermurun:
Agreed.

samsja (approved):

Issue #470, opened 2022-08-03
Bug: SQLite document store doesn't handle delete-then-insert properly


JohannesMessner:
The SQLite gets confused about uniqueness conditions when deleting and then adding Documents.
This will throw an Exception:

```python
da = DocumentArray(storage='sqlite')

with da:
    da.extend([Document(id=str(i)) for i in range(3)])
with da:
    del da[1]
    da.append(Document(id='new'))

for d in da:
    print(d.id)
```

But this works fine:

```python
da = DocumentArray(storage='sqlite')

with da:
    da.extend([Document(id=str(i)) for i in range(3)])
with da:
    del da[1]

with da:
    da.append(Document(id='new'))

for d in da:
    print(d.id)
```

The other backends (Elastic, Qdrant, Weaviate) have no issues with either of them.

PR #471, opened 2022-08-03
fix: keep offset2id updated when deleting in sqlite

JohannesMessner:
fixes the bug described here: https://github.com/jina-ai/docarray/issues/470

### DIFF
--- a/tests/unit/array/test_sequence.py
+++ b/tests/unit/array/test_sequence.py
@@ -119,8 +119,19 @@ def test_sqlite_del_and_append(index):
 
 
 @pytest.mark.parametrize('index', [1, '1', slice(1, 2), [1]])
-def test_sqlite_del_and_append(index):
-    da = DocumentArray(storage='sqlite')
+@pytest.mark.parametrize(
+    'storage, config',
+    [
+        ('memory', None),
+        ('weaviate', {'n_dim': 3, 'distance': 'l2-squared'}),
+        ('annlite', {'n_dim': 3, 'metric': 'Euclidean'}),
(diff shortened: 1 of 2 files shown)


tests/unit/array/test_sequence.py
```diff
@@ -103,3 +103,31 @@ def test_context_manager_from_disk(storage, config, start_storage, tmpdir, tmpfi
...
+
+@pytest.mark.parametrize('index', [1, '1', slice(1, 2), [1]])
+def test_sqlite_del_and_append(index):
+    da = DocumentArray(storage='sqlite')
+
+    with da:
+        da.extend([Document(id=str(i)) for i in range(5)])
+    with da:
+        da[1] = Document(id='new')
+        da._save_offset2ids()
```
JoanFM, line 130: this is an awkward test. Why do we need to test something with the private method used?

tests/unit/array/test_sequence.py
```diff
@@ -103,3 +103,31 @@ def test_context_manager_from_disk(storage, config, start_storage, tmpdir, tmpfi
...
+
+@pytest.mark.parametrize('index', [1, '1', slice(1, 2), [1]])
+def test_sqlite_del_and_append(index):
+    da = DocumentArray(storage='sqlite')
+
+    with da:
+        da.extend([Document(id=str(i)) for i in range(5)])
+    with da:
+        del da[1]
+        da._save_offset2ids()
```
JoanFM, line 116: same here?

JoanFM (changes requested):

JoanFM (approved):

PR #473, opened 2022-08-04
feat(elastic): allow kwargs to be pass at extend

alphinside:
Goals:

- Enable elastic search `parallel_bulk` parameter passing from docarray `extend` method


docarray/array/storage/elastic/backend.py
```diff
@@ -81,6 +81,19 @@ def _init_storage(
         self._index_name_offset2id = 'offset2id__' + config.index_name
         self._config = config
 
+        # Only accept several params for es parallel bulk operation
+        accepted_bulk_config = [
+            "thread_count",
```
JoanFM, line 86: please use single quote

docarray/array/storage/elastic/backend.py
```diff
@@ -40,7 +40,7 @@ class ElasticConfig:
     es_config: Dict[str, Any] = field(default_factory=dict)
     index_text: bool = False
     tag_indices: List[str] = field(default_factory=list)
-    batch_size: int = 64
+    bulk_config: Dict[str, Any] = field(default_factory=dict)
```
JoanFM, line 43: is there a way this can be made non breaking change? can we detect if batch size is added?

JoanFM (changes requested):
I am not convinced by passing this in the config? maybe we should just pass *args and **kwargs?

JoanFM:
> I am not convinced by passing this in the config? maybe we should just pass *args and **kwargs?

@samsja @JohannesMessner any opinion?

docarray/array/storage/elastic/backend.py
alphinside: noted

docarray/array/storage/elastic/backend.py
alphinside: sure, will add it

alphinside: done, if the bulk config is provided, the configuration from bulk config will take priority

samsja:
> > I am not convinced by passing this in the config? maybe we should just pass *args and **kwargs?
> 
> @samsja @JohannesMessner any opinion?

I believe we should stick to pass **kwars, *args for now. THe only issue will not be able to pass kwargs when doing operation via python operator

docarray/array/storage/elastic/backend.py
```diff
@@ -159,10 +159,32 @@ def _build_client(self):
...
+    def _send_requests(self, request, **kwargs) -> List[Dict]:
         """Send bulk request to Elastic and gather the successful info"""
+        bulk_config = {}
+
+        # for backward compatibility
+        if 'chunk_size' not in kwargs:
+            bulk_config['chunk_size'] = self._config.batch_size
+
+        # Only accept several params for es parallel bulk operation
+        accepted_config = [
```
JoanFM, line 171: this is not needed. Leave the freedom to user. We do not want to mantain this for every backend. We let the power to the user. We should add the corresponding docstring and that is.

docarray/array/storage/elastic/backend.py
```diff
@@ -159,10 +159,32 @@ def _build_client(self):
         client.indices.refresh(index=self._config.index_name)
         return client
 
-    def _send_requests(self, request) -> List[Dict]:
+    def _send_requests(self, request, **kwargs) -> List[Dict]:
         """Send bulk request to Elastic and gather the successful info"""
+        bulk_config = {}
```
JoanFM, line 164: do not call bulk_config. From our side it should be simply calling kwargs, would not add any semantic meaning

docs/advanced/document-store/elasticsearch.md
```diff
@@ -92,41 +92,77 @@ da2.summary()
...
+
 [0.14890289 0.3168339  0.03050802 0.06785086 0.94719299 0.32490566
  ...]
 ```
 
 Other functions behave the same as in-memory DocumentArray.
 
+### Bulk request configuration
+
+You can configure how bulk requests is being sent to Elasticsearch when adding documents by adding additional `kwargs` when adding documents. Bulk operation is used for several index operations; like (`index` or `delete`). See [the official Documentation](https://elasticsearch-py.readthedocs.io/en/v8.3.3/helpers.html) for more details. Notes that only the following parameters are accepted, the others will be ignored
```
JoanFM, line 124: This is not configuration. This is customization perhaps

JoanFM: Remove reference to what is accepted and what not. User's responsability

JoanFM (changes requested):
Some comments

docarray/array/storage/elastic/backend.py
alphinside: noted

docarray/array/storage/elastic/backend.py
alphinside: noted

docs/advanced/document-store/elasticsearch.md
alphinside: noted

JohannesMessner (approved):
LGTM, thanks for the great contribution!

JoanFM (approved):

PR #478, opened 2022-08-06
feat(elastic): add bulk get operation

alphinside:
Goals:

- Implement a more efficient approach on getting multiple docs by using elastic search client method `mget` rather than `get` one by one

### DIFF
--- a/docarray/array/storage/elastic/getsetdel.py
+++ b/docarray/array/storage/elastic/getsetdel.py
@@ -79,7 +79,7 @@ class GetSetDelMixin(BaseGetSetDelMixin):
                 for doc in es_docs
                 if doc['found'] is True
             ]
-            docs_not_found = [doc['_id'] for doc in es_docs if doc['found'] is False]
+            docs_id_not_found = [doc['_id'] for doc in es_docs if doc['found'] is False]
 
             accumulated_docs.extend(docs)
             accumulated_docs_not_found.extend(docs_not_found)
--- a/docarray/array/storage/elastic/getsetdel.py
+++ b/docarray/array/storage/elastic/getsetdel.py
@@ -60,12 +60,15 @@ class GetSetDelMixin(BaseGetSetDelMixin):
         :param ids:  ids of the document
         :return: Iterable[Document]
         """
+        accumulated_docs = []
+        accumulated_docs_not_found = []
+
+        if len(ids) < 1:
+            return accumulated_docs
+
         # Handle if doc len is more than MAX_ES_RETURNED_DOCS
         nrof_arr_split = math.ceil(len(ids) / MAX_ES_RETURNED_DOCS)
         split_ids = np.array_split(ids, nrof_arr_split)
-
-        accumulated_docs = []
-        accumulated_docs_not_found = []
         for split in split_ids:
             es_docs = self._client.mget(index=self._config.index_name, ids=split)[
                 'docs'
--- a/docarray/array/storage/elastic/getsetdel.py
+++ b/docarray/array/storage/elastic/getsetdel.py
@@ -85,7 +85,7 @@ class GetSetDelMixin(BaseGetSetDelMixin):
             accumulated_docs_not_found.extend(docs_not_found)
 
         if len(accumulated_docs_not_found) > 0:
-            raise KeyError(accumulated_docs_not_found)
+            raise KeyError((accumulated_docs_not_found,accumulated_docs))
 
         return accumulated_docs
 
--- a/docarray/array/storage/elastic/getsetdel.py
+++ b/docarray/array/storage/elastic/getsetdel.py
@@ -63,7 +63,7 @@ class GetSetDelMixin(BaseGetSetDelMixin):
         accumulated_docs = []
         accumulated_docs_not_found = []
 
-        if len(ids) < 1:
+        if not ids
             return accumulated_docs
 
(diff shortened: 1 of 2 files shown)


docarray/array/storage/elastic/getsetdel.py
```diff
@@ -1,9 +1,13 @@
-from typing import Iterable, Dict
+from typing import Iterable, Dict, Sequence
+import math
+import numpy as np
 
 from docarray.array.storage.base.getsetdel import BaseGetSetDelMixin
 from docarray.array.storage.base.helper import Offset2ID
 from docarray import Document
 
+MAX_ES_RETURNED_DOCS = 10000
```
samsja, line 9: lets put this const variable in the `GetSelDelMixin` class as a static attribute

docarray/array/storage/elastic/getsetdel.py
```diff
@@ -50,6 +54,41 @@ def _get_doc_by_id(self, _id: str) -> 'Document':
...
+    def _get_docs_by_ids(self, ids: Sequence[str]) -> Iterable['Document']:
+        """Concrete implementation of base class' ``_get_docs_by_ids``
+
+        :param ids:  ids of the document
+        :return: Iterable[Document]
+        """
+        accumulated_docs = []
+        accumulated_docs_not_found = []
+
+        if len(ids) < 1:
```
samsja, line 66:
```suggestion
        if not ids
```

docarray/array/storage/elastic/getsetdel.py
```diff
@@ -50,6 +54,41 @@ def _get_doc_by_id(self, _id: str) -> 'Document':
...
+                for doc in es_docs
+                if doc['found'] is True
+            ]
+            docs_not_found = [doc['_id'] for doc in es_docs if doc['found'] is False]
+
+            accumulated_docs.extend(docs)
+            accumulated_docs_not_found.extend(docs_not_found)
+
+        if len(accumulated_docs_not_found) > 0:
+            raise KeyError(accumulated_docs_not_found)
```
samsja, line 88:
```suggestion
            raise KeyError((accumulated_docs_not_found,accumulated_docs))
```

let's return the accumulated docs in the as well in the exception. So that we can still retrieve the document even if one of the ID is wrong by cathing the exeption

docarray/array/storage/elastic/getsetdel.py
```diff
@@ -50,6 +54,41 @@ def _get_doc_by_id(self, _id: str) -> 'Document':
...
+            es_docs = self._client.mget(index=self._config.index_name, ids=split)[
+                'docs'
+            ]
+
+            docs = [
+                Document.from_base64(doc['_source']['blob'])
+                for doc in es_docs
+                if doc['found'] is True
+            ]
+            docs_not_found = [doc['_id'] for doc in es_docs if doc['found'] is False]
```
samsja, line 82:
```suggestion
            docs_id_not_found = [doc['_id'] for doc in es_docs if doc['found'] is False]
```
lets rename it to show that it is the ids and not the docs directly

docarray/array/storage/elastic/getsetdel.py
```diff
@@ -50,6 +54,41 @@ def _get_doc_by_id(self, _id: str) -> 'Document':
...
+
+            accumulated_docs.extend(docs)
+            accumulated_docs_not_found.extend(docs_not_found)
```
samsja, line 85:
```sugestion
            accumulated_docs_id_not_found.extend(docs_not_found)

```
Same renaming here

samsja:
@alphinside great PR again. It will improve perf for sure

I added some comments.

Can you add more information to the PR description ? Just precise that you are using `mget` from the ES client rather than the `get` to get thing by bulk.

docarray/array/storage/elastic/getsetdel.py
alphinside: noted

tests/unit/array/storage/elastic/test_get.py
```diff
@@ -0,0 +1,54 @@
+from docarray import Document, DocumentArray
+import numpy as np
+import pytest
+
+
+def test_success_get_bulk_data(start_storage):
+    nrof_docs = 20000
```
samsja, line 7:
```suggestion
@pytest.mark.parametrize('nrof_docs', [10, 100, 10_000, 10_100, 20_000, 20_100])
def test_success_get_bulk_data(start_storage, nrof_docs):

```

let's just test different values.

tests/unit/array/storage/elastic/test_get.py
```diff
@@ -0,0 +1,54 @@
...
+            [
+                Document(id=f'r{i}', embedding=np.ones((3,)) * i)
+                for i in range(nrof_docs)
+            ]
+        )
+
+    with pytest.raises(KeyError) as e:
+        elastic_doc[['r1', 'r11', 'r21'], 'id']
+
+    assert e.value.args[0] == ['r11', 'r21']
```
samsja, line 54:
```suggestion
    assert e.value.args[0] == ['r11', 'r21']
    assert len(e.value.args[1]) == 1
```

lets check that we can get the retrieve document as well

samsja (changes requested):
We are almost good to go. I just added two comments on the tests

PR #481, opened 2022-08-10
feat(elastic): add find kwargs for perf tuning

alphinside:
Goals:

- Enable Elastic Search knn search performance tuning by configure `num_candidates` param during `find` method call

### DIFF
--- a/docs/advanced/document-store/elasticsearch.md
+++ b/docs/advanced/document-store/elasticsearch.md
@@ -219,7 +219,7 @@ Embeddings Nearest Neighbours with "price" at most 7:
  embedding=[4. 4. 4.],  price=4
  ```
 
-Additionally you can tune the approximate kNN for speed or accuracy by providing `num_candidates` kwargs when calling the `find` method
+Additionally you can tune the approximate kNN for speed or accuracy by providing `num_candidates` kwarg when calling the `find` method:
 
 ```python
 results = da.find(np_query, filter=filter, limit=n_limit, num_candidates=100)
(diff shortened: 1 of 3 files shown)


samsja (approved):
LGTM, let's wait for @JohannesMessner and/or @nan-wang before merging.

Nice PR :smiley:

docs/advanced/document-store/elasticsearch.md
```diff
@@ -219,6 +219,16 @@ Embeddings Nearest Neighbours with "price" at most 7:
  embedding=[4. 4. 4.],  price=4
  ```
 
+Additionally you can tune the approximate kNN for speed or accuracy by providing `num_candidates` kwargs when calling the `find` method
```
JohannesMessner, line 222:
```suggestion
Additionally you can tune the approximate kNN for speed or accuracy by providing `num_candidates` kwarg when calling the `find` method:
```

JohannesMessner (commented):
Looking good to me as well!
Just one thought:
- Are there other kwargs that could be passed to the `self._client.knn_search(...)` call, other than `num_candidates`? If so then we should enable all of those kwargs to be passed on.

alphinside:
> Looking good to me as well! Just one thought:
> 
> * Are there other kwargs that could be passed to the `self._client.knn_search(...)` call, other than `num_candidates`? If so then we should enable all of those kwargs to be passed on.

I'm not really sure whether any other params will need to be configured to support and not breaking docarray functionalities, can you give suggestion? https://elasticsearch-py.readthedocs.io/en/v8.3.3/api.html#elasticsearch.Elasticsearch.knn_search

In my opinion, only `num_candidates` can be changed flexibly which provided via `knn` kwarg. `routing` param may give some freedom to point out to some es shards, but currently I'm not really have sufficient knowledge for this, also I don't see any index creation on specific shards on docarray

JohannesMessner (approved):
Looks good then!
There are no more suggestions from my side, I was just wondering if there is other stuff that one could potentially pass to `self._client.knn_search()`, since I am not so familiar with the ES client itself. If that's not the case then we are all set!

PR #488, opened 2022-08-15
fix: set subindices directly via access path

JohannesMessner:
Issue is described here: https://github.com/jina-ai/docarray/issues/489

### DIFF
--- a/docarray/array/storage/base/getsetdel.py
+++ b/docarray/array/storage/base/getsetdel.py
@@ -11,6 +11,19 @@ from docarray.array.storage.base.helper import Offset2ID
 from docarray import Document, DocumentArray
 
 
+def _check_valid_values_nested_set(docs, values):
+    docs, values = DocumentArray(docs), DocumentArray(values)
+    if len(docs) != len(values):
+        raise ValueError(
+            f'length of docs to set({len(docs)}) does not match '
+            f'length of values({len(values)})'
+        )
+    if docs[:, 'id'] != values[:, 'id']:
+        raise ValueError(
+            'Setting Documents by traversal paths with different IDs is not supported'
+        )
+
+
 class BaseGetSetDelMixin(ABC):
     """Provide abstract methods and derived methods for ``__getitem__``, ``__setitem__`` and ``__delitem__``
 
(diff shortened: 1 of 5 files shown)


docarray/array/storage/base/getsetdel.py
```diff
@@ -173,15 +186,26 @@ def _set_docs_by_ids(self, ids, docs: Iterable['Document'], mismatch_ids: Dict):
...
+        if isinstance(set_index, tuple):  # handled later in recursive call
+            return
+        if isinstance(set_index, str) and set_index.startswith('@'):
+            # 'nested' (non root-level) set, update entire subindex directly
+            _check_valid_values_nested_set(DocumentArray(self[set_index]), docs)
+            if set_index in subindices.keys():
+                subindex_da = subindices[set_index]
+                with subindex_da:
+                    subindices[set_index].clear()
+                    subindices[set_index].extend(docs)
```
alaeddine-13, line 202: why not
```suggestion
                with subindex_da:
                    subindex_da.clear()
                    subindex_da.extend(docs)
```

docarray/array/storage/base/getsetdel.py
```diff
@@ -173,15 +186,26 @@ def _set_docs_by_ids(self, ids, docs: Iterable['Document'], mismatch_ids: Dict):
...
-                    da.extend(DocumentArray(docs)[selector])  # same hack here
+    def _update_subindices_set(self, set_index, docs):
+        subindices = getattr(self, '_subindices', None)
+        if not subindices:
+            return
+        if isinstance(set_index, tuple):  # handled later in recursive call
+            return
+        if isinstance(set_index, str) and set_index.startswith('@'):
+            # 'nested' (non root-level) set, update entire subindex directly
+            _check_valid_values_nested_set(DocumentArray(self[set_index]), docs)
```
alaeddine-13, line 197: do we need to wrap into `DocumentArray` ? it's cast to DocumentArray inside the check function

docarray/array/storage/base/getsetdel.py
```diff
@@ -173,15 +186,26 @@ def _set_docs_by_ids(self, ids, docs: Iterable['Document'], mismatch_ids: Dict):
...
+            if set_index in subindices.keys():
```
alaeddine-13, line 198:
```suggestion
            if set_index in subindices:
```

docarray/array/storage/base/getsetdel.py
```diff
@@ -173,15 +186,26 @@ def _set_docs_by_ids(self, ids, docs: Iterable['Document'], mismatch_ids: Dict):
...
+                subindex_da = subindices[set_index]
+                with subindex_da:
+                    subindices[set_index].clear()
+                    subindices[set_index].extend(docs)
+        else:  # root level set, update subindices iteratively
+            for subindex_selector, subindex_da in subindices.items():
+                old_ids = DocumentArray(self[set_index])[subindex_selector, 'id']
+                with subindex_da:
+                    del subindex_da[old_ids]
+                    subindex_da.extend(DocumentArray(docs)[subindex_selector])
```
alaeddine-13, line 208: probably this is not the correct way to update the subindex but I'm afraid there is no better efficient way:
if you delete document chunks (the root document is positioned in a middle position) from the subindex and you append its new chunks to the subindex, we are changing the order of docs within the subindex, and I think order matters in some minor cases

JohannesMessner: This is correct, we knew about this and decided that we don't want to make guarantees about the order in the subindex. If you care about it i think you can still get the order from the serialized chunks that are stored in the root index

docarray/array/storage/base/getsetdel.py
JohannesMessner: You are right, I'll change it

JohannesMessner, line 197:
```suggestion
            _check_valid_values_nested_set(self[set_index], docs)
```

alaeddine-13:
I'm also wondering, shouldn't we have the same logic for _update_subindices_del ?

alaeddine-13 (approved):

PR #496, opened 2022-08-22
fix: update offset2id when deleting in subindex

JohannesMessner:
Goals:

make sure that `offset2ids` is correctly updated when deleting in a subindex.
Previously, the call to `del_docs_by_ids()` did not do it, but using `del` does it under the hood.


JohannesMessner:
@anneyang720 this is a PR for your suggestion, thanks!

alaeddine-13 (approved):

samsja (approved):

PR #504, opened 2022-08-26
fix: annlite offsetmapping

numb3r3:
Goals:

-  1. fix the offsetmapping in annlite, (`in_memory=False`)
-  2. `__del__` should be placed in `getsetdel.py`
-  3. should close lmdb cursor in `__del__` so that lmdb can be reused in another annlite instance


samsja:
Will this PR be compatible with old annlite version ? 

This to know if we have to pin the annlite version to the last one or if it is retro compatible

alaeddine-13 (approved):
LGTM

numb3r3:
> Will this PR be compatible with old annlite version ?

yes

PR #506, opened 2022-08-28
Change journal mode to WAL

linkerlin:
THE `WAL` mode is mush faster then `DELETE` mode for the sqlite.

Goals:

- speed up the sqlite backend


hanxiao:
hi thanks for your contribution, unfortunately, if the commit message is incorrect the CI wont continue and your contribution wont be credited by the release. Could you please do
```bash
git commit --amend
# change the message to `fix(store): use wal instead of delete for sqlite backend` without quote
git force push
```

PR #523, opened 2022-09-05
fix: consider tuples may not be created from yaml

JoanFM:
Goals:

Try without breaking to enable the usage of `columns` in backend configurations when they are started from YAML as Lists.


alaeddine-13 (approved):
I prefer casting to tuple in intialization (there's some _normalize_columns method used). We can avoid all these changes

Issue #461, opened 2022-08-01
Add `root_id` to Document


JohannesMessner:
`root_id` is like `parent_id`, but instead of pointing one level above, it points all the way to the top (root).

https://github.com/jina-ai/docarray/pull/460

JohannesMessner:
We decided that the overhead is no worth the benefit in this case, since most scenarios do not require deep nesting.
However, there is one exception: list-types in multi-modal documents look like a single nesting to the user, but are actually two levels of nesting under the hood. This means that there is no straightforward way to get the root doc from a list doc.

So the new plan is the following:

- We make `root_id` as multimodal-only feature
- This will be a `@property` so we don't have to change the proto. Under the hood it will access a field in `_metadata`
- The information in `_metadata` is set only once, when the dataclass is converted to Document. We make no guarantees about any behaviour if the user changes the document tree structure afterwards

JoanFM:
> We decided that the overhead is no worth the benefit in this case, since most scenarios do not require deep nesting. However, there is one exception: list-types in multi-modal documents look like a single nesting to the user, but are actually two levels of nesting under the hood. This means that there is no straightforward way to get the root doc from a list doc.
> 
> So the new plan is the following:
> 
> * We make `root_id` as multimodal-only feature
> * This will be a `@property` so we don't have to change the proto. Under the hood it will access a field in `_metadata`
> * The information in `_metadata` is set only once, when the dataclass is converted to Document. We make no guarantees about any behaviour if the user changes the document tree structure afterwards

I do not like this, this will make this feature almost not accessible from non-python clients. Changing the proto should not be a big problem if the root_id is passed at the end of the list of attributes.

I would say, either we go with it or we do not, we do not go only half way

JohannesMessner:
Ok I see your point; the main reason behind making it a property was because in this design it would be multi-modal only. If it is a real field but not really supported for vanilla Documents that would be weird imo.

Let me think about a way that we can "go for it" with multi-modal and vanilla docs, while also avoiding the big overheads. Otherwise I'll put it on hold again.

JoanFM:
Will not do as it is

PR #531, opened 2022-09-08
feat: combine embeddings

### LINKED ISSUE #512: Add `.combine_embeddings()` method do DocumentArray
Right now it is painful to combine embeddings from different nesting levels and set them at the top level, especially until [`root_id`](https://github.com/jina-ai/docarray/issues/461) is solved.

So this should do the following: It takes an access path (or list of access paths?) (e.g. @.[image, main_text]), and combines the embeddings in those docs using one of 'sum', 'mean', 'concat', or a provided model.

JohannesMessner:
Right now it is painful to combine embeddings from different nesting levels and set them at the top level, especially until https://github.com/jina-ai/docarray/issues/461 is solved.

So this should do the following: It takes an access path (or list of access paths?) (e.g. @.[image, main_text]), and combines the embeddings in those docs using one of 'sum', 'mean', 'concat', or a provided model.

**It should handle the following cases:**

- [x] numpy embedding / predefined combiner
  - [x] test
- [x] torch embedding / predefined combiner
  - [x] test
- [ ] tf embedding / predefined combiner
  - [ ] test
- [ ] paddle embedding / predefined combiner
  - [ ] test
 ----
- [x] torch  embedding model
  - [ ] test
- [ ] tf embedding model
  - [ ] test
- [ ] paddle embedding model
  - [ ] test
- [ ] onnx embedding model
  - [ ] test  
----
- [x] numpy embedding / callable
  - [ ] test
- [x] torch embedding / callable
  - [ ] test
- [x] tf embedding / callable
  - [ ] test
- [x] paddle embedding / callable
  - [ ] test

**Other ToDos:**
- [ ] docs
- [ ] rename the method: `fuse_embeddings()`?
- [ ] refactor. Currently the code is quite ugly
- [ ] refactor examples in the docs

**Possible follow-up PRs:**
- [ ] enable a flag `uniform_nesting` which tells us that every doc in the da has the same number of relevant chunks. This would allow us to vectorize the combine operation
- [ ] implement to_numpy
- [ ] implement flag to discard chunk embeddings after root embedding has been set

Closes #512


JohannesMessner:
Note on the current implementation: It works on a per-Document basis and then `apply()`s this over all docs in a DocumentArray.

Obviously, this can be expensive in terms of performance, since it doesn't utilize batched/vectorized operations of a provided model or other combiner.

This is done for the following reason: Documents might not have a uniform structure, e.g. the first doc has 5 chunks, the second one has only 3. This is not a problem for a combiner like `sum`, or `mean` (or any reasonable embedding model), since it can take multiple inputs and projects them into a fixed length vector.
But a structure like this prohibits batching/vectorizing, since tensor shapes would not be the same.

Thus, in order to allow maximum compatability, we have no choice but to operate on a per-doc basis, at the cost of performance.

In a second iteration we could detect if a docarray has uniform nesting, and then switch over to a fast vectorized implementation.

bwanglzu:
great feature, like it! i wouldn't name it as `combine_embeddings`, but something like `fuse_embeddings` or something related. `fusion` is a common term in the multimedia community to "blend" the representations. Other than that will review once it's done.

Should we still maintain paddle?

PR #543, opened 2022-09-19
fix: adapt to annlite changes

alaeddine-13:
annlite in 0.3.12, renames private method `_search_documents` which was used by docarray and makes it a public method.
Therefore, we correct the method name in docarray


numb3r3 (approved):
LGTM

Issue #549, opened 2022-09-21
Pipeline hangs when using load_uri_to_image_tensor with large or error images


hanxiao:
Title generated by GPT3
Source: https://jina-ai.slack.com/archives/C0169V26ATY/p1663780772405119
Please consult the original message for more details. I will not follow the thread.

---
We are using the load_uri_to_image_tensor to load images from various endpoints. I found that for some endpoints, if the image is very large or if I get an error response when trying to urlopen , the whole embedding/indexing pipeline seems to hang. Adding a timeout and a try/except statement to _uri_to_blob helped get around this error (see below) Is there a way to achieve the same thing with the built-in parameters provided by docarray / jina?

```python
def _uri_to_blob(uri: str) -> bytes:
    """Convert uri to blob
    Internally it reads uri into blob.
    :param uri: the uri of Document
    :return: blob bytes.
    """
    if urllib.parse.urlparse(uri).scheme in {'http', 'https', 'data'}:
        try:
            req = urllib.request.Request(uri, headers={'User-Agent': 'Mozilla/5.0'})
            with urllib.request.urlopen(req, timeout=5) as fp:
                return fp.read()
        except Exception as e: 
            raise FileNotFoundError(f'`{uri}`: error pulling')
    elif os.path.exists(uri):
        try:
            with open(uri, 'rb') as fp:
                return fp.read()
        except Exception as e:
            raise FileNotFoundError(f'`{uri}`: error pulling')
    else:
        raise FileNotFoundError(f'`{uri}` is not a URL or a valid local path')
```

TEChopra1000:
Happy to answer questions about this issue. It would be helpful to have some better exception handling here, and more granular control of urllib.request timeouts.

JoanFM:
Hey @TEChopra1000,

Can you check if #556 would solve your issue?

Thanks

PR #556, opened 2022-09-22
feat: allow pass parameters to load_uri_to*

JoanFM:
Goals:

Capture properly Exceptions from `load_to_uri*` methods and allow to pass extra keyword arguments as `timeout`

### DIFF
--- a/docarray/document/mixins/blob.py
+++ b/docarray/document/mixins/blob.py
@@ -9,13 +9,14 @@ if TYPE_CHECKING:
 class BlobDataMixin:
     """Provide helper functions for :class:`Document` to handle binary data."""
 
-    def load_uri_to_blob(self: 'T') -> 'T':
+    def load_uri_to_blob(self: 'T', **kwargs) -> 'T':
         """Convert :attr:`.uri` to :attr:`.blob` inplace.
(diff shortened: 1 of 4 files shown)


TEChopra1000 (commented):
This looks good. Thanks for the quick turn around! Do you think a try/exception statement would help with error handling?

JoanFM:
> This looks good. Thanks for the quick turn around! Do you think a try/exception statement would help with error handling? I.e. #549

Hey @TEChopra1000,

This should not affect, since the Exceptions should be directly raised.

samsja:
What if someone want to change a parameters from another function that is called inside this one ?

In this pr we only assume that user would only want to change `urllib.request.urlopen` parameters but why not could he want to change `req = urllib.request.Request` as well for instance ?

JoanFM:
> What if someone want to change a parameters from another function that is called inside this one ?
> 
> In this pr we only assume that user would only want to change `urllib.request.urlopen` parameters but why not could he want to change `req = urllib.request.Request` as well for instance ?

Do you suggest simply to add timeout to `_uri_to_blob` but keep `**kwargs` in the API public functions?

samsja (approved):

PR #559, opened 2022-09-23
fix: keep redis default values for vector similarity parameters

alaeddine-13:
closes: https://github.com/jina-ai/docarray/issues/557
Avoid introducing new default values for vector similarity configuration in redis backend, rather use None and leave it up to the database to decide the default value.


JoanFM (approved):

samsja (approved):
lgtm

I really like this pattern we should always use it as well when passing kwargs to other function

bwanglzu:
fyi @AnneYang720

Issue #545, opened 2022-09-20
Concatenation feature in DocumentArray is broken


Jackmin801:
Example code is taken from here: https://docarray.jina.ai/fundamentals/documentarray/construct/#construct-from-multiple-documentarray
```
from docarray import DocumentArray

da1 = DocumentArray.empty(3)
da2 = DocumentArray.empty(4)
da3 = DocumentArray.empty(5)
print(da1 + da2 + da3)

da1 += da2
print(da1)
```

Expected Behavior:
```
<DocumentArray (length=12) at 5024988176>
<DocumentArray (length=7) at 4525853328>
```

Current Behavior:
```
<DocumentArray (length=12) at 140264677317840>
<DocumentArray (length=16) at 140264677380944>
```

Reproduction of the bug in colab environment (with a bit more experimentation on what might be going on):
https://colab.research.google.com/drive/1XySeJY_A7d5rf7n0FIa27snBxJawrASR?usp=sharing

PR #582, opened 2022-10-02
fix(array): fix add not performing deep copy

Jackmin801:
Goals:

- [x] Fix https://github.com/jina-ai/docarray/issues/545


Jackmin801:
There is another occurrence of this [code](https://github.com/jina-ai/docarray/blob/main/docarray/array/base.py#L16). Might want to fix that one too.
But I don't know where this is used though. Wouldn't be able to test it.

alaeddine-13 (approved):
Thanks for your contribution, lgtm, will check if CI is passing

alaeddine-13:
@Jackmin801  it looks like some tests are failing in tests/unit/document/test_plot.py, can you take a look ?

Jackmin801:
Is the large drop in codecov normal? Not familiar with using codecov but looks pretty alarming to me.
How would I go about figuring out why the codecov dropped so much?

Issue #544, opened 2022-09-20
Extend operation on DocumentArray applies change before validation


Jackmin801:
```python
from docarray import Document

d = Document(id='Meow')
d.chunks.extend([1, 2, 3])
d.summary()
```

**Expected Behavior**
Throws an error and keeps the Document intact.

**Current Behavior**
Throws an error but modifies the Document.

Reproduction of the bug in colab environment:
https://colab.research.google.com/drive/1n0olEkZ9iKSIvtq1kXbLv5hv59MsSv1L?usp=sharing

Jackmin801:
Maybe bug? Opening again

Jackmin801:
I think it might be worth solving because someone may “Ask forgiveness, not permission” in their code and use a try except case in this scenario. The try except would not work because even though the exception is hit, the damage is done.

Not sure of the tradeoffs though. But if it's as simple as changing an evaluation order then I think it's worth having.

JoanFM:
I guess the code is to reimplement some of the `dunder` methods as `__add__` and all this `Pythonic` ContainerSeqLike objects, but yes it may be worth.

samsja:
It is understandable why this fail, the chunk in just a MutableSequence that can be extend by anything. 

We should do a check, as @Jackmin801 said this is about tradeoff with performance. ( I wish sometime Python could do type checking without overhead).

The code that does that is here : https://github.com/jina-ai/docarray/blob/main/docarray/array/storage/base/seqlike.py#L64-L70

IMO we should check anyway. It is just adding one if statement

samsja:
More in depth look. 

Actually we don't even do extend by bulk, we do a for loop and use append (this is because we need to keep track of the offset2id so this is the only solution). So there is a very minimal overhead to check if the value is a Document or not.

I don't see any reason not to do it @JoanFM

JoanFM:
So let's do this @samsja , it should be quick

JoanFM:
Decided not to pursue. Reasons explained in the PR opened

PR #583, opened 2022-10-02
fix(array): in memory extend should check types before extend

Jackmin801:
Goals:

- Fixes https://github.com/jina-ai/docarray/issues/544


JoanFM (changes requested):
Can you please provide a performance comparison of the different implementations? 

I am not sure it is a desired change to add

PR #587, opened 2022-10-04
feat: milvus document store

JohannesMessner:
# Goals

Add Milvus as a supported Document Store

## ToDo

**Outline:**

- [x] file structure
- [x] implement getsetdel
  - [x] optimized methods
- [x] implement seqlike
  - [x] optimized methods
- [x] implement backend
- [x] implement find
- [x] implement init
- [x] documentation
  - [x] docstrings
  - [x] backend summary page
- [x] tests
  - [x] tests for specifics
    - [x] specific config params
    - [x] loading behaviour (decorator)
- [x] ~~update benchmark~~ needs to happen in a separate PR after this is merged
- [x] ~~add `search_params` in the configuration (and document it)~~ document kwargs that can be passed to find etc.
- [x] Bug: `limit` does not work for filter-only `find` call
- [x] Issue: `.extend()` does not use bulk operations

**Other stuff:**
- [x] find better way to store `offset2ids` that does not require a new collection
- [x] use context manager to load and release database to/from memory

**New dependencies:**
This PR adds one new `test` dependency: `pytest-mock`.

**Known limitations:**
- Cannot store documents with large-ish tensors. This is because str type is not yet supported in Milvus, so the serialized document has to to into a varchar column. If the tensor is too large, the base64 encoded document is too long for the maximum allowed varchar length.
- due to slow collection load and release, the `with da:` context manager should be used whenever accessing more than a few documents in a non-batched way.
- ~~the mechanism that enables collection loading in the `with da:` context manager logs some errors to the command line; this is not an issue, since these backend errors are expected and handled properly, but it is irritating for the user. Currently investigating if Milvus offers a better API to tackle this issue.~~ this is solved


docarray/array/storage/milvus/getsetdel.py
```diff
@@ -1,69 +1,66 @@
...
+        self._collection.load()
         res = self._collection.query(
-            expr=f'document_id in {id_list_str}',
-            # output_fields=["book_id", "book_intro"],
+            expr=f'document_id in {ids_to_milvus_expr(ids)}',
+            output_fields=['serialized'],
             consistency_level='Strong',
         )
-        # TODO(johannes) handle output an convert to da
+        self._collection.release()
```
JoanFM, line 53: this looks like something that would require a context manager usage

JoanFM (changes requested):

docarray/array/storage/milvus/getsetdel.py
JohannesMessner: you from their side, or our `with da`?

I was already thinking about putting this in our context manager, but that would tie this memory loading/releasing to the updating of offset2id. I /think/ that is what we want, but we need to think about corner cases.

I think I will keep it like this for now in order to make fast progress on all the features that we have to support, but we have to keep this in mind for sure.

JoanFM: I was saying from their side? They do not offer this as ctxt manager? if not, we need to do the `try: finally` manually

JoanFM (changes requested):

docarray/array/storage/milvus/getsetdel.py
JohannesMessner: No they don't. I think the following is the best course of action, also in order to avoid loading and releasing constantly when it is not necessary:

- we have an internal context manager that does the load and release wherever it is needed; basically just wrapping their load and release
- the `with da` also does the load and release, so that we can avoid loading and releasing at every operation
- If the we are in the scope of `with da`, the internal context manager does nothing, since `with da` handles it; if we are outside of such a scope, the internal context manager does it

Is this overkill?

JoanFM (changes requested):

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,456 @@
...
+
+(milvus-limitations)=
+## Known limitations of the Milvus Document Store
+
+The Milvus Document Store implements the entire DocumentArray API, but there are some limitations that you should be aware of.
+
+(milvus-collection-loading)=
+### Collection loading
+
+In Milvus, every search or query operation requires the index to be loaded into memory.
```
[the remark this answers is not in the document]
JohannesMessner: I assume so, though Milvus' documentation doesn't specify.

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+
+```bash
+docker-compose up
+```
+
+You can find more installation guidance in the [Milvus documentation](https://milvus.io/docs/v2.1.x/install_standalone-docker.md).
+
+### Create DocumentArray with Milvus backend
+
+Assuming service is started using the default configuration (i.e. the server's gRPC address `http://localhost:19530`), you can
```
NicholasDunham, line 83:
```suggestion
Assuming the service is started using the default configuration (i.e. the server's gRPC address is `http://localhost:19530`), you can 
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+Assuming service is started using the default configuration (i.e. the server's gRPC address `http://localhost:19530`), you can 
+instantiate a DocumentArray with Milvus storage like so:
+
+```python
+from docarray import DocumentArray
+
+da = DocumentArray(storage='milvus', config={'n_dim': 10})
+```
+
+Here, `config` is configuration for the created Milvus collection,
```
NicholasDunham, line 92:
```suggestion
Here, `config` is configuration for the new Milvus collection,
```

Not essential, but I think it would be good to make clear that this is configuration for a newly-created collection, not a previously-created one.

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+and `n_dim` is a mandatory field that specified the dimensionality of stored embeddings.
```
NicholasDunham, line 93:
```suggestion
and `n_dim` is a mandatory field that specifies the dimensionality of stored embeddings.
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+You can find a complete specification of the Milvus `config` {ref}`here <milvus-config>`.
```
NicholasDunham, line 94:
```suggestion
For more information about the Milvus `config`, refer to the {ref}`specification <milvus-config>`.
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+
+To access a previously persisted DocumentArray, you can specify the `collection_name`, the `host`  and the `port`.
```
NicholasDunham, line 96:
```suggestion
To access a previously persisted DocumentArray, specify the `collection_name`, the `host`,  and the `port`. 
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+| `collection_name`   | Qdrant collection name client                                                                                                                                                                                                                 | **Random collection name generated**                 |
+| `host`              | Hostname of the Milvus server                                                                                                                                                                                                                 | 'localhost'                                          |
+| `port`              | port of the Milvus server                                                                                                                                                                                                                     | 6333                                                 |
```
NicholasDunham, line 125:
```suggestion
| `port`              | Port of the Milvus server                                                                                                                                                                                                                     | 6333                                                 |
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+| `distance`          | Distance metric to be used during search. Can be 'IP', 'L2', 'JACCARD', 'TANIMOTO', 'HAMMING', 'SUPERSTRUCTURE' or 'SUBSTRUCTURE'.                                                                                                            | 'IP' (inner product)                                 |
+| `index_type`        | Type of the (ANN) search index. Can be 'HNSW', 'FLAT', 'ANNOY', or one of multiple variants of IVF and RHNSW. A full list of supported index types can be found [here](https://milvus.io/docs/v2.1.x/build_index.md#Prepare-index-parameter). | 'HNSW                                                |
```
NicholasDunham, line 127:
```suggestion
| `index_type`        | Type of the (ANN) search index. Can be 'HNSW', 'FLAT', 'ANNOY', or one of multiple variants of IVF and RHNSW. Refer to the [list of supported index types](https://milvus.io/docs/v2.1.x/build_index.md#Prepare-index-parameter). | 'HNSW'                                               |
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+| `index_params`      | A dictionary of parameters used for index building. The allowed parameters depend on the index type, and can be found [here](https://milvus.io/docs/v2.1.x/index.md).                                                                         | {'M': 4, 'efConstruction': 200} (assumes HNSW index) |
```
NicholasDunham, line 128:
```suggestion
| `index_params`      | A dictionary of parameters used for index building. The [allowed parameters](https://milvus.io/docs/v2.1.x/index.md) depend on the index type.                                                                         | {'M': 4, 'efConstruction': 200} (assumes HNSW index) |
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+
+Such filters can be constructed using the [filter expression language defined by Milvus](https://milvus.io/docs/v2.1.x/boolean.md).
+Filters operate on the `tags` of a Document, which are stored as `columns` in the Milvus database.
+
+
+### Example of `.find` with filtered vector search
+
+
+Consider Documents with embeddings `[0,0,0]` up to ` [9,9,9]` where the Document with embedding `[i,i,i]`
+has as tag `price` with value `i`. We can create such example with the following code:
```
NicholasDunham, line 242:
```suggestion
has as tag `price` with value `i`. We can create such an example with the following code:
```

"create this example" would also work

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+
+Points with "price" at most 7:
+
+	embedding=[6. 6. 6.],	 price=6
+	embedding=[7. 7. 7.],	 price=7
+	embedding=[1. 1. 1.],	 price=1
+	embedding=[2. 2. 2.],	 price=2
+```
+
+## Advancded options
```
NicholasDunham, line 361:
```suggestion
## Advanced options
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+(milvus-limitations)=
+## Known limitations of the Milvus Document Store
+
+The Milvus Document Store implements the entire DocumentArray API, but there are some limitations that you should be aware of.
+
+(milvus-collection-loading)=
+### Collection loading
+
+In Milvus, every search or query operation requires the index to be loaded into memory.
+This includes simple Document access through DocArray,
```
NicholasDunham, line 426:
```suggestion
This includes simple Document access through DocArray.
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+
+doc = Document(tensor=np.random.rand(224, 224))
+doc.embed(...)
+doc.tensor = None
+
+da.append(doc)
+```
+
+````{dropdown} Why does this limitation exist?
+By default, DocArray stores three columns in any Document Store: The Document id's, the Document embeddings and
```
NicholasDunham, line 501:
```suggestion
By default, DocArray stores three columns in any Document Store: The Document ids, the Document embeddings and
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+a serialized (Base64 encoded) representation of the Document itself.
+
+In Milvus, the the serialized Document are stored in a column of type 'VARCHAR', which imposes a limit of allowed length
```
NicholasDunham, line 504:
```suggestion
In Milvus, the the serialized Documents are stored in a column of type 'VARCHAR', which imposes a limit of allowed length
```

NicholasDunham (changes requested):

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+
+Such filters can be constructed using the [filter expression language defined by Milvus](https://milvus.io/docs/v2.1.x/boolean.md).
+Filters operate on the `tags` of a Document, which are stored as `columns` in the Milvus database.
+
+
+### Example of `.find` with filtered vector search
+
+
+Consider Documents with embeddings `[0,0,0]` up to ` [9,9,9]` where the Document with embedding `[i,i,i]`
+has as tag `price` with value `i`. We can create such an example with the following code:
```
NicholasDunham, line 242:
```suggestion
has a tag `price` with value `i`. We can create such an example with the following code:
```

NicholasDunham (changes requested):
One very minor change; other than that, LGTM

NicholasDunham (approved):

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+                self._collection = coll
+                self._loaded_when_enter = False
+                self._connection_alias = connection_alias
+
+            def __enter__(self):
+                self._loaded_when_enter = (
+                    loading_progress(
+                        self._collection.name, using=self._connection_alias
+                    )['loading_progress']
+                    != '0%'
```
JoanFM, line 309: this does not look good at all

JoanFM (changes requested):

docarray/array/storage/milvus/backend.py
JohannesMessner: Yeah, I hate it. Unfortunately it seems to be the only way that the Milvus API exposes.

docarray/array/mixins/find.py
```diff
@@ -96,7 +96,7 @@ def find(
         limit: Optional[Union[int, float]] = 20,
         metric_name: Optional[str] = None,
         exclude_self: bool = False,
-        filter: Optional[Dict] = None,
+        filter: Optional[Union[Dict, str]] = None,
```
samsja, line 99:
```suggestion
        filter: Union[Dict, str, None] = None,
```

Not sure if this is not a bit better like this

docarray/array/storage/base/seqlike.py
```diff
@@ -14,13 +14,13 @@ def _update_subindices_append_extend(self, value):
                 if len(docs_selector) > 0:
                     da.extend(docs_selector)
 
-    def insert(self, index: int, value: 'Document'):
+    def insert(self, index: int, value: 'Document', **kwargs):
```
samsja, line 17: lets add the kwargs to the docstring as well

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+from docarray.array.storage.base.backend import BaseBackendMixin, TypeMap
+from docarray.helper import dataclass_from_dict, _safe_cast_int
+
+if TYPE_CHECKING:
+    from docarray.typing import (
+        DocumentArraySourceType,
+    )
+
+
+def always_true_expr(primary_key: str) -> str:
```
samsja, line 29: I would make these helper function private

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+    Returns a Milvus expression that is always true, thus allowing for the retrieval of all entries in a Collection
+    Assumes that the primary key is of type DataType.VARCHAR
+
+    :param primary_key: the name of the primary key
+    :return: a Milvus expression that is always true for that primary key
+    """
+    return f'({primary_key} in ["1"]) or ({primary_key} not in ["1"])'
+
+
+def ids_to_milvus_expr(ids):
```
samsja, line 40: same here I would say private

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+    ids = ['"' + _id + '"' for _id in ids]
+    return '[' + ','.join(ids) + ']'
+
+
+def _sanitize_collection_name(name):
```
samsja, line 45: can we add a docstring :) ?

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+                self.append(_docs)
+
+    def _create_or_reuse_collection(self):
+        if has_collection(self._config.collection_name, using=self._connection_alias):
+            return Collection(
+                self._config.collection_name, using=self._connection_alias
+            )
+
+        document_id = FieldSchema(
+            name='document_id', dtype=DataType.VARCHAR, max_length=1024, is_primary=True
```
samsja, line 133: lets make this constant (1024) class method in UPPERCASE

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+        embedding = FieldSchema(
+            name='embedding', dtype=DataType.FLOAT_VECTOR, dim=self._config.n_dim
+        )
+        serialized = FieldSchema(
+            name='serialized', dtype=DataType.VARCHAR, max_length=65_535
```
samsja, line 139: same here class method UPPERCASE

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+
+        additional_columns = []
+        for col, coltype in self._config.columns.items():
+            mapped_type = self._map_type(coltype)
+            if mapped_type == DataType.VARCHAR:
+                field_ = FieldSchema(name=col, dtype=mapped_type, max_length=1024)
```
samsja, line 146: same here class method UPPERCASE

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+        if has_collection(
+            self._config.collection_name + '_offset2id', using=self._connection_alias
+        ):
+            return Collection(
+                self._config.collection_name + '_offset2id',
+                using=self._connection_alias,
+            )
+
+        document_id = FieldSchema(
+            name='document_id', dtype=DataType.VARCHAR, max_length=1024
```
samsja, line 180: same here class method UPPERCASE

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+        offset = FieldSchema(
+            name='offset', dtype=DataType.VARCHAR, max_length=1024, is_primary=True
```
samsja, line 183: same here class method UPPERCASE

samsja (changes requested):
minor comments

docarray/array/mixins/find.py
JoanFM: I have always preferred the Optional

docarray/array/storage/milvus/backend.py
Jackmin801: Gigachad ask us to shame this :P

JoanFM: Can u explain what it tries to do?

docarray/array/mixins/find.py
samsja: I prefer it as well for normal use but for Optional[Union[]] I am not sure

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+        # table and load the given `docs`
+        if _docs is None:
+            return
+        elif isinstance(_docs, Iterable):
+            self.clear()
+            self.extend(_docs)
+        else:
+            self.clear()
+            if isinstance(_docs, Document):
+                self.append(_docs)
```
alaeddine-13, line 138:
```suggestion
        self.clear()
        if isinstance(_docs, Iterable):
            self.extend(_docs)
        else:
            if isinstance(_docs, Document):
                self.append(_docs)
```

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+            mapped_type = self._map_type(coltype)
+            if mapped_type == DataType.VARCHAR:
+                field_ = FieldSchema(name=col, dtype=mapped_type, max_length=1024)
+            else:
+                field_ = FieldSchema(name=col, dtype=mapped_type)
+            additional_columns.append(field_)
+
+        schema = CollectionSchema(
+            fields=[document_id, embedding, serialized, *additional_columns],
+            description='DocumentArray collection',
```
alaeddine-13, line 172: maybe we should add this as a parameter

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+        return Collection(
```
alaeddine-13, line 174: Does this constructor actually build the collection in the database ?

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+            [doc.to_base64(**self._config.serialize_config) for doc in docs],
+            *extra_columns,
+        ]
+
+    @staticmethod
+    def _docs_from_query_response(response):
+        return DocumentArray([Document.from_base64(d['serialized']) for d in response])
+
+    @staticmethod
+    def _docs_from_search_response(
```
alaeddine-13, line 256: let's move to find mixin

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+    def _map_embedding(self, embedding):
+        if embedding is not None:
+            from docarray.math.ndarray import to_numpy_array
+
+            embedding = to_numpy_array(embedding)
+
+            if embedding.ndim > 1:
+                embedding = np.asarray(embedding).squeeze()
+        else:
+            embedding = np.zeros(self._config.n_dim)
```
alaeddine-13, line 286: I think for some distances (maybe cosine), this needs epsilon added or to be random, can we check ?

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def __getstate__(self):
+        d = dict(self.__dict__)
+        del d['_collection']
+        del d['_offset2id_collection']
+        return d
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+        self._collection = self._create_or_reuse_collection()
```
alaeddine-13, line 297: what about
```python
connections.connect(
            alias=self._connection_alias, host=config.host, port=config.port
        )
```
don't we need to perform it here ?

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+    def loaded_collection(self, collection=None):
+        """
+        Context manager to load a collection and release it after the context is exited.
+        If the collection is already loaded when entering, it will not be released while exiting.
+
+        :param collection: the collection to load. If None, the collection of this indexer is used.
+        :return: Context manager for the provided collection.
+        """
+
+        class LoadedCollectionMngr:
```
alaeddine-13, line 320: is this a name we have to follow ?
if not, let's rename to `LoadedCollectionManager`

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+        self._collection.load()
+        self._offset2id_collection.load()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._collection.release()
+        self._offset2id_collection.release()
+        super().__exit__(exc_type, exc_val, exc_tb)
+
+    def loaded_collection(self, collection=None):
```
alaeddine-13, line 311: it looks like this is used in our implemented operations + loading/releasing is offered using DA context manager.
Is the user supposed to use this in cases other than the DA context manager ?
If there are no such cases, I suggest we make this private

docarray/array/storage/milvus/backend.py
JohannesMessner: So this context manager loads the collection into memory and releases it at the end, and it's used in `find()` and a bunch of other methods.

The `with da:` context manager also performs this function, so that users can bunch together multiple operations while loading/releasing only once.

For that to work, this `loaded_collection` manager has to detect during `__enter__` if the collection is already loaded. In this case it should not be released at exit, because the `with da` manager takes care of that.

Unfortunately, `collection.is_loaded()` is not a thing that exists in pymilvus, and this seems to be the only way to detect if a collection is loaded or not.
The reasoning behind the `!= '0%'` check (as opposed to `== '100%'`) is that any status between 0 and 100 means that some entity higher up in the stack is messing with loading behaviour, and thus we should let it handle releasing as well.

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+        """
+        Context manager to load a collection and release it after the context is exited.
+        If the collection is already loaded when entering, it will not be released while exiting.
+
+        :param collection: the collection to load. If None, the collection of this indexer is used.
```
JohannesMessner, line 316:
```suggestion
        :param collection: the collection to load. If None, the main collection of this indexer is used.
```

docarray/array/document.py
```diff
@@ -139,6 +141,16 @@ def __new__(
...
         ...
 
+    @overload
+    def __new__(
+        cls,
+        _docs: Optional['DocumentArraySourceType'] = None,
+        storage: str = 'milvus',
+        config: Optional[Union['MilvusConfig', Dict]] = None,
+    ) -> 'DocumentArrayMilvus':
+        """Create a Redis-powered DocumentArray object."""
```
JohannesMessner, line 151:
```suggestion
        """Create a Milvus-powered DocumentArray object."""
```

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,342 @@
...
+        ]
+
+    @staticmethod
+    def _docs_from_query_response(response):
+        return DocumentArray([Document.from_base64(d['serialized']) for d in response])
+
+    @staticmethod
+    def _docs_from_search_response(
+        responses,
+    ) -> 'Union[List[DocumentArray], DocumentArray]':
```
alaeddine-13, line 258: I think autoboxing is done in parent Find mixin. In backends the return type is always List[DocumentArray]

docarray/array/storage/milvus/find.py
```diff
@@ -0,0 +1,56 @@
...
+
+class FindMixin:
+    def _find(
+        self,
+        query: 'MilvusArrayType',
+        limit: int = 10,
+        filter: Optional[Dict] = None,
+        param=None,
+        **kwargs
+    ) -> Union['DocumentArray', List['DocumentArray']]:
```
alaeddine-13, line 27: the return type is always List['DocumentArray'] in the backend level, the parent Find mixin will do auto boxing

docarray/array/storage/milvus/getsetdel.py
```diff
@@ -0,0 +1,93 @@
...
+                expr=f'document_id in {_ids_to_milvus_expr(ids)}',
+                output_fields=['serialized'],
+                **kwargs,
+            )
+        if not res:
+            raise KeyError(f'No documents found for ids {ids}')
+        docs = self._docs_from_query_response(res)
+        # sort output docs according to input id sorting
+        ids_list = list(ids)
+        return DocumentArray(sorted(docs, key=lambda d: ids_list.index(d.id)))
```
alaeddine-13, line 65: imo this is not efficient (O(n²log(n))).
2 efficient methods:
* modify the query so that the database returns docs in the specified order, not sure this is possible though
* make 1 loop over docs and make it a map of ids to docs, then make a loop over ids list and do results_da.append(map[doc_id])

docarray/array/storage/milvus/getsetdel.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def _del_docs_by_ids(self, ids: 'Iterable[str]', **kwargs) -> 'DocumentArray':
+        kwargs = self._update_consistency_level(**kwargs)
+        self._collection.delete(
+            expr=f'document_id in {_ids_to_milvus_expr(ids)}', **kwargs
+        )
```
alaeddine-13, line 71: respect batch size

docarray/array/storage/milvus/getsetdel.py
```diff
@@ -0,0 +1,93 @@
...
+    ):
+        # delete old entries
+        kwargs = self._update_consistency_level(**kwargs)
+        self._collection.delete(
+            expr=f'document_id in {_ids_to_milvus_expr(ids)}',
+            **kwargs,
+        )
+        # insert new entries
+        payload = self._docs_to_milvus_payload(docs)
+        self._collection.insert(payload, **kwargs)
```
alaeddine-13, line 84: respect batch size

docarray/array/storage/milvus/seqlike.py
```diff
@@ -0,0 +1,55 @@
...
+        self._set_doc_by_id(value.id, value, **kwargs)
+        self._offset2ids.append(value.id)
+
+    def _extend(self, values: Iterable['Document'], **kwargs):
+        docs = list(values)
+        if not docs:
+            return
+        kwargs = self._update_consistency_level(**kwargs)
+        payload = self._docs_to_milvus_payload(docs)
+        self._collection.insert(payload, **kwargs)
```
alaeddine-13, line 54: batch size

docarray/array/storage/milvus/backend.py
JohannesMessner: The DA context manager also handled offset2id, and could potentially do other stuff in the future. I think it can be useful to expose this if the user wants to make sure that nothing other than loading/releasing is happening.

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+    environment:
+      - ETCD_AUTO_COMPACTION_MODE=revision
+      - ETCD_AUTO_COMPACTION_RETENTION=1000
+      - ETCD_QUOTA_BACKEND_BYTES=4294967296
+      - ETCD_SNAPSHOT_COUNT=50000
+    volumes:
+      - ${DOCKER_VOLUME_DIRECTORY:-.}/volumes/etcd:/etcd
+    command: etcd -advertise-client-urls=http://127.0.0.1:2379 -listen-client-urls http://0.0.0.0:2379 --data-dir /etcd
+
+  minio:
```
alaeddine-13, line 35: I'm afraid this  can get stale at some point (although we are actually pinning container versions).
It would be really great if we find this yaml in the official versioned documentation and point to it

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+
+`````
+
+Then
+
+```bash
+docker-compose up
+```
+
+You can find more installation guidance in the [Milvus documentation](https://milvus.io/docs/v2.1.x/install_standalone-docker.md).
```
alaeddine-13, line 79: exactly the link provided in this page, maybe instead of putting the whole yaml file, we just link to https://github.com/milvus-io/milvus/releases/download/v2.1.4/milvus-standalone-docker-compose.yml

docarray/array/storage/milvus/find.py
```diff
@@ -0,0 +1,56 @@
...
+    def _filter(self, filter, limit=10, **kwargs):
+        kwargs = self._update_consistency_level(**kwargs)
+        with self.loaded_collection():
+            results = self._collection.query(
+                expr=filter,
+                limit=limit,
+                output_fields=['serialized'],
+                **kwargs,
+            )
+        return self._docs_from_query_response(results)[:limit]
```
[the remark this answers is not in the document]
JohannesMessner: yes the backend does not seem to apply a limit for filter-only queries, I am not sure that exposing it will be a good experience.

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+| `host`              | Hostname of the Milvus server                                                                                                                                                                                                                 | 'localhost'                                          |
+| `port`              | Port of the Milvus server                                                                                                                                                                                                                     | 6333                                                 |
+| `distance`          | Distance metric to be used during search. Can be 'IP', 'L2', 'JACCARD', 'TANIMOTO', 'HAMMING', 'SUPERSTRUCTURE' or 'SUBSTRUCTURE'.                                                                                                            | 'IP' (inner product)                                 |
```
alaeddine-13, line 126: let's also point to the supported distances in the official docs

docarray/array/storage/milvus/find.py
JohannesMessner: I was going by the type hints provided in the backend implementation docs, we need to change it there then:
image

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+| `serialize_config`  | [Serialization config of each Document](../../../fundamentals/document/serialization.md)                                                                                                                                                      | {}                                                   |
+ | `consistency_level` | [Consistency level](https://milvus.io/docs/v2.1.x/consistency.md#Consistency-levels) for Milvus database operations. Can be 'Session', 'Strong', 'Bounded' or 'Eventually'.                                                                   | 'Session'                                            |
+| `columns`           | Additional columns to be stored in the datbase, taken from Document `tags`.                                                                                                                                                                   | None                                                 |
+
+## Minimal example
+
+Create `docker-compose.yml`:
+
+`````{dropdown} docker-compose.yml
```
alaeddine-13, line 138: isn't this already covered ?
If we need a command maybe just this: 
```
wget https://github.com/milvus-io/milvus/releases/download/v2.1.4/milvus-standalone-docker-compose.yml -O docker-compose.yml
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+
+from docarray import DocumentArray
+
+N, D = 5, 128
+
+da = DocumentArray.empty(
+    N, storage='milvus', config={'n_dim': D, 'distance': 'IP'}
+)  # init
+with da:
+    da.embeddings = np.random.random([N, D])
```
alaeddine-13, line 215: I don't think this interface is the first thing users should see.
I believe append/extend should be the first one

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+    )
+
+print('\nIndexed Prices:\n')
+for embedding, price in zip(da.embeddings, da[:, 'tags__price']):
+    print(f'\tembedding={embedding},\t price={price}')
+```
+
+Consider we want the nearest vectors to the embedding `[8. 8. 8.]`, with the restriction that
+prices must follow a filter. As an example, retrieved Documents must have `price` value lower than
+or equal to `max_price`. We can encode this information in Milvus using `filter = f'price <= {max_price}'`.
```
alaeddine-13, line 273: I'd avoid the word encode

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+        ]
+    )
+
+print('\nIndexed Prices:\n')
+for embedding, price in zip(da.embeddings, da[:, 'tags__price']):
+    print(f'\tembedding={embedding},\t price={price}')
+```
+
+Suppose you want to filter results such that
+retrieved Documents must have a `price` value less than or equal to `max_price`. You can encode
```
alaeddine-13, line 334: same here for encode

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,512 @@
...
+```
+
+
+(milvus-limitations)=
+## Known limitations of the Milvus Document Store
+
+The Milvus Document Store implements the entire DocumentArray API, but there are some limitations that you should be aware of.
+
+(milvus-collection-loading)=
+### Collection loading
```
alaeddine-13, line 423: can we add a reference from milvus docs to consolidate this fact ?

tests/conftest.py
```diff
@@ -36,10 +34,43 @@ def start_storage():
...
+    while not es.ping():
+        time.sleep(0.5)
+
+
+def _wait_for_milvus():
+    from pymilvus import connections, has_collection
+    from pymilvus.exceptions import MilvusUnavailableException
+
+    milvus_conn_alias = f'pytest_localhost_19530'
+    connections.connect(alias=milvus_conn_alias, host='localhost', port=19530)
```
alaeddine-13, line 50: is the alias needed at all here?

tests/conftest.py
```diff
@@ -36,10 +34,43 @@ def start_storage():
...
+
+
+@pytest.fixture
+def milvus_cleanup():
+    yield
+    from pymilvus import list_collections, drop_collection
+
+    alias = f'docarray_localhost_19530'  # assumes default host and port are used
+    for c in list_collections(using=alias):
+        drop_collection(c, using=alias)
```
alaeddine-13, line 76: I wouldn't care about the clean up, the tester should just remove the docker container + collections between tests should have different names (unless, keeping all collections in the same testing session hugely impacts the test performance)

docarray/array/storage/milvus/backend.py
JoanFM: I would use a simple context manager, I think is great value to have the simplest the better

tests/conftest.py
JohannesMessner: yup, unfortunately

tests/conftest.py
JohannesMessner: You are right, this is not used anymore anyways. will remove

JohannesMessner:
@alaeddine-13 I agree with all your comments regarding the documentation. However, I modeled this after the documentation of the other backends, and I think keeping it consistent with them is more important in this case. I think there is a lot of value in people being to able to do the same setup steps etc. with all of the backends.

docarray/array/storage/milvus/backend.py
alaeddine-13: then let's document this as part of the public interface in the Collection loading section

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+    host: str = 'localhost'
+    port: Optional[Union[str, int]] = 19530  # 19530 for gRPC, 9091 for HTTP
+    distance: str = 'IP'  # metric_type in milvus
+    index_type: str = 'HNSW'
+    index_params: Dict = field(
+        default_factory=lambda: {
+            'M': 4,
+            'efConstruction': 200,
+        }
+    )  # passed to milvus at index creation time. The default assumes 'HNSW' index type
```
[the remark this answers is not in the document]
JohannesMessner: yes, these are required

docarray/array/storage/milvus/backend.py
JohannesMessner: We wait for the loading to be completed, so no error should arise from this.

docarray/array/storage/milvus/backend.py
JohannesMessner: I would say no, what we are describing here is the schema, not the collection itself. And the schema is not something that the user influences.
All parameters for the collection itself are exposed in `collection_config`.

docarray/array/storage/milvus/backend.py
JohannesMessner: yes, that's what it looks like from checking their code

docarray/array/storage/milvus/getsetdel.py
```diff
@@ -0,0 +1,93 @@
...
+        # delete old entries
+        self._clear_offset2ids_milvus()
+        # insert current entries
+        ids = self._offset2ids.ids
+        if not ids:
+            return
+        offsets = [str(i) for i in range(len(ids))]
+        dummy_vectors = [np.zeros(1) for _ in range(len(ids))]
+        collection = self._offset2id_collection
+        collection.insert([offsets, ids, dummy_vectors])
```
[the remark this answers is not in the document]
JohannesMessner: When implementing this initially I though there was more dynamic stuff going on with offset2id, hence this approach. I'll see if I can change it.

docarray/array/storage/milvus/getsetdel.py
JohannesMessner: Realistically I am not sure if such an in-Python implementation would actually be faster than `sorted()` which is implemented in C.
If the list length goes to infinity then obviously yes, but this is constrained by the number of documents that can fit into memory, so the length here will never be gigantic.

docarray/array/storage/milvus/getsetdel.py
JohannesMessner: as far as I can tell, milvus does not take a batch size for these operations

docarray/array/storage/milvus/getsetdel.py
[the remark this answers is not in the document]
JohannesMessner: Storing as vector does not work because it needs to be fixed-length; serializing and storing as VarChar would work, but with a very large DocumentArray we might run into the VarChar max length restriction. I would prefer it to keep it this way, just so we don't have to worry about it, especially since this is not a method that is expected to be called very often.

docarray/array/storage/milvus/backend.py
```diff
@@ -0,0 +1,320 @@
...
+            raise ValueError('Empty config is not allowed for Milvus storage')
+        elif isinstance(config, dict):
+            config = dataclass_from_dict(MilvusConfig, config)
+
+        if config.collection_name is None:
+            id = uuid.uuid4().hex
+            config.collection_name = 'docarray__' + id
+        self._config = config
+        self._config.columns = self._normalize_columns(self._config.columns)
+
```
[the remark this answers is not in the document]
JohannesMessner: I am not sure what you mean, I have not found such a check in qdrant

docarray/array/storage/milvus/backend.py
JohannesMessner: I'd rather keep it next to `_docs_from_query_response()` and that one is used in multiple places, so I'll keep it here.

docarray/array/storage/milvus/backend.py
JohannesMessner: this is not an issue

JohannesMessner:
Test is failing after merging master, it is passing on my local. Working on it now

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,495 @@
+(milvus)=
+# Milvus
+
+One can use [Milvus](https://milvus.io/) as the Document store for DocumentArray. It is useful when one wants to have faster Document retrieval on embeddings, i.e. `.match()`, `.find()`.
+
+````{tip}
+This feature requires `pymilvus`. You can install it via `pip install "docarray[milvus]".`
```
alaeddine-13, line 7:
```suggestion
This feature requires `pymilvus`. You can install it via `pip install "docarray[milvus]"`. 
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,495 @@
...
+
+```python
+from docarray import DocumentArray
+
+da = DocumentArray(storage='milvus', config={'n_dim': 10})
+```
+
+Here, `config` is configuration for the new Milvus collection,
+and `n_dim` is a mandatory field that specifies the dimensionality of stored embeddings.
+For more information about the Milvus `config`, refer to the {ref}`specification <milvus-config>`.
```
alaeddine-13, line 94: @alexcg1 pointed out this style guide before: https://developers.google.com/style/link-text
```suggestion
For more information about the Milvus `config`, refer to the {ref}`config <milvus-config>`.
```

docs/advanced/document-store/milvus.md
```diff
@@ -0,0 +1,495 @@
...
+```
+
+Currently, dynamically setting a consistency level is supported for the following operations:
+`.append()`, `.extend()`, and `.insert()`.
+
+### Passing search parameters
+
+In Milvus you can [pass parameters to the search operation](https://milvus.io/docs/v2.1.x/search.md#Conduct-a-vector-search) which [depend on the used index type](https://milvus.io/docs/v2.1.x/index.md).
+
+In DocumentArray, this ability is exposed through the `param` argument in the `.find()` method:
```
alaeddine-13, line 372:
```suggestion
In DocumentArray, this ability is exposed through the `param` argument in the `~docarray.array.mixins.find` method:
```

JohannesMessner:
I need to update the PR to play nice with the new mode that disables offset2id

alaeddine-13 (approved):

hanxiao:
image
This should be wrapped with ```console

samsja (approved):
LGTM what a big PR :smiley:

PR #590, opened 2022-10-06
chore: housekeeping

### LINKED ISSUE #580: Bug: document stores disappeared in api reference
Despite being listed in the `.rst`, most document stores do not appear in the table; only in memory and sqlite are visible

Image

Image

hanxiao:
fix: better error message on bad selector
docs(dataclass): add uri type
docs: fix #580 
feat(dataclass): add uri type


JoanFM:
Highlight also `URI` to ease the Release Notes

PR #602, opened 2022-10-10
feat: add filter for qdrant

dongxiang123:
Goals:

As the Qdrant backend, there is no filter supported, the PR is to support Qdrant backend with a filter feature in Docarray supported.  
- ...

### DIFF
--- a/docarray/array/storage/redis/backend.py
+++ b/docarray/array/storage/redis/backend.py
@@ -8,7 +8,7 @@ from docarray.array.storage.base.backend import BaseBackendMixin, TypeMap
 from docarray.helper import dataclass_from_dict, random_identity, filter_dict
 
 from redis import Redis
-from redis.commands.search.field import NumericField, TextField, VectorField
+from redis.commands.search.field import NumericField, TextField, VectorField, GeoField
 from redis.commands.search.indexDefinition import IndexDefinition
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


JoanFM (changes requested):

JoanFM (changes requested):
Where is this feature tested?

setup.py
```diff
@@ -65,7 +65,7 @@
             'strawberry-graphql',
         ],
         'qdrant': [
-            'qdrant-client~=0.7.3',
+            'qdrant-client>=0.7.3',
```
JoanFM, line 68: fix it to 0.8.0

JoanFM (changes requested):

setup.py
alaeddine-13: @JoanFM I think a better approach is to allow least compatible version and if user attempts to use a feature, he is responsible that his qdrant client and qdrant backend support it.
However @dongxiang123 yesterday we discovered that we need to introduce an upper limit for thr version, let's find it and put it

JohannesMessner: We found that there is an icompatability with `qdrant-client==0.8.0`, so for now we have to keep it as is, otherwise it breaks - or we bump to 0.8.0 and also fix this issue.
With v0.7.3 it still seems possible to use qdrant v8, but I haven't done exhaustive testing on that. @dongxiang123 will find out ;)

Issue #529, opened 2022-09-07
`ModuleNotFoundError` when using elasticsearch as storage backend


violenil:
Code examples from [this](https://docarray.jina.ai/advanced/document-store/elasticsearch/) documentation with elasticsearch as storage result in `ModuleNotFoundError`. 

Minimal example:

```
from docarray import DocumentArray, Document

da = DocumentArray(
    storage='elasticsearch',
    config={'index_name': 'new_index', 'n_dim': 128},
)

with da:
    da.extend([Document() for _ in range(1000)])

da.summary()
```

Example code runs through and displays `.summary()`:

```
╭─────────────── Documents Summary ───────────────╮
│                                                 │
│   Type                   DocumentArrayElastic   │
│   Length                 1000                   │
│   Homogenous Documents   True                   │
│   Common Attributes      ('id',)                │
│   Multimodal dataclass   False                  │
│                                                 │
╰─────────────────────────────────────────────────╯
╭───────────────────── Attributes Summary ─────────────────────╮
│                                                              │
│   Attribute   Data type   #Unique values   Has empty value   │
│  ──────────────────────────────────────────────────────────  │
│   id          ('str',)    1000             False             │
│                                                              │
╰──────────────────────────────────────────────────────────────╯
╭──────── DocumentArrayElastic Config ────────╮
│                                             │
│   n_dim             128                     │
│   distance          cosine                  │
│   hosts             http://localhost:9200   │
│   index_name        new_index             │
│   es_config         {}                      │
│   index_text        False                   │
│   tag_indices       []                      │
│   batch_size        64                      │
│   ef_construction   None                    │
│   m                 None                    │
│   columns           []                      │
│                                             │
╰─────────────────────────────────────────────╯
```

But then raises `ModuleNotFoundError`:

```
Exception ignored in: <function BaseGetSetDelMixin.__del__ at 0x7fa010267e50>
Traceback (most recent call last):
  File "/Users/jinadev/opt/anaconda3/envs/venv-now/lib/python3.9/site-packages/docarray/array/storage/base/getsetdel.py", line 330, in __del__
  File "/Users/jinadev/opt/anaconda3/envs/venv-now/lib/python3.9/site-packages/docarray/array/storage/elastic/getsetdel.py", line 128, in _save_offset2ids
  File "/Users/jinadev/opt/anaconda3/envs/venv-now/lib/python3.9/site-packages/docarray/array/storage/elastic/backend.py", line 217, in _update_offset2ids_meta
[...]

PR #604, opened 2022-10-11
fix(array): make docarrays persist atexit

Jackmin801:
Goals:

- [x] Fix issue mentioned in https://github.com/jina-ai/docarray/issues/529
- [x] Add fix to base so it propagates to other backends
- [ ] Make sure other backends aren't broken

### DIFF
--- a/tests/unit/array/test_sequence.py
+++ b/tests/unit/array/test_sequence.py
@@ -1,3 +1,5 @@
+import gc
+import sys
 import tempfile
 import uuid
 
@@ -260,3 +262,37 @@ def test_set_and_append(index, storage, config):
         da.append(Document(id='new_new'))
 
     assert da[:, 'id'] == ['0', 'new', '2', '3', '4', 'new_new']
+
+
+@pytest.mark.parametrize(
+    'da_cls,config',
+    [
+        (DocumentArrayInMemory, lambda: None),
+        (DocumentArraySqlite, lambda: None),
+        # DocArrays backed by Weaviate in 0.17.0 has a refcount bug
+        # Remove this comment when the bug is fixed
+        # (DocumentArrayWeaviate, lambda: WeaviateConfig(n_dim=3)),
+        (DocumentArrayQdrant, lambda: QdrantConfig(n_dim=3)),
+        (DocumentArrayElastic, lambda: ElasticConfig(n_dim=3)),
+        (DocumentArrayRedis, lambda: RedisConfig(n_dim=3)),
+    ],
+)
+def test_deletion(da_cls, config, start_storage, capsys):
+    # Mock Deletion
+    class MockDA(da_cls):
+        def __del__(self):
+            assert not self._deleted
+            print('Goodbye')
+            self._deleted = True
+
+    da = MockDA(config=config())
+    assert not da._deleted
+    assert (
+        sys.getrefcount(da) == 2
+    ), f'{da_cls.__name__} initialized with {sys.getrefcount(da) - 1} references'
+
+    del da
+    gc.collect()
+    output = capsys.readouterr()
+    assert output.out == "Goodbye\n"
+    assert output.err == ""
--- a/docarray/array/base.py
+++ b/docarray/array/base.py
@@ -1,3 +1,5 @@
+import atexit
+from weakref import WeakMethod
 from typing import MutableSequence, TYPE_CHECKING, Union, Iterable
 
 from docarray import Document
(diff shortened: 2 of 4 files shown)


Jackmin801:
Of the new tests added, 1 of them is expected to fail.

DocArrays backed by Weaviate seem to have 2 references to them upon creation.
This means they cannot be garbage collected when they go out of scope, possibly creating a memory leak

docarray/array/storage/base/backend.py
```diff
@@ -21,6 +23,17 @@ def _init_storage(
...
+
+    def _register_exit_hook(self):
+        weakref_del = WeakMethod(self.__del__)
+
+        def exit_hook():
+            if weakref_del() is not None:
+                weakref_del()()
+
+        atexit.register(exit_hook)
+        self._deleted = False
```
JoanFM, line 36: not a fan of this state? why is it needed?

JoanFM: why not handle this in the context manager?

JoanFM (changes requested):

docarray/array/storage/base/backend.py
Jackmin801: Hrmm it's so that it doesn't try to persist the DA twice on SystemExit (program exit).

When SystemExit occurs, the Python Interpreter will first go through all the **hooks** registered by the **atexit** module.
Variables then start getting destroyed which will trigger their **dunder del destructor**.

Using the **dunder del won't work** for our case because the modules our destructors depend on may have already been deallocated by the Python Interpreter at the time the destructor is called in dunder del.
Thus, we want it such that during SystemExit, the **atexit hook** is the method that is used to persist the DA because it will be evaluated at a time when the dependencies are still alive.

But the dunder del will still trigger after the **atexit hook** is complete. So we need a way to make the destructor singleton (it only triggers once, subsequent calls will do nothing).
The proposal at the moment is to implemented the singleton by having this self._deleted flag that tracks whether or not the destructor has been called.

There are of course other ways to implement the singleton destructor behavior so I'm open to suggestions on how to go about this

JoanFM, line 36: Okey, then initialize `self._deleted = False` at `__init__`

JoanFM (changes requested):

JoanFM:
Hey @Jackmin801 , please solve the commit linting problem to have the `tests` run

Jackmin801:
Ah ok. I purposely failed it cause I wasnt sure if I should be constantly running the CI. Seems like they are a bit expensive

docarray/array/storage/base/backend.py
Jackmin801: Done. It does make it that the `__del__` and this logic get further apart in the codebase though. Previously they were in the same folder. Now they are 2 hierarchies apart

tests/unit/array/test_sequence.py
```diff
@@ -260,3 +262,37 @@ def test_set_and_append(index, storage, config):
...
+            self._deleted = True
+
+    da = MockDA(config=config())
+    assert not da._deleted
+    assert (
+        sys.getrefcount(da) == 2
+    ), f'{da_cls.__name__} initialized with {sys.getrefcount(da) - 1} references'
+
+    del da
+    gc.collect()
```
JoanFM, line 295: I am not sure this is guarantee of anything

JoanFM: I do not like having a test relying on this

JoanFM (changes requested):

tests/unit/array/test_sequence.py
Jackmin801: But I don't think there is a nice way to test the module exceptions. What we are looking for is a way to capture behavior of a SystemExit.

I have thought of some janky ways to do it though.

One such janky way is to have the test spawn a subprocess and assert that the cleanup of the subprocess doesn't raise any errors. It's kind of hacky so I'm still looking for a nicer way. But this might be the only way to go about it.

Jackmin801: We could also assert that the DocArray created by the subprocess can be loaded correctly.

But yea, using subprocess in tests seems a bit hacky

alaeddine-13 (commented):
for this logic, we will be update offset2id twice if we use the context manager.
What I propose is:
1. rename _delete to _synced
2. _synced is initially true
3. _synced = False on append/extend/update/delete operations
4. _synced = True when context manager exits
I also think it's clearer to put just _save_offset2id (or any logic that cannot run during program exit) in the exit hook, not all of __del__.

@JoanFM what do you think ?

JoanFM:
> 4\. I also think it's clearer to put just _save_offset2id (or any logic that cannot run during program exit) in the exit hook, not all of **del**.

I like this idea

Jackmin801:
There are four candidate places we can sync:
1. `[self.sync]` 		When the user explicitly calls da.sync()
2. `[self.__exit__]` 	When the user exits the DocArray’s context manager
3. `[self.__del__]`	        When the variable gets garbage collected
4. `[atexit.register]`	  When the program exits

DocArray will only support (1) and (2) -- explicit sync and sync when exiting context manager.

We will not support syncing when the variable gets garbage collected or when the program exits.
Because:
1. Not supporting them would make it such that we do not need to keep track of a _synced flag. Which simplifies the logic.
2. They are too implicit and unpredictable. Using explicit sync and the pythonic context manager are intentional and have easy to understand, predictable behavior. Garbage collection and trying to sync on program exit can have edge cases that can make them tricky for users to use and would breed bad practice in their code.

Jackmin801:
We are no longer pursuing adding the behavior of persistence to atexit. The intended persistence behavior will be implemented in https://github.com/jina-ai/docarray/pull/625

PR #606, opened 2022-10-11
feat: add config for pca in annlite

jemmyshin:
Goals: Enable loading indexer with PCA in DocumentArray AnnLite backend.

In AnnLite we load the model by their `params_hash` which compute by hashing `n_dim`, `n_components`... 
So in order to successfully load the dumped indexer, we need to pass the `n_components` to AnnLite.
 
- add a new parameter `n_components`, which is the output dimension of PCA, in `annlite/backend.py` AnnliteConfig

### DIFF
--- a/docarray/array/storage/annlite/backend.py
+++ b/docarray/array/storage/annlite/backend.py
@@ -27,6 +27,7 @@ class AnnliteConfig:
     ef_construction: Optional[int] = None
     ef_search: Optional[int] = None
     max_connection: Optional[int] = None
+    n_components: Optional[int] = None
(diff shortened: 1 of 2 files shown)


JoanFM (changes requested):
what about the `annlite` version? is it compatible?

alaeddine-13 (commented):
can you make sure it's documented here ?
https://docarray.jina.ai/advanced/document-store/annlite/#config
also double check if we miss other params

jemmyshin:
> 

No need to change the version, we already have this `n_components` in `AnnLite` long time ago

JoanFM (approved):

Issue #605, opened 2022-10-11
Figure it out how to manage multiple dependency that handle different modality


samsja:
# Context

We need to figure it out how to manage multiple dependency that handle different modality.

The idea could be that we stick to one external dependency per modality. Ex, pillow for Image, librosa for audio etcc.

But it is not settle, this ticket is to have this discussion.

JoanFM:
To be handled in v2

PR #607, opened 2022-10-11
feat(audio): allow to load from uri with a custom sample rate

samsja:
# Context

It might be useful to load an audio with a given sample rate. This is for example needed for the [WhisperExecutor](https://github.com/samsja/audio2image/tree/main/executors/whisper)

Atm the only way to use this executor properly is to do 

```python
from docarray import Document
import librosa

d = Document(uri='audio.wav')
d.tensor = librosa.load(d.uri, sr=16_000)[0]
```

# What this pr do

Allow to load with a sample rate, the above example would become

```python
from docarray import Document
d = Document(uri='audio.wav').load_uri_to_audio_tensor(sample_rate = 16_000)
```
This feel more docarray-ic

This PR need to use librosa as dependency therefore can only be merged once we figure it out what do to with different dependency for multi modality see this [issue](https://github.com/jina-ai/docarray/issues/605)

Moreover using librosa as a potiential backend could allow more audio features, like reading for mp3 etcc

### DIFF
--- a/docarray/document/mixins/audio.py
+++ b/docarray/document/mixins/audio.py
@@ -1,5 +1,5 @@
 import wave
-from typing import Union, BinaryIO, TYPE_CHECKING
+from typing import Union, BinaryIO, TYPE_CHECKING, Optional
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


docarray/document/mixins/audio.py
```diff
@@ -68,3 +78,16 @@ def load_uri_to_audio_tensor(self: 'T') -> 'T':
...
+    def _load_uri_to_audio_tensor_librosa(
+        self: 'T', sample_rate: Optional[float] = None
+    ) -> 'T':
+        """Convert an audio :attr:`.uri` into :attr:`.tensor` inplace
+        :param sample_rate: the audio sample rate to load the files
+        :return: Document itself after processed
+        """
+        import librosa
+
+        self.tensor, self.tags['sample_rate'] = librosa.load(self.uri, sr=sample_rate)
```
JoanFM, line 91: why is this important, `sample_rate` just put into `self.tags["sample_rate"]`?

JoanFM (changes requested):

samsja:
Not in main concern anymore

Issue #528, opened 2022-09-07
Documentation, clarify the use of da as context manager (`with` statement) when da is backed by vector database


samsja:
# Context

Most of the time ppl should do 
```python
with da:
    ...
```

when working with vector database DA.

The issue is that the documentation is not clear when and why they should do it.

This need to be clarified

samsja:
branch merged : https://github.com/jina-ai/docarray/pull/613

PR #610, opened 2022-10-11
[Old] docs(array): add section on persistence mutation and context manager

Jackmin801:
Goals:

- [x] Address documentation gap mentioned in https://github.com/jina-ai/docarray/issues/528

### DIFF
--- a/docs/advanced/document-store/index.md
+++ b/docs/advanced/document-store/index.md
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
        [4., 4., 4.]])
 ```
 
+## Persistence, mutations and context manager
+
+Having DocumentArrays that are backed by an external store does however introduce an extra consideration into the way one thinks about DocumentArrays.
+The DocumentArray object created in our Python program is now a view of the underlying implementation in the external store.
+This means that our DocumentArray object in Python can be out of sync with what is persisted to the external store.
+
+**For example**
+```python
+from docarray import DocumentArray, Document
+
+da1 = DocumentArray(storage='redis', config=dict(n_dim=3, index_name="my_index"))
+da1.append(Document())
+print(f"Length of da1 is {len(da1)}")
+
+da2 = DocumentArray(storage='redis', config=dict(n_dim=3, index_name="my_index"))
+print(f"Length of da2 is {len(da2)}")
+```
+**Output**
+```console
+Length of da1 is 1
+Length of da2 is 0
+```
+
+Executing this script multiple times yields the same result.
+
+When we ran the line `da1.append(Document())`, we expect the DocumentArray with `index_name="my_index"` to now have a length of `1`.
+However, when we try to create another view of the DocumentArray in `da2`, we get a fresh DocumentArray.
+
+We also expect the script to increment the length of the DocumentArrays every time we run it.
+This is because the previous run should have saved the length of the DocumentArray with `index_name="my_index"` and our most recent run will append a new document, incrementing the length by `+1` each time.
+
+However, it seems like our append operation is also not being persisted.
+
+````{dropdown} What actually happened here?
+The DocumentArray actually did persist.
+But not in the way we might expect.
+Since we did not use the `with` context manager or scope our mutation, the persistence logic is being evaluated when the program exits.
+`da1` is destroyed first, persisting the DocumentArray of length `1`.
+But when `da2` is destroyed, it persists a DocumentArray of length `0` to the same index in Redis as `da1`, overriding it's value.
+
+This means that if we had not created `da2`, the overriding would not have occured and the script will actually increment the length of the DocumentArray correctly.
+You can prove this to yourself by commenting out the last 2 lines of the script and running the script repeatedly.
+
+**Script**
+```python
+from docarray import DocumentArray, Document
+
+da1 = DocumentArray(storage='redis', config=dict(n_dim=3, index_name="my_index"))
+da1.append(Document())
+print(f"Length of da1 is {len(da1)}")
+
+# da2 = DocumentArray(storage='redis', config=dict(n_dim=3, index_name="my_index"))
+# print(f"Length of da2 is {len(da2)}")
+```
+
+**First run output**
+```console
+Length of da1 is 1
+```
+**Second run output**
+```console
+Length of da1 is 2
+```
+**Third run output**
+```console
+Length of da1 is 3
+```
+````
+
+Now that we know the issue, let's explore some idioms we can use to work with DocumentArrays backed by external storage in a more predictable manner
+### Using Context Manager
+The recommended way is to use the DocumentArray as a context manager like so:
+
+```python
+from docarray import DocumentArray, Document
+
+da1 = DocumentArray(storage='redis', config=dict(n_dim=3, index_name="my_index"))
+with da1:  # Let's use the context manager to make sure we persist the mutation
+    da1.append(Document())  #
+print(f"Length of da1 is {len(da1)}")
+
+da2 = DocumentArray(storage='redis', config=dict(n_dim=3, index_name="my_index"))
+print(f"Length of da2 is {len(da2)}")
+```
+**First run output**
+```console
+Length of da1 is 1
+Length of da2 is 1
+```
+**Second run output**
+```console
+Length of da1 is 2
+Length of da2 is 2
+```
+**Third run output**
+```console
+Length of da1 is 3
+Length of da2 is 3
+```
+
+The append we made to the DocumentArray is now persisted properly. Hurray!
+
+### Using Scope
+Another method that is sometimes useful when multiple DocumentArrays are involed is to perform the mutation in a function scope.
+The DocumentArrays will always persist when they fall out of scope.
+
+```python
+from docarray import DocumentArray, Document
+
+# Let's wrap our mutation logic in a function scope
+def foo():
+    da1 = DocumentArray(storage='redis', config=dict(n_dim=3, index_name="my_index"))
+    da1.append(Document())
+    print(f"Length of da1 is {len(da1)}")
+
+
+foo()
+
+da2 = DocumentArray(storage='redis', config=dict(n_dim=3, index_name="my_index"))
+print(f"Length of da2 is {len(da2)}")
+```
+**First run output**
+```console
+Length of da1 is 1
+Length of da2 is 1
+```
+**Second run output**
+```console
+Length of da1 is 2
+Length of da2 is 2
+```
+**Third run output**
+```console
+Length of da1 is 3
+Length of da2 is 3
+```
+
+However, one needs to be extra cautious when opting for this method.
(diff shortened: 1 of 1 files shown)


JoanFM (commented):
Have we checked if the Examples and code snippets do follow these considerations?

docs/advanced/document-store/index.md
```diff
@@ -15,11 +15,11 @@ benchmark
 ```
 
 Documents inside a DocumentArray can live in a [document store](https://en.wikipedia.org/wiki/Document-oriented_database) instead of in memory, e.g. in SQLite, Redis.
-Comparing to the in-memory storage, the benefit of using an external store is often about longer persistence and faster retrieval. 
+The benefit of using an external store over an in-memory store is often about longer persistence and faster retrieval. 
 
 The look-and-feel of a DocumentArray with external store is **almost the same** as a regular in-memory DocumentArray. This allows users to easily switch between backends under the same DocArray idiom.  
 
-Take SQLite as an example, using it as the store backend of a DocumentArray is as simple as follows:
+Take SQLite as an example, using it as the storage backend of a DocumentArray is as simple as follows:
```
NicholasDunham, line 22:
```suggestion
Take SQLite as an example. Using it as the storage backend of a DocumentArray is as simple as follows:
```

docs/advanced/document-store/index.md
```diff
@@ -58,19 +58,19 @@ da.summary()
...
-Note that  `da` was modified inside a `with` statement. This context manager ensures that the the `DocumentArray` indices,
+Note that `da` was modified inside a `with` statement. This context manager ensures that the the `DocumentArray` indices,
 which allow users to access the  `DocumentArray` by position (allowing statements such as `da[1]`),
 are properly mapped and saved to the storage backend.
 This is the recommended default usage to modify a DocumentArray that lives on a document store to avoid
 unexpected behaviors that can yield to, for example, inaccessible elements by position.
 
 
-Creating, retrieving, updating, deleting Documents are identical to the regular {ref}`DocumentArray<documentarray>`. All DocumentArray methods such as `.summary()`, `.embed()`, `.plot_embeddings()` should work out of the box.
+Creating, retrieving, updating, and deleting Documents are identical to that of a regular {ref}`DocumentArray<documentarray>`. All DocumentArray methods such as `.summary()`, `.embed()`, `.plot_embeddings()` should also work out of the box.
```
NicholasDunham, line 68:
```suggestion
The procedures for creating, retrieving, updating, and deleting Documents are identical to those for a regular {ref}`DocumentArray<documentarray>`. All DocumentArray methods such as `.summary()`, `.embed()`, `.plot_embeddings()` should also work out of the box.
```

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+```
+**Output**
+```console
+Length of da1 is 1
+Length of da2 is 0
+```
+
+Executing this script multiple times yields the same result.
+
+When we ran the line `da1.append(Document())`, we expect the DocumentArray with `index_name="my_index"` to now have a length of `1`.
```
NicholasDunham, line 374:
```suggestion
When we run the line `da1.append(Document())`, we expect the DocumentArray with `index_name="my_index"` to now have a length of `1`.
```

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+However, when we try to create another view of the DocumentArray in `da2`, we get a fresh DocumentArray.
+
+We also expect the script to increment the length of the DocumentArrays every time we run it.
+This is because the previous run should have saved the length of the DocumentArray with `index_name="my_index"` and our most recent run will append a new document, incrementing the length by `+1` each time.
+
+However, it seems like our append operation is also not being persisted.
+
+````{dropdown} What actually happened here?
+The DocumentArray actually did persist.
```
NicholasDunham, line 383:
```suggestion
The DocumentArray actually did persist,
```

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+But not in the way we might expect.
```
NicholasDunham, line 384:
```suggestion
but not in the way we might expect.
```

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+Since we did not use the `with` context manager or scope our mutation, the persistence logic is being evaluated when the program exits.
+`da1` is destroyed first, persisting the DocumentArray of length `1`.
+But when `da2` is destroyed, it persists a DocumentArray of length `0` to the same index in Redis as `da1`, overriding it's value.
```
NicholasDunham, line 387:
```suggestion
But when `da2` is destroyed, it persists a DocumentArray of length `0` to the same index in Redis as `da1`, overriding its value.
```

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+
+This means that if we had not created `da2`, the overriding would not have occured and the script will actually increment the length of the DocumentArray correctly.
```
NicholasDunham, line 389:
```suggestion
This means that if we had not created `da2`, the overriding would not have occured and the script would actually increment the length of the DocumentArray correctly.
```

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+```console
+Length of da1 is 2
+```
+**Third run output**
+```console
+Length of da1 is 3
+```
+````
+
+Now that we know the issue, let's explore some idioms we can use to work with DocumentArrays backed by external storage in a more predictable manner
```
NicholasDunham, line 418:
```suggestion
Now that we know the issue, let's explore some idioms we can use to work with DocumentArrays backed by external storage in a more predictable manner.
```

NicholasDunham (changes requested):
Some minor grammatical changes

Jackmin801:
@JoanFM The examples given are correct to the latest release of docarray.

However, the considerations are not true for all backends. Some persist even when not using the context manager. Though the correctness of these backends will be unaffected by following the suggestions.

If this is the intended consideration for using DocArrays backed by vector databases I think it's best to document it that way rather than go through all the nuance of which implementations may or may not require context manager. Which we don't define and can't guarantee will be consistent in future releases.

Jackmin801:
Thanks @NicholasDunham 👍🏻

JohannesMessner:
> Basically. By enforcing these considerations, we are telling users that if they perform any mutations, using context manager or scopes is guaranteed to have the expected behaviour of persistence.
> 
> Outside of this, we cannot guarantee anything. So users should probably not do any mutations outside of context manager or scope.

I agree with this philosophy for the reason than @Jackmin801, and also for a slightly different one: If we tell users to always use a context manager they will have a much easier time switching between storage backends.

Additionally I think we should come up with a soft guideline for backend developers, do we generally prefer eager or lazy persistence?

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
        [4., 4., 4.]])
 ```
 
+## Persistence, mutations and context manager
+
+Having DocumentArrays that are backed by an external store does however introduce an extra consideration into the way one thinks about DocumentArrays.
```
JohannesMessner, line 351:
```suggestion
Having DocumentArrays that are backed by a document store introduces an extra consideration into the way you think about DocumentArrays.
```

Small thing, but let's not assume that people have read the section above. They might have come from an outside link, or just scanned over the rest bc they very looking specifically for this section.

Also, we generally prefer `you` over `one` or `we`.

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+The DocumentArray object created in our Python program is now a view of the underlying implementation in the external store.
+This means that our DocumentArray object in Python can be out of sync with what is persisted to the external store.
```
JohannesMessner, line 353:
```suggestion
The DocumentArray object created in your Python program is now a view of the underlying implementation in the external store.
This means that your DocumentArray object in Python can be out of sync with what is persisted to the external store.
```

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+```
+**Output**
+```console
+Length of da1 is 1
+Length of da2 is 0
+```
+
+Executing this script multiple times yields the same result.
+
+When we run the line `da1.append(Document())`, we expect the DocumentArray with `index_name="my_index"` to now have a length of `1`.
```
JohannesMessner, line 374:
```suggestion
When we run the line `da1.append(Document())`, we expect the DocumentArray with `index_name='my_index'` to now have a length of `1`.
```

docs/advanced/document-store/index.md
```diff
@@ -346,6 +346,179 @@ array([[7., 7., 7.],
...
+**Third run output**
+```console
+Length of da1 is 3
+Length of da2 is 3
+```
+
+The append we made to the DocumentArray is now persisted properly. Hurray!
+
+### Using Scope
+Another method that is sometimes useful when multiple DocumentArrays are involed is to perform the mutation in a function scope.
```
JohannesMessner, line 452: Honestly I don't think that we should even tell users about this.
Using the context manager seems safer and more predictable, and it would be our single (official) way of doing things, which is always good.

JohannesMessner (changes requested):
Nice explanation, really clear!

As you will find out over time I am quite picky about documentation, so I left a few things that I would change ;)

Also please change `we` to `you` everywhere, we want to be consistent about this. For reference: https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md#-contributing-documentation

docs/advanced/document-store/index.md
samsja: More than that actually I am not sure that the behavior will persist indefinitely (with future releases). And users should never do that so I would agree not to tell about that part.

Jackmin801: Yup agree. Seems like bad practice so probably shouldn't recommend it though it works in the current release.

Jackmin801: Should we not support it then? I think @alaeddine-13 is proposing we don't have the persistence logic in `__del__` at all in his [latest comment on the SystemExit behavior PR](https://github.com/jina-ai/docarray/pull/604#pullrequestreview-1138612018)

JoanFM: No function scope at all, this should be removed. Plus in Python u do not know when they will be released

Jackmin801:
Merging the [mirror of this branch in Jina repo](https://github.com/jina-ai/docarray/pull/613) is better. Closing this PR

Issue #591, opened 2022-10-06
Bug `optional` and default `none` does not create an empty Document holder


hanxiao:
Bug, as discussed with @samsja @JohannesMessner @alaeddine-13 

```python
from typing import Optional

from docarray import dataclass, Document
from docarray.typing import Text


@dataclass
class WebDoc:
    foo: Text
    bar: Optional[Text] = None


d = Document(WebDoc(foo='hello'))
d.bar.text = 'world' # wont work

assert len(d.chunks) == 2  # fail
assert isinstance(d.bar, Document)  # fail
```

samsja:
will be taken care of later

PR #616, opened 2022-10-12
fix: allow optional str int etcc in dataclass

samsja:
# Context 

Original issue : https://github.com/jina-ai/docarray/issues/591

Optional field and docarray dataclass does not play well

Indeed for instance one would except that if I create a Document out of a dataclass with an `Optional[Text]` field with default value `None` that the Document will contain an empty Document in bar (added to the chunk). Atm These chunk Document is not created as shown in this code

```python
from typing import Optional

from docarray import dataclass, Document
from docarray.typing import Text


@dataclass
class WebDoc:
    foo: Text
    bar: Optional[Text] = None


d = Document(WebDoc(foo='hello'))
d.bar.text = 'world' # wont work

assert len(d.chunks) == 2  # fail
assert isinstance(d.bar, Document)  # fail
>>>Traceback (most recent call last):
  File "/home/sami/trash/test.py", line 16, in <module>
    d = Document(WebDoc(foo='hello'))
  File "/home/sami/Documents/workspace/Jina/docarray/docarray/document/__init__.py", line 147, in __init__
    super().__init__(*args, **kwargs)
  File "/home/sami/Documents/workspace/Jina/docarray/docarray/base.py", line 38, in __init__
    self._data = type(self)._from_dataclass(_obj)._data
  File "/home/sami/Documents/workspace/Jina/docarray/docarray/document/mixins/multimodal.py", line 116, in _from_dataclass
    raise TypeError(
TypeError: Unsupported type annotation on field `None`
```

# What this pr do

it allows to do
```python


  @dataclass
  class MultiModalDoc:
      foo: Optional[Text] = None

  d = Document(MultiModalDoc())

  d.foo.text = 'world'  
```
This mean it will create a chunk empty Document. Before it was doing nothing

This work as well with non docarray type.
```python
@dataclass
class MultiModalDoc:
    foo: Optional[str] = None

d = Document(MultiModalDoc())
d.tags['foo'] = 'blabla'
```

### DIFF
--- a/docarray/document/mixins/multimodal.py
+++ b/docarray/document/mixins/multimodal.py
@@ -1,7 +1,15 @@
 import base64
 import typing
+from typing import Optional
 
-from docarray.dataclasses.types import Field, is_multimodal, _is_field
+from docarray.dataclasses.types import (
+    Field,
+    is_multimodal,
+    _is_field,
+    _is_optional,
+    _is_optional_field,
+    _get_optional_subtype,
+)
 from docarray.dataclasses.types import AttributeType
 
 if typing.TYPE_CHECKING:
@@ -31,8 +39,6 @@ class MultiModalMixin:
         multi_modal_schema = {}
         for key, field in obj.__dataclass_fields__.items():
             attribute = getattr(obj, key)
-            if attribute is None:
-                continue
 
             if field.type in [str, int, float, bool] and not _is_field(field):
                 tags[key] = attribute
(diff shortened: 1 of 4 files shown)


JoanFM:
The explanation of this PR does not make sense, what is bar and what is foo? can u clarify what is the wrong behavior and what is expected after this?

Also this seems a bug and therefore should be a fix:

tests/unit/array/test_dataclass.py
```diff
@@ -0,0 +1,60 @@
...
+
+def test_dataclass_opregetional_value():
+    @dataclass
+    class MultiModalDoc:
+        foo: Text
+
+    d = Document(MultiModalDoc(foo='sfdssf'))
+
+
+def test_dataclass_optional_value():
```
JoanFM, line 18: why `bar`? where does this bar come from?

JoanFM (changes requested):

tests/unit/array/test_dataclass.py
samsja: my bad it should be foo. This test is not passing atm anyway

JoanFM: adjust PR description

docarray/document/mixins/multimodal.py
```diff
@@ -41,12 +46,48 @@ def _from_dataclass(cls, obj) -> 'Document':
...
+
+            elif field.type in [bytes, Optional[bytes]] and not _is_field(field):
+                tags[key] = (
+                    base64.b64encode(attribute).decode() if attribute else attribute
+                )
+                multi_modal_schema[key] = {
+                    'attribute_type': AttributeType.PRIMITIVE,
+                    'type': field.type.__name__
+                    if not (_is_optional(field.type))
+                    else f'Optional[{_get_optional_subtype(field.type).__name__}]',
```
samsja, line 69: Here I am open to suggestion should I just remove the Optional in the type name ?

samsja:
Question, should we allow the same behavior with List ?
```python
@dataclass
class WebDoc:
    foo: List[Optional[Text]] = [None] 
```
or 
```python
@dataclass
class WebDoc:
    foo: Optional[List[Text]] = None
```

JoanFM:
Are we 100% this is the expected behavior? I am not sure if I would expect it not to create the `Optional` document but do it lazily perhaps?

For instance:

```python
from typing import Optional

from docarray import dataclass, Document
from docarray.typing import Text


@dataclass
class WebDoc:
    foo: Text
    bar: Optional[Text] = None


d = Document(WebDoc(foo='hello'))
d.bar.text = 'world' # wont work

assert len(d.chunks) == 2  # fail
assert isinstance(d.bar, Document)  # fail
```

Are we sure that you would expect `d.chunks` to be 2 even before creating the `d.bar`? 

Wouldn't this be more expected?

```python
from typing import Optional

from docarray import dataclass, Document
from docarray.typing import Text


@dataclass
class WebDoc:
    foo: Text
    bar: Optional[Text] = None



d = Document(WebDoc(foo='hello'))

len(d.chunks) to be 1
d.bar not existing?

d.bar.text = 'world' # wont work

assert len(d.chunks) == 2  #SUCCEED
assert isinstance(d.bar, Document)   #SUCCEED

```

I am however not sure this is even possible

samsja:
@JoanFM I am not sure neither.

My though is that when you add a field with the default value, the default value will be passed to the Document at init.

```python
@dataclass
class WebDoc:
    foo: Text = 'foo'

d = Document(WebDoc())
```

under the hood this will just call `Document(text='foo')`

Here I am following the exact same idea and under the hood if the default value is None it will call `Document(text=None)`

JoanFM:
> My though is that when you add a field with the default value, the default value will be passed to the Document at init.

Okey, make sense. Make sure to document clearly

PR #617, opened 2022-10-12
feat: add support for labeled dataset to evaluate function

guenthermi:
Goals:

- Allow the user to add labels to the documents instead of passing an addtional ground truth `DocumentArray`

### DIFF
--- a/docarray/array/mixins/evaluation.py
+++ b/docarray/array/mixins/evaluation.py
@@ -9,9 +9,38 @@ if TYPE_CHECKING:
     from docarray import Document, DocumentArray
 
 
+def _evaluate_deprecation(f):
+    """Raises a deprecation warning if the user executes the evaluate function with
+    the old interface and adjust the input to fit the new interface."""
+
+    def func(*args, **kwargs):
+        if len(args) > 1:
+            if not (isinstance(args[1], Callable) or isinstance(args[1], str)):
+                kwargs['ground_truth'] = args[1]
+                args = [args[0]] + list(args[2:])
+                warnings.warn(
+                    'The `other` attribute in `evaluate()` is transfered from a '
+                    'positional attribute into the keyword attribute `ground_truth`.'
+                    'Using it as a positional attribute is deprecated and will be removed '
+                    'in the next version.',
+                    DeprecationWarning,
+                )
+        if 'other' in kwargs:
+            kwargs['ground_truth'] = kwargs['other']
+            warnings.warn(
+                '`other` is renamed to `groundtruth` in `evaluate()`, the usage of `other` is '
+                'deprecated and will be removed in the next version.',
+                DeprecationWarning,
+            )
+        return f(*args, **kwargs)
+
+    return func
+
+
 class EvaluationMixin:
     """A mixin that provides ranking evaluation functionality to DocumentArrayLike objects"""
 
+    @_evaluate_deprecation
     def evaluate(
         self,
         metric: Union[str, Callable[..., float]],
@@ -19,18 +48,20 @@ class EvaluationMixin:
         hash_fn: Optional[Callable[['Document'], str]] = None,
         metric_name: Optional[str] = None,
         strict: bool = True,
-        label_tag='label',
+        label_tag: Optional[str] = 'label',
         **kwargs,
     ) -> Optional[float]:
         """
         Compute ranking evaluation metrics for a given `DocumentArray` when compared
         with a groundtruth.
 
-        This implementation expects the documents and their matches to have labels
-        annotated inside the tag with the key specified in the `label_tag` attribute.
-        Alternatively, one can provide a `ground_truth` DocumentArray that is
-        structurally identical to `self`. In this case, this function compares the
-        `matches` of `documents` inside the `DocumentArray`.
+        If one provides a `ground_truth` DocumentArray that is structurally identical
+        to `self`, this function compares the `matches` of `documents` inside the
+        `DocumentArray` to this `ground_truth`.
+        Alternatively, one can directly annotate the documents by adding labels in the
+        form of tags with the key specified in the `label_tag` attribute.
+        Those tags need to be added to `self` as well as to the documents in the
+        matches properties.
 
         This method will fill the `evaluations` field of Documents inside this
         `DocumentArray` and will return the average of the computations
@@ -46,6 +77,7 @@ class EvaluationMixin:
         :param strict: If set, then left and right sides are required to be fully
             aligned: on the length, and on the semantic of length. These are preventing
             you to evaluate on irrelevant matches accidentally.
+        :param label_tag: Specifies the tag which contains the labels.
         :param kwargs: Additional keyword arguments to be passed to `metric_fn`
         :return: The average evaluation computed or a list of them if multiple metrics
             are required
@@ -57,7 +89,7 @@ class EvaluationMixin:
         elif label_tag in self[0].tags:
             if ground_truth:
                 warnings.warn(
-                    'An ground_truth attribute is provided but does not '
+                    'A ground_truth attribute is provided but does not '
                     'contain matches. The labels are used instead and '
                     'ground_truth is ignored.'
                 )
@@ -117,7 +149,10 @@ class EvaluationMixin:
                     1 if m.tags[label_tag] == d.tags[label_tag] else 0 for m in targets
                 ]
             else:
-                raise RuntimeError(f'Unsupported groundtruth type {ground_truth_type}')
+                raise RuntimeError(
+                    'Could not identify which kind of ground truth'
+                    'information is provided to evaluate the matches.'
+                )
 
             r = metric_fn(binary_relevance, max_rel=max_rel, **kwargs)
             d.evaluations[metric_name] = NamedScore(
--- a/docarray/array/mixins/evaluation.py
+++ b/docarray/array/mixins/evaluation.py
@@ -48,7 +48,7 @@ class EvaluationMixin:
         hash_fn: Optional[Callable[['Document'], str]] = None,
         metric_name: Optional[str] = None,
         strict: bool = True,
-        label_tag: Optional[str] = 'label',
+        label_tag: str = 'label',
         **kwargs,
     ) -> Optional[float]:
         """
--- a/docarray/array/mixins/evaluation.py
+++ b/docarray/array/mixins/evaluation.py
@@ -22,14 +22,14 @@ def _evaluate_deprecation(f):
                     'The `other` attribute in `evaluate()` is transfered from a '
                     'positional attribute into the keyword attribute `ground_truth`.'
                     'Using it as a positional attribute is deprecated and will be removed '
-                    'in the next version.',
+                    'soon.',
                     DeprecationWarning,
                 )
         if 'other' in kwargs:
             kwargs['ground_truth'] = kwargs['other']
             warnings.warn(
                 '`other` is renamed to `groundtruth` in `evaluate()`, the usage of `other` is '
-                'deprecated and will be removed in the next version.',
+                'deprecated and will be removed soon.',
                 DeprecationWarning,
             )
         return f(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


guenthermi:
The Wikipedia article linked in the `r_precision` [docstring](https://github.com/jina-ai/docarray/blob/944b91b3f20301ceb5c79690122e9bf292286e3f/docarray/math/evaluation.py#L14-L28) does not seems to be correct. According to the article the metric is calculated based on the Top-N results, where N is determined by the number of relevant documents in the dataset. However, the implementation sets N as the last position of a relevant document as done [here](https://gist.github.com/bwhite/3726239#file-rank_metrics-py-L40-L66)

JoanFM (changes requested):

docarray/array/mixins/evaluation.py
```diff
@@ -14,31 +14,64 @@ class EvaluationMixin:
 
     def evaluate(
         self,
-        other: 'DocumentArray',
         metric: Union[str, Callable[..., float]],
+        ground_truth: Optional['DocumentArray'] = None,
```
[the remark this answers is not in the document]
guenthermi: It is a breaking change anyway, because we need to turn it from a mandatory into an optional attribute, but I can change it back to `other` if you think it makes a difference. I just find this name confusion, especially as an keyword argument.

JoanFM, line 18: please avoid breaking the interface, `ground_truth` can be added as an alias, u can have a `deprecation` decorator that has `other` equivalent to `groundtruth`

docarray/array/mixins/evaluation.py
```diff
@@ -14,31 +14,64 @@ class EvaluationMixin:
...
-        """Compute ranking evaluation metrics for a given `DocumentArray` when compared with a groundtruth.
+        """
+        Compute ranking evaluation metrics for a given `DocumentArray` when compared
+        with a groundtruth.
 
-        This implementation expects to provide a `groundtruth` DocumentArray that is structurally identical to `self`. It is based
-        on comparing the `matches` of `documents` inside the `DocumentArray.
+        This implementation expects the documents and their matches to have labels
+        annotated inside the tag with the key specified in the `label_tag` attribute.
+        Alternatively, one can provide a `ground_truth` DocumentArray that is
```
JoanFM, line 31: This in my opinion should be the principal case, and the other the alternative

docarray/array/mixins/evaluation.py
```diff
@@ -14,31 +14,64 @@ class EvaluationMixin:
 
     def evaluate(
         self,
-        other: 'DocumentArray',
         metric: Union[str, Callable[..., float]],
+        ground_truth: Optional['DocumentArray'] = None,
         hash_fn: Optional[Callable[['Document'], str]] = None,
         metric_name: Optional[str] = None,
         strict: bool = True,
+        label_tag='label',
```
JoanFM, line 22: label_tag should be an Optional, plus please provide `type hints`

docarray/array/mixins/evaluation.py
```diff
@@ -14,31 +14,64 @@ class EvaluationMixin:
...
+            are required
         """
+        if len(self) == 0:
+            raise ValueError('It is not possible to evaluate an empty DocumentArray')
+        if ground_truth and len(ground_truth) > 0 and ground_truth[0].matches:
+            ground_truth_type = 'matches'
+        elif label_tag in self[0].tags:
+            if ground_truth:
+                warnings.warn(
+                    'An ground_truth attribute is provided but does not '
```
JoanFM, line 60:
```suggestion
                    'A ground_truth attribute is provided but does not '
```

JoanFM (changes requested):

docarray/array/mixins/evaluation.py
guenthermi: ok, yes I can change this in the docstring

docarray/array/mixins/evaluation.py
guenthermi: The problem is, that the function has atm two mandatory attributes `other` and `metric` (in this order). It does not make sense to keep `other` mandatory if the DocumentArray and its matches itself have labels. So I have to make it optional. If I do this I can not preserve the order since `metric` as a mandatory attribute has to be placed before `other`.

guenthermi: The only way how it might work is to make metric optional as well by setting a default metric.

JoanFM: other and ground_truth need to be equivalent, this 100% sure

JoanFM: this can be achieved by a decorator

docarray/array/mixins/evaluation.py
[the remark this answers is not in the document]
guenthermi: Yes and optional is also missing.

docarray/array/mixins/evaluation.py
```diff
@@ -53,7 +86,7 @@ def evaluate(
         metric_name = metric_name or metric_fn.__name__
         results = []
         caller_max_rel = kwargs.pop('max_rel', None)
-        for d, gd in zip(self, other):
+        for d, gd in zip(self, ground_truth):
```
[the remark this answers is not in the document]
guenthermi: This is done if `strict=True`

docarray/array/mixins/evaluation.py
guenthermi: @bwanglzu , the function in docarray.math is called `precision_at_k`, setting a `k` is possible but not mandatory since the matches usually contain only the Top-K

guenthermi: Just for documentation: we agreed on checking the type of the first argument. If it is a `DocumentArray` we set ground_truth to metric and raise a deprecation warning. We also handle cases where someone provides a other keyword argument and raise a deprecation warning.

docarray/array/mixins/evaluation.py
```diff
@@ -68,14 +101,23 @@ def evaluate(
...
+                    )
+                binary_relevance = [
+                    1 if hash_fn(m) in desired else 0 for m in d.matches
+                ]
+            elif ground_truth_type == 'labels':
+                binary_relevance = [
+                    1 if m.tags[label_tag] == d.tags[label_tag] else 0 for m in targets
+                ]
+            else:
+                raise RuntimeError(f'Unsupported groundtruth type {ground_truth_type}')
```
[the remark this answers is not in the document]
guenthermi: It should be unreachable code, but probably it is better to just remove it

docarray/array/mixins/evaluation.py
bwanglzu: oh i remembered

docarray/array/mixins/evaluation.py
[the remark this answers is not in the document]
samsja: I would say better to keep it but just have a better error message. Smth like 
```suggestion
                raise RuntimeError(f'Something went wrong with the groudtruth')
```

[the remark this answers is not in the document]
JoanFM: Describe what can be wrong, without talking about the internal name, but something the user can relate to

LMMilliken (approved):

tests/unit/array/mixins/test_eval_class.py
```diff
@@ -36,13 +36,120 @@ def test_eval_mixin_perfect_match(metric_fn, kwargs, storage, config, start_stor
...
+def test_eval_mixin_perfect_match_labeled(
+    metric_fn, kwargs, storage, config, start_storage
+):
+    da1 = DocumentArray.empty(10)
+    for d in da1:
+        d.tags = {'label': 'A'}
+    da1.embeddings = np.random.random([10, 256])
+    da1_index = DocumentArray(da1, storage=storage, config=config)
+    for d in da1_index:
+        d.tags = {'label': 'A'}
```
bwanglzu, line 80: is this still needed?

docarray/array/mixins/evaluation.py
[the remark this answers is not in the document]
JoanFM: it is not breaking, passing from mandatory to Optional will not break other's code, it just adds more options

bwanglzu (commented):
minor comments, make sure da team agree `ground_truth` as a breaking change

tests/unit/array/mixins/test_eval_class.py
```diff
@@ -36,13 +36,120 @@ def test_eval_mixin_perfect_match(metric_fn, kwargs, storage, config, start_stor
...
+    da2 = copy.deepcopy(da1)
+    for d in da2:
+        d.tags = {'label': 'B'}
+    da1_index = DocumentArray(da2, storage=storage, config=config)
+    da1.match(da1_index, exclude_self=True)
+    r = da1.evaluate(metric_fn, **kwargs)
+    assert isinstance(r, float)
+    assert r == 0.0
+    for d in da1:
+        d: Document
```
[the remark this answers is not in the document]
guenthermi: oh I don't know, I will remove it

tests/unit/array/mixins/test_eval_class.py
guenthermi: No, you are right, I think I can remove it.

docarray/array/mixins/evaluation.py
```diff
@@ -9,36 +9,101 @@
...
         :param kwargs: Additional keyword arguments to be passed to `metric_fn`
-        :return: The average evaluation computed or a list of them if multiple metrics are required
+        :return: The average evaluation computed or a list of them if multiple metrics
+            are required
         """
+        if len(self) == 0:
+            raise ValueError('It is not possible to evaluate an empty DocumentArray')
+        if ground_truth and len(ground_truth) > 0 and ground_truth[0].matches:
+            ground_truth_type = 'matches'
+        elif label_tag in self[0].tags:
```
JoanFM, line 89: what happens here if `label_tag` is None? is `label_tag` really Optional?

JoanFM (changes requested):

docarray/array/mixins/evaluation.py
guenthermi: ah sorry, yes it should not be optional

docarray/array/mixins/evaluation.py
```diff
@@ -48,7 +48,7 @@ def evaluate(
         hash_fn: Optional[Callable[['Document'], str]] = None,
         metric_name: Optional[str] = None,
         strict: bool = True,
-        label_tag: Optional[str] = 'label',
+        label_tag: str = 'label',
```
JoanFM, line 51: I'd rather check what happens with `None`. One should be able to put `None`

JoanFM (changes requested):

tests/unit/array/mixins/test_eval_class.py
```diff
@@ -217,10 +327,72 @@ def test_diff_match_len_in_gd(storage, config, metric_fn, start_storage, kwargs)
...
+    ],
+)
+def test_useless_groundtruth_warning_should_raise(storage, config, start_storage):
+    da1 = DocumentArray.empty(10)
+    for d in da1:
+        d.tags = {'label': 'A'}
+    da1.embeddings = np.random.random([10, 256])
+    da1_index = DocumentArray(da1, storage=storage, config=config)
+    for d in da1_index:
+        d.tags = {'label': 'A'}
```
bwanglzu, line 394: minor issue

bwanglzu (approved):
LGTM!

JoanFM (approved):

docarray/array/mixins/evaluation.py
guenthermi: Ok so do you mean, that it should be optional and default to `label`.  If a user does not want to use labels, i.e., pass a `ground_truth` it should be ok to pass `None`. However, if it is set to `None` and no `ground_truth` is passed, an exception is raised.

PR #621, opened 2022-10-13
refactor: change r precision calculation and documentation

guenthermi:
Goals:

- correct the implementation of R-Precision
- add hints in the documentation about potentially incorrect evaluation scores
- add tests for all metric functions
- add links to developer reference of metric functions in the documentation

### DIFF
--- a/docarray/math/evaluation.py
+++ b/docarray/math/evaluation.py
@@ -12,20 +12,25 @@ def _check_k(k):
 
 
 def r_precision(binary_relevance: List[int], **kwargs) -> float:
-    """R Precision after all relevant documents have been retrieved
-    Relevance is binary (nonzero is relevant).
+    """R-Precision determines the precision in the fist R documents, where R is the
+    number of documents relevant to the query.
+
+    Relevance is considered binary by this function (nonzero is relevant).
+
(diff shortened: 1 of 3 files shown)


guenthermi:
tests will be added in #617

docarray/math/evaluation.py
```diff
@@ -12,20 +12,25 @@ def _check_k(k):
 
 
 def r_precision(binary_relevance: List[int], **kwargs) -> float:
```
JoanFM, line 14: there needs to be some tests testing the new behavior

JoanFM (changes requested):

docarray/math/evaluation.py
guenthermi: Yes I agree. Currently there are no tests for the metric functions. I have created a test which tests the behavior together with the evaluate function functions in #617 (which has to be adopted -after one of those Pars is merged), but I can also create tests which directly test the functions docarray.math here.

bwanglzu (commented):
add tests otherwise looks good

docarray/math/evaluation.py
JoanFM: add unit tests yes

JoanFM (approved):

PR #622, opened 2022-10-14
fix(annlite): bump annlite to v0.4.0

jemmyshin:
Goals:
Recently we are going to release `Annlite` v0.4.0. We replace `lmdb` with `rocksdb` to reduce memory footprint.
More details are [here](https://docs.google.com/document/d/1AcOsp-iGLdP3XHeQSa0PMpnjSIGg49RSy1EiDiHVNXU/edit#), green lines in these charts (rocksdb and meta table on disk) are our implementation now.
This pr fixes the errors in `Annlite`-related unit tests introduced by file lock in `rocksdb`.
We will make a formal pr once we confirm that `Annlite` won't break the `docarray` test.

- Manually close `Annlite` in unit test in order to solve the file lock issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,7 +133,7 @@ jobs:
           pip install --no-cache-dir ".[test]"
           pip install --no-cache-dir ".[full]"
           pip install --no-cache-dir ".[qdrant]"
-          pip install --no-cache-dir ".[annlite]"
+          pip install --pre annlite
           pip install --no-cache-dir ".[weaviate]"
           pip install --no-cache-dir ".[elasticsearch]"
           pip install --no-cache-dir ".[redis]"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,10 +130,10 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install wheel
           # pip does not properly resolve dependency versions with syntax pip install --no-cache-dir ".[test,full]"
+          pip install --pre annlite
           pip install --no-cache-dir ".[test]"
           pip install --no-cache-dir ".[full]"
           pip install --no-cache-dir ".[qdrant]"
-          pip install --pre annlite
           pip install --no-cache-dir ".[weaviate]"
           pip install --no-cache-dir ".[elasticsearch]"
           pip install --no-cache-dir ".[redis]"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,7 +131,6 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install wheel
           # pip does not properly resolve dependency versions with syntax pip install --no-cache-dir ".[test,full]"
-          pip install --pre annlite
           pip install --no-cache-dir ".[test]"
           pip install --no-cache-dir ".[full]"
           pip install --no-cache-dir ".[qdrant]"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
               repo: context.repo.repo,
               comment_id: ${{ steps.fc.outputs.comment-id }},
             })
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2.5.0
         with:
           fetch-depth: 0
       - run: 'echo "module.exports = {extends: [''@commitlint/config-conventional'']}" > commitlint.config.js'
@@ -51,9 +51,9 @@ jobs:
   lint-flake-8:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2.5.0
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
       - name: Lint with flake8
@@ -67,11 +67,11 @@ jobs:
   check-black:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2.5.0
         with:
           fetch-depth: 0
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
       - id: file_changes
@@ -84,9 +84,9 @@ jobs:
   import-test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2.5.0
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
       - name: Prepare enviroment
@@ -102,11 +102,12 @@ jobs:
     runs-on: ubuntu-latest
     needs: [commit-lint, lint-flake-8, check-black, import-test]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2.5.0
       - id: set-matrix
         run: |
           sudo apt-get install jq
-          echo "::set-output name=matrix::$(bash scripts/get-all-test-paths.sh)"
+          export value=$(bash scripts/get-all-test-paths.sh)
+          echo "matrix=$value" >> $GITHUB_OUTPUT
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
 
@@ -120,9 +121,9 @@ jobs:
         python-version: [3.7]
         test-path: ${{fromJson(needs.prep-testbed.outputs.matrix)}}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2.5.0
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: Prepare enviroment
@@ -144,7 +145,7 @@ jobs:
         run: |
           pytest --suppress-no-test-exit-code --cov=docarray --cov-report=xml \
             -v -s -m "not gpu" ${{ matrix.test-path }}
-          echo "::set-output name=codecov_flag::docarray"
+          echo "codecov_flag=docarray" >> $GITHUB_OUTPUT
         timeout-minutes: 30
         env:
           JINA_AUTH_TOKEN: "${{ secrets.JINA_AUTH_TOKEN }}"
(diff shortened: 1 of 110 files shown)


JoanFM (changes requested):
Do not merge this. This should be fixed by Annlite or closed when ctxt manager leaves

docarray/array/storage/annlite/backend.py
```diff
@@ -134,3 +134,6 @@ def __setstate__(self, state):
 
     def __len__(self):
         return self._annlite.index_size
+
+    def close(self):
+        self._annlite.close()
```
JoanFM, line 139: I do not think there is a `close` interface in DocArray

tests/unit/array/mixins/test_empty.py
```diff
@@ -24,14 +24,15 @@
...
-
     # Assert .empty provides a da with 0 docs
     if config:
         da = da_cls.empty(config=config)
     else:
         da = da_cls.empty()
 
     assert len(da) == 0
+    if da_cls == DocumentArrayAnnlite:
+        da.close()
```
JoanFM, line 35: remove this, this is not what we are testing, we cannot test custom behavior like this. WE AIM FOR SAME interface

tests/unit/array/mixins/test_io.py
```diff
@@ -45,6 +45,13 @@ def test_document_save_load(
     da = da_cls(docs, config=config())
     da.insert(2, Document(id='new'))
     da.save(tmp_file, file_format=method, encoding=encoding)
+    if da_cls == DocumentArrayAnnlite:
+        da_info = [
+            [d.id for d in da],
+            [d.embedding for d in da],
+            [d.content for d in da],
+        ]
+        da.close()
```
JoanFM, line 54: remove all these exceptions

JoanFM (changes requested):

docarray/array/storage/annlite/backend.py
jemmyshin: Yes, this is only for annlite backend. We need this because the file lock needed to be closed before access it again.

tests/unit/array/mixins/test_empty.py
jemmyshin: Do you think we need to refactor the annlite-based test? We need to manually close it before another test starts. Otherwise, the file lock will block us.

JoanFM: I do not like imposing the `close` API, this needs to be fix on ANNLite backend IMO

samsja:
Maybe instead of having a `close` method we can acquire and release the lock when using the 
```python
with da:
   ...
```
context manager

JoanFM:
> Maybe instead of having a `close` method we can acquire and release the lock when using the
> 
> ```python
> with da:
>    ...
> ```
> 
> context manager

This is a reasonable approach

tests/unit/array/mixins/test_io.py
```diff
@@ -43,20 +44,29 @@ def test_document_save_load(
...
+
     da.insert(2, Document(id='new'))
     da.save(tmp_file, file_format=method, encoding=encoding)
 
+    with da:
+        da_info = {
+            'id': [d.id for d in da],
+            'embedding': [d.embedding for d in da],
+            'content': [d.content for d in da],
+        }
```
numb3r3, line 56:
```suggestion
    with da:
        da.insert(2, Document(id='new'))
        da.save(tmp_file, file_format=method, encoding=encoding)
        da_info = {
            'id': [d.id for d in da],
            'embedding': [d.embedding for d in da],
            'content': [d.content for d in da],
        }
```

tests/unit/array/mixins/test_io.py
```diff
@@ -200,13 +210,23 @@ def test_from_to_pd_dataframe(da_cls, config, start_storage):
...
-    assert len(da_cls.load_binary(bytes(da_cls.empty(2, config=config)))) == 2
+    if da_cls == DocumentArrayAnnlite:
+        b = da_cls.empty(2, config=config)
+        b._annlite.close()
+
+        d = da_cls.from_bytes(b.to_bytes(), config=config)
+        assert len(d) == 2
+        d._annlite.close()
+    else:
+        assert len(da_cls.load_binary(bytes(da_cls.empty(2, config=config)))) == 2
```
numb3r3, line 221:
```suggestion
    with da_cls.empty(2, config=config) as da:
        da_bytes = da.to_bytes()
    
    with da_cls.from_bytes(da_bytes, config=config) as db:
        assert len(db) == len(2)
```

tests/unit/array/mixins/test_io.py
```diff
@@ -200,13 +210,23 @@ def test_from_to_pd_dataframe(da_cls, config, start_storage):
...
 
     da = da_cls.empty(2, config=config)
 
     da[:, 'embedding'] = [[1, 2, 3], [4, 5, 6]]
     da[:, 'tensor'] = [[1, 2], [2, 1]]
     da[0, 'tags'] = {'hello': 'world'}
+    if da_cls == DocumentArrayAnnlite:
+        da._annlite.close()
     da2 = da_cls.load_binary(bytes(da))
```
numb3r3, line 230:
```suggestion
    with da_cls.empty(2, config=config) as da:
        da[:, 'embedding'] = [[1, 2, 3], [4, 5, 6]]
        da[:, 'tensor'] = [[1, 2], [2, 1]]
        da[0, 'tags'] = {'hello': 'world'}
        da_bytes = bytes(da)
     
    with da_cls.load_binary(da_bytes) as da2:
```

numb3r3 (changes requested):
I leave some suggested context usages in the unittest

tests/unit/array/mixins/test_empty.py
```diff
@@ -31,12 +31,14 @@ def test_empty_non_zero(da_cls, config, start_storage):
     else:
         da = da_cls.empty()
 
-    assert len(da) == 0
+    with da:
```
JoanFM, line 34: this should not be needed. For `read` operations this should not be needed

tests/unit/array/mixins/test_io.py
```diff
@@ -200,13 +209,22 @@ def test_from_to_pd_dataframe(da_cls, config, start_storage):
...
+            assert len(db) == 2
+    else:
+        assert len(da_cls.load_binary(bytes(da_cls.empty(2, config=config)))) == 2
 
     da = da_cls.empty(2, config=config)
 
     da[:, 'embedding'] = [[1, 2, 3], [4, 5, 6]]
     da[:, 'tensor'] = [[1, 2], [2, 1]]
     da[0, 'tags'] = {'hello': 'world'}
+    if da_cls == DocumentArrayAnnlite:
```
JoanFM, line 226: remove this

tests/unit/array/test_advance_indexing.py
```diff
@@ -700,21 +700,23 @@ def test_offset2ids_persistence(storage, config, start_storage):
...
+    with DocumentArray(storage=storage, config=config) as da1:
+        assert da1[:, 'id'] == da_ids
         da1.extend([Document(id=i) for i in 'abc'])
+        da1_ids = da1[:, 'id']
         assert len(da1) == 8
 
-    da2 = DocumentArray(storage=storage, config=config)
-    assert da2[:, 'id'] == da1[:, 'id']
+    with DocumentArray(storage=storage, config=config) as da2:
+        assert da2[:, 'id'] == da1_ids
```
JoanFM, line 719: we should not need here a context manager

tests/unit/array/test_sequence.py
```diff
@@ -226,11 +226,10 @@ def test_del_and_append(index, storage, config):
 
     with da:
         da.extend([Document(id=str(i)) for i in range(5)])
-    with da:
         del da[index]
```
JoanFM, line 229: I think this was on purpose, please keep it

JoanFM (changes requested):
In the tests, make sure `ctxt` manager is only used for `write` operations, otherwise there is something wrong and we need to rethink implementation

tests/unit/array/mixins/test_io.py
```diff
@@ -43,20 +43,31 @@ def test_document_save_load(
...
-    da.save(tmp_file, file_format=method, encoding=encoding)
+
+    with da:
+        da.insert(2, Document(id='new'))
+        da.save(tmp_file, file_format=method, encoding=encoding)
+        da_info = {
+            'id': [d.id for d in da],
+            'embedding': [d.embedding for d in da],
+            'content': [d.content for d in da],
+        }
```
numb3r3, line 54: don't need the ctx, right?

numb3r3 (changes requested):

tests/unit/array/mixins/test_io.py
jemmyshin: yes, will remove this

JoanFM (approved):

JoanFM (changes requested):
Please reenable the test that was disabled by @alaeddine-13 PR #649

jemmyshin:
> Please reenable the test that was disabled by @alaeddine-13 PR #649

Done

numb3r3 (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -131,6 +131,7 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install wheel
           # pip does not properly resolve dependency versions with syntax pip install --no-cache-dir ".[test,full]"
+          pip install --pre annlite
```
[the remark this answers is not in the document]
numb3r3: This pre-release has been deleted in latest commit

tests/unit/array/mixins/test_content.py
```diff
@@ -40,6 +40,9 @@ def test_content_empty_getter_return_none(cls, content_attr, start_storage):
         da = cls()
     assert getattr(da, content_attr) is None
 
+    if cls == DocumentArrayAnnlite:
```
alaeddine-13, line 43: for example here, DA will create an annlite instance with temp file name. I don't think anything would access the same temp file concurrently, why do we need to release the lock ?

jemmyshin: This is because we need to test with different parameters, it's like a for loop, need to close it before another iteration.

JoanFM:
I do not understand this PR. Now u are saying 0.4.0 is already released and all these tests were not failing even without `closing`. So why are we doing all this mess?

jemmyshin:
No, we haven't released annlite 0.4.0 yet. But when we release it, the docarray test will fail without these changes. So this is what this pr is doing: make sure the ci won't fail after 0.4.0 is released.

JoanFM:
@jemmyshin I prefer then to have a PR fixing `annlite` version, and then pushing a change to unpin it and see what breaks and how

PR #631, opened 2022-10-17
feat: add jina to full requirment

samsja:
# Context

DocArray rely on Jina for the `DocumentArray.post` method. 

Nevertheless there are no direct dependency in the requirement (only in test).

# What this PR do:

This PR add jina to the `full` requirement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ setup(
             'fastapi',
             'uvicorn',
             'strawberry-graphql',
+            'jina>=3.10.1',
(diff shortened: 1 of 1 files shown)


samsja:
We don't want to explicitly introduce the a cyclic dependency with Jina which depend itself on docarray. 

The need to install Jina is clearly explained in the [documentation about the post method](https://docarray.jina.ai/fundamentals/documentarray/post-external/).

PR #635, opened 2022-10-17
fix: support qdrant 0.8

dongxiang123:
Goals:

- ...This PR is aimed to fix the support for qdrant from Docarray
- ...

### DIFF
--- a/docarray/array/storage/qdrant/backend.py
+++ b/docarray/array/storage/qdrant/backend.py
@@ -171,7 +171,13 @@ class BackendMixin(BaseBackendMixin):
     def _get_offset2ids_meta(self) -> List[str]:
         if not self._collection_exists(self.collection_name_meta):
             return []
-        return self.client.retrieve(collection_name=self.collection_name_meta, ids=[1])
+        results = self.client.retrieve(
+            collection_name=self.collection_name_meta, ids=[1]
+        )
+        if len(results) == 0:
+            return []
+        else:
+            return results[0].payload.get('offset2id', [])
 
     def _update_offset2ids_meta(self):
         if not self._collection_exists(self.collection_name_meta):
--- a/docarray/array/storage/qdrant/backend.py
+++ b/docarray/array/storage/qdrant/backend.py
@@ -81,6 +81,7 @@ class BackendMixin(BaseBackendMixin):
             config.collection_name = self._tmp_collection_name()
 
         self._n_dim = config.n_dim
+        self._distance = config.distance
         self._serialize_config = config.serialize_config
 
         self._client = QdrantClient(host=config.host, port=config.port)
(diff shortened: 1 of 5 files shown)


docarray/array/storage/qdrant/backend.py
```diff
@@ -133,11 +134,11 @@ def _initialize_qdrant_schema(self):
                 full_scan_threshold=self._config.full_scan_threshold,
                 m=self._config.m,
             )
+            vectors_config = VectorParams(size=self.n_dim, distance=Distance.COSINE)
```
JoanFM, line 137: the distance should be a parameter

docarray/array/storage/qdrant/backend.py
```diff
@@ -175,10 +176,11 @@ def _get_offset2ids_meta(self) -> List[str]:
         ).result.payload['offset2id']
 
     def _update_offset2ids_meta(self):
+        vectors_config = VectorParams(size=1, distance=Distance.COSINE)
```
JoanFM, line 179: same here

tests/unit/array/docker-compose.yml
```diff
@@ -10,7 +10,7 @@ services:
       AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED: 'true'
       PERSISTENCE_DATA_PATH: '/var/lib/weaviate'
   qdrant:
-    image: qdrant/qdrant:v0.7.0
+    image: qdrant/qdrant:v0.10.0
```
JoanFM, line 13: please update to 0.8, not 0.10. 0.10 has spme other problems that will be dealt with in different PR

JoanFM (changes requested):

tests/unit/array/docker-compose.yml
JohannesMessner, line 13: Are we sure we want to go all they way up to v10? I vaguely remember that there was some issue with qdrant 10, have we resolved that? @JoanFM

JohannesMessner (commented):
With this we should also bump our `qdrant-client` dependency to version 8, no? It doesn't look like this preserves compatability with v7, but maybe I wrong

docarray/array/storage/qdrant/backend.py
alaeddine-13: not the same, this is just a hack for offset2id

tests/unit/array/docker-compose.yml
dongxiang123: Yes, with this, we can bump our client to version 8.

dongxiang123:
> check there is another docker compose to change

Yes, that is right.

docarray/array/storage/qdrant/backend.py
alaeddine-13: @dongxiang123  no need to specify n_dim for offset2id collection.
Please re-apply your old changes in `_update_offset2ids_meta` method:
`vectors_config = VectorParams(size=1, distance=Distance.COSINE)`

setup.py
```diff
@@ -102,6 +102,7 @@
             'onnxruntime',
             'jupyterlab',
             'transformers>=4.16.2',
+            'qdrant-client~=0.8.0',
```
JoanFM, line 105: no need to add it here

JoanFM (changes requested):

JoanFM (approved):

PR #643, opened 2022-10-17
feat: multiple metrics in evaluate

guenthermi:
Goals:

- allow users to pass multiple metrics to the evaluation function
- calculate all metric scores at once

### DIFF
--- a/docarray/array/mixins/evaluation.py
+++ b/docarray/array/mixins/evaluation.py
@@ -181,7 +181,9 @@ class EvaluationMixin:
                     'information is provided to evaluate the matches.'
                 )
             for metric_name, metric_fn in zip(metric_names, metric_fns):
-                r = metric_fn(binary_relevance, max_rel=max_rel, **kwargs)
+                if 'max_rel' in metric_fn.__code__.co_varnames:
+                    kwargs['max_rel'] = max_rel
+                r = metric_fn(binary_relevance, **kwargs)
                 d.evaluations[metric_name] = NamedScore(
                     value=r, op_name=str(metric_fn), ref_id=d.id
                 )
--- a/docarray/array/mixins/evaluation.py
+++ b/docarray/array/mixins/evaluation.py
@@ -50,6 +50,7 @@ def _evaluate_deprecation(f):
         )
         if len(args) > 1:
             if type(args[1]) is str:
+                args = list(args)
                 args[1] = [args[1]]
                 warnings.warn(list_warning_msg % 'metrics', DeprecationWarning)
         for key in ['metrics', 'metric_names']:
--- a/docarray/array/mixins/evaluation.py
+++ b/docarray/array/mixins/evaluation.py
@@ -9,9 +9,38 @@ if TYPE_CHECKING:
     from docarray import Document, DocumentArray
 
 
+def _evaluate_deprecation(f):
+    """Raises a deprecation warning if the user executes the evaluate function with
+    the old interface and adjust the input to fit the new interface."""
+
+    def func(*args, **kwargs):
+        if len(args) > 1:
+            if not (isinstance(args[1], Callable) or isinstance(args[1], str)):
+                kwargs['ground_truth'] = args[1]
+                args = [args[0]] + list(args[2:])
+                warnings.warn(
+                    'The `other` attribute in `evaluate()` is transfered from a '
+                    'positional attribute into the keyword attribute `ground_truth`.'
+                    'Using it as a positional attribute is deprecated and will be removed '
+                    'in the next version.',
+                    DeprecationWarning,
+                )
+        if 'other' in kwargs:
+            kwargs['ground_truth'] = kwargs['other']
+            warnings.warn(
+                '`other` is renamed to `groundtruth` in `evaluate()`, the usage of `other` is '
+                'deprecated and will be removed in the next version.',
+                DeprecationWarning,
+            )
+        return f(*args, **kwargs)
+
+    return func
+
+
 class EvaluationMixin:
     """A mixin that provides ranking evaluation functionality to DocumentArrayLike objects"""
 
+    @_evaluate_deprecation
     def evaluate(
         self,
         metric: Union[str, Callable[..., float]],
@@ -19,18 +48,20 @@ class EvaluationMixin:
         hash_fn: Optional[Callable[['Document'], str]] = None,
         metric_name: Optional[str] = None,
         strict: bool = True,
-        label_tag='label',
+        label_tag: Optional[str] = 'label',
         **kwargs,
     ) -> Optional[float]:
         """
         Compute ranking evaluation metrics for a given `DocumentArray` when compared
         with a groundtruth.
 
-        This implementation expects the documents and their matches to have labels
-        annotated inside the tag with the key specified in the `label_tag` attribute.
-        Alternatively, one can provide a `ground_truth` DocumentArray that is
-        structurally identical to `self`. In this case, this function compares the
-        `matches` of `documents` inside the `DocumentArray`.
+        If one provides a `ground_truth` DocumentArray that is structurally identical
+        to `self`, this function compares the `matches` of `documents` inside the
+        `DocumentArray` to this `ground_truth`.
+        Alternatively, one can directly annotate the documents by adding labels in the
+        form of tags with the key specified in the `label_tag` attribute.
+        Those tags need to be added to `self` as well as to the documents in the
+        matches properties.
 
         This method will fill the `evaluations` field of Documents inside this
         `DocumentArray` and will return the average of the computations
@@ -46,6 +77,7 @@ class EvaluationMixin:
         :param strict: If set, then left and right sides are required to be fully
             aligned: on the length, and on the semantic of length. These are preventing
             you to evaluate on irrelevant matches accidentally.
+        :param label_tag: Specifies the tag which contains the labels.
         :param kwargs: Additional keyword arguments to be passed to `metric_fn`
         :return: The average evaluation computed or a list of them if multiple metrics
             are required
@@ -57,7 +89,7 @@ class EvaluationMixin:
         elif label_tag in self[0].tags:
             if ground_truth:
                 warnings.warn(
-                    'An ground_truth attribute is provided but does not '
+                    'A ground_truth attribute is provided but does not '
                     'contain matches. The labels are used instead and '
                     'ground_truth is ignored.'
                 )
@@ -117,7 +149,10 @@ class EvaluationMixin:
                     1 if m.tags[label_tag] == d.tags[label_tag] else 0 for m in targets
                 ]
             else:
-                raise RuntimeError(f'Unsupported groundtruth type {ground_truth_type}')
+                raise RuntimeError(
+                    'Could not identify which kind of ground truth'
+                    'information is provided to evaluate the matches.'
+                )
 
             r = metric_fn(binary_relevance, max_rel=max_rel, **kwargs)
             d.evaluations[metric_name] = NamedScore(
--- a/docarray/array/mixins/evaluation.py
+++ b/docarray/array/mixins/evaluation.py
@@ -41,6 +41,21 @@ def _evaluate_deprecation(f):
                     f'usage of `{old_key}` is deprecated and will be removed soon.',
                     DeprecationWarning,
                 )
+
+        # transfer metrics and metric_names into lists
+        list_warning_msg = (
+            'The attribute `%s` now accepts a list instead of a '
+            'single element. Passing a single element is deprecated and will soon not '
+            'be supported anymore.'
+        )
+        if len(args) > 1:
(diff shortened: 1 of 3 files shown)


docarray/array/mixins/evaluation.py
```diff
@@ -1,59 +1,147 @@
...
+                or isinstance(args[1], str)
+                or isinstance(args[1], list)
+            ):
+                kwargs['ground_truth'] = args[1]
+                args = [args[0]] + list(args[2:])
+                warnings.warn(
+                    'The `other` attribute in `evaluate()` is transfered from a '
+                    'positional attribute into the keyword attribute `ground_truth`.'
+                    'Using it as a positional attribute is deprecated and will be removed '
+                    'in the next version.',
```
JoanFM, line 30: not the next. say soon

docarray/array/mixins/evaluation.py
```diff
@@ -1,59 +1,147 @@
...
     from docarray import Document, DocumentArray
 
 
+def _evaluate_deprecation(f):
+    """Raises a deprecation warning if the user executes the evaluate function with
+    the old interface and adjust the input to fit the new interface."""
+
+    def func(*args, **kwargs):
+        if len(args) > 1:
+            if not (
```
JoanFM, line 19: why not applying `positive logic`?

JoanFM (changes requested):
is this a breaking change with respect to the return of `evaluate`?

docarray/array/mixins/evaluation.py
guenthermi: I can not check for instanceof(DocumentArray) because it results in a circular dependency.

docarray/array/mixins/evaluation.py
guenthermi: I will first change this in https://github.com/jina-ai/docarray/pull/617

guenthermi:
> is this a breaking change with respect to the return of `evaluate`?

Yes, I could return only a single number if the user passes only one metric but having different return types for different inputs is a bit strange

guenthermi:
depends on https://github.com/jina-ai/docarray/pull/617

docarray/array/mixins/evaluation.py
```diff
@@ -43,14 +48,16 @@ class EvaluationMixin:
     @_evaluate_deprecation
     def evaluate(
         self,
-        metric: Union[str, Callable[..., float]],
+        metrics: Union[
+            Union[str, Callable[..., float]], List[Union[str, Callable[..., float]]]
```
JoanFM, line 52: Do the same deprecation with `metric` and `metric_name`. Make sure the name is equivalent, so that we do not add breaking changes

JoanFM (changes requested):

docarray/array/mixins/evaluation.py
```diff
@@ -25,13 +30,17 @@ def func(*args, **kwargs):
...
-                DeprecationWarning,
-            )
+        for old_key, new_key in zip(
+            ['other', 'metric', 'metric_name'],
+            ['ground_truth', 'metrics', 'metric_names'],
+        ):
+            if old_key in kwargs:
+                kwargs[new_key] = kwargs[old_key]
+                warnings.warn(
+                    '`other` is renamed to `groundtruth` in `evaluate()`, the usage of `other` is '
```
samsja, line 40: I guess you want to use new key and old key in ther error message

guenthermi: oh good catch!

docarray/array/mixins/evaluation.py
JoanFM: I agree, there is not so much we can do

samsja: I agree as well on returning always a dict otherwise it can be confusing even though it is a breaking change

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -3,7 +3,7 @@
 After you get `.matches` from the last chapter, you can easily evaluate matches against the groundtruth via {meth}`~docarray.array.mixins.evaluation.EvaluationMixin.evaluate`.
```
JohannesMessner, line 3:
```suggestion
After you get `.matches`, you can evaluate matches against the groundtruth via {meth}`~docarray.array.mixins.evaluation.EvaluationMixin.evaluate`.
```
I know this has nothing to do with this PR, but let's get away form treating our docs like a book

JoanFM (approved):

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -18,7 +18,7 @@ example_da.embeddings = np.random.random([10, 3])
 
 example_da.match(example_da)
 
-example_da.evaluate(metric='precision_at_k')
+example_da.evaluate(metrics='precision_at_k')
```
JohannesMessner, line 21: Can we add an example passing multiple metrics?

JohannesMessner (changes requested):

docs/fundamentals/documentarray/evaluation.md
guenthermi: Ok, I will add an example

alexcg1:
@guenthermi remember to tag @NicholasDunham for any future PRs that include docs (or docstrings)

bwanglzu (commented):
two comments:

1. i don't like the fact that `metrics='precision_at_k'` should be supported, i suppose `metrics` should be always accept a `list`, not a `str`.
2. can we provide a `all` function to evaluate all metrics we have?

guenthermi:
> two comments:
> 
> 1. i don't like the fact that `metrics='precision_at_k'` should be supported, i suppose `metrics` should be always accept a `list`, not a `str`.
> 2. can we provide a `all` option to evaluate all metrics we have?
> 
> it's like what we're doing in `callbacks`, user should always pass a list of callbacks even user only need one. I'm expecting we:
> 
> ```python
> .evaluate(metrics=['precision_at_k'])
> ```

I don't care much if we support passing only a string or not. Having an `all` keyword is not a good idea, because the metric functions have different use cases, e.g., the ndcg metric should not be applied on a top k ranking while for other metrics this makes sense.

JoanFM:
> > two comments:
> > 
> > 1. i don't like the fact that `metrics='precision_at_k'` should be supported, i suppose `metrics` should be always accept a `list`, not a `str`.
> > 2. can we provide a `all` option to evaluate all metrics we have?
> > 
> > it's like what we're doing in `callbacks`, user should always pass a list of callbacks even user only need one. I'm expecting we:
> > ```python
> > .evaluate(metrics=['precision_at_k'])
> > ```
> 
> I don't care much if we support passing only a string or not. @JoanFM @samsja @JohannesMessner Do you have an opinion about it?
> 
> Having an `all` keyword is not a good idea, because the metric functions have different use cases, e.g., the ndcg metric should not be applied on a top k ranking while for other metrics this makes sense.

I believe, that to keep backwards compatibility we need to accept at least `metric=STR` but I agree, it is cleaner if user needs to pass list. But if it is complex, this seems like a minor issue to me

JoanFM:
> > > two comments:
> > > 
> > > 1. i don't like the fact that `metrics='precision_at_k'` should be supported, i suppose `metrics` should be always accept a `list`, not a `str`.
> > > 2. can we provide a `all` option to evaluate all metrics we have?
> > > 
> > > it's like what we're doing in `callbacks`, user should always pass a list of callbacks even user only need one. I'm expecting we:
> > > ```python
> > > .evaluate(metrics=['precision_at_k'])
> > > ```
> > 
> > 
> > I don't care much if we support passing only a string or not. @JoanFM @samsja @JohannesMessner Do you have an opinion about it?
> > Having an `all` keyword is not a good idea, because the metric functions have different use cases, e.g., the ndcg metric should not be applied on a top k ranking while for other metrics this makes sense.
> 
> I believe, that to keep backwards compatibility we need to accept at least `metric=STR` but I agree, it is cleaner if user needs to pass list. But if it is complex, this seems like a minor issue to me

Agree, `all` is not good

bwanglzu:
make sense, ignore my `all` suggestion.

guenthermi:
> > > two comments:
> > > 
> > > 1. i don't like the fact that `metrics='precision_at_k'` should be supported, i suppose `metrics` should be always accept a `list`, not a `str`.
> > > 2. can we provide a `all` option to evaluate all metrics we have?
> > > 
> > > it's like what we're doing in `callbacks`, user should always pass a list of callbacks even user only need one. I'm expecting we:
> > > ```python
> > > .evaluate(metrics=['precision_at_k'])
> > > ```
> > 
> > 
> > I don't care much if we support passing only a string or not. @JoanFM @samsja @JohannesMessner Do you have an opinion about it?
> > Having an `all` keyword is not a good idea, because the metric functions have different use cases, e.g., the ndcg metric should not be applied on a top k ranking while for other metrics this makes sense.
> 
> I believe, that to keep backwards compatibility we need to accept at least `metric=STR` but I agree, it is cleaner if user needs to pass list. But if it is complex, this seems like a minor issue to me

We could transform strings into a list of one string in the deprecation decorator and adjust the documentation. So I would not say that it is complex

JoanFM:
> > > > two comments:
> > > > 
> > > > 1. i don't like the fact that `metrics='precision_at_k'` should be supported, i suppose `metrics` should be always accept a `list`, not a `str`.
> > > > 2. can we provide a `all` option to evaluate all metrics we have?
> > > > 
> > > > it's like what we're doing in `callbacks`, user should always pass a list of callbacks even user only need one. I'm expecting we:
> > > > ```python
> > > > .evaluate(metrics=['precision_at_k'])
> > > > ```
> > > 
> > > 
> > > I don't care much if we support passing only a string or not. @JoanFM @samsja @JohannesMessner Do you have an opinion about it?
> > > Having an `all` keyword is not a good idea, because the metric functions have different use cases, e.g., the ndcg metric should not be applied on a top k ranking while for other metrics this makes sense.
> > 
> > 
> > I believe, that to keep backwards compatibility we need to accept at least `metric=STR` but I agree, it is cleaner if user needs to pass list. But if it is complex, this seems like a minor issue to me
> 
> We could transform strings into a list of one string in the deprecation decorator and adjust the documentation. So I would not say that it is complex

then go for it

bwanglzu (approved):
LGTM!

JoanFM (approved):

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -1,9 +1,9 @@
 # Evaluate Matches
 
-After you get `.matches` from the last chapter, you can easily evaluate matches against the groundtruth via {meth}`~docarray.array.mixins.evaluation.EvaluationMixin.evaluate`.
+After you get `.matches`, you can evaluate matches against the groundtruth via {meth}`~docarray.array.mixins.evaluation.EvaluationMixin.evaluate`.
```
samsja, line 3: Lets briefly say that we can get matches by calling the find method

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -1,9 +1,9 @@
...
 
 ```python
-da_predict.evaluate(ground_truth=da_groundtruth, metric='...', **kwargs)
+da_predict.evaluate(ground_truth=da_groundtruth, metrics=['...'], **kwargs)
```
samsja, line 6: Here it is not clear what is groud truth. (I know it was unclear before as well but lets try to fix it in this PR)

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -18,7 +18,7 @@ example_da.embeddings = np.random.random([10, 3])
 
 example_da.match(example_da)
 
-example_da.evaluate(metric='precision_at_k')
+example_da.evaluate(metrics=['precision_at_k'])
```
samsja, line 21: we needto have example data I believe otherwise user can't understand what is happening. Just random data

samsja (changes requested):
the docstring for evaluate are not reported because of the decorator

new:
image

old:
image

guenthermi:
> I think that we can make the documentation a bit better for evaluation. @NicholasDunham what do u think ?

Maybe we can rewrite the documentation in another PR. There are also some old features of the `evaluate` functions which are not yet documented.

samsja:
> Maybe we can rewrite the documentation in another PR. There are also some old features of the `evaluate` functions which are not yet documented and today, I have not more time to do this.

Okay then lets open an issue for improving this documentation and please align with @NicholasDunham on how do refactor it

guenthermi:
> > Maybe we can rewrite the documentation in another PR. There are also some old features of the `evaluate` functions which are not yet documented and today, I have not more time to do this.
> 
> Okay then lets open an issue for improving this documentation and please align with @NicholasDunham on how do refactor it

I created #651 for the documentation changes.

JohannesMessner (approved):

samsja (approved):

Issue #626, opened 2022-10-14
Bug: unintuitive behavior from csv method return a da but does not modify in place


samsja:
# Context

Calling the the `from_csv` method of a DocumentArray will return a new DocumentArray with the csv content loaded but will not modify the orignal DocumenArray inplace

```python
from docarray import DocumentArray

da = DocumentArray()
da.from_csv(os.path.join(cur_dir, 'toydata/docs.csv'))
assert len(da) > 0 # This will faill da is still empty
```

This work though
```python
from docarray import DocumentArray

da = DocumentArray()
da = da.from_csv(os.path.join(cur_dir, 'toydata/docs.csv'))
assert len(da) > 0 # This will faill da is still empty
```

PR #646, opened 2022-10-18
fix: from csv behaviour

anna-charlotte:
Goals:

- Solving this issue: https://github.com/jina-ai/docarray/issues/626

### DIFF
--- a/docarray/array/mixins/io/from_gen.py
+++ b/docarray/array/mixins/io/from_gen.py
@@ -20,8 +20,8 @@ if TYPE_CHECKING:
 class FromGeneratorMixin:
     """Provide helper functions filling a :class:`DocumentArray`-like object with a generator."""
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
         self.from_csv = self.__from_csv
 
     @classmethod
--- a/docarray/array/mixins/io/from_gen.py
+++ b/docarray/array/mixins/io/from_gen.py
@@ -134,16 +134,15 @@ class FromGeneratorMixin:
         """
         return cls._from_generator('from_csv', *args, **kwargs)
 
-    def from_csv_called_from_instance(cls: Type['T'], *args, **kwargs) -> 'T':
+    def from_csv_called_from_instance(cls: Type['T'], *args, **kwargs) -> None:
         """
         # noqa: DAR101
         # noqa: DAR102
         # noqa: DAR201
         """
-        warnings.warn(
-            'Calling from_csv() from a DocumentArray instance does not change the instance in-place. Instead of calling from_csv() from a DocumentArray instance, you should probably use `DocumentArray.from_csv(...)`.'
+        raise Exception(
+            'Do not call the classmethod from_csv() from a DocumentArray instance. Instead call it from the class itself: `DocumentArray.from_csv(...)`'
         )
-        return cls._from_generator('from_csv', *args, **kwargs)
 
     @classmethod
     @overload
--- a/docarray/array/mixins/io/from_gen.py
+++ b/docarray/array/mixins/io/from_gen.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import (
     Type,
     TYPE_CHECKING,
@@ -18,6 +19,10 @@ if TYPE_CHECKING:
 
 class FromGeneratorMixin:
     """Provide helper functions filling a :class:`DocumentArray`-like object with a generator."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.from_csv = self.__from_csv
 
     @classmethod
     def _from_generator(cls: Type['T'], meth: str, *args, **kwargs) -> 'T':
@@ -127,6 +132,17 @@ class FromGeneratorMixin:
         # noqa: DAR102
         # noqa: DAR201
         """
+        return cls._from_generator('from_csv', *args, **kwargs)
+
+    def __from_csv(cls: Type['T'], *args, **kwargs) -> 'T':
+        """
+        # noqa: DAR101
+        # noqa: DAR102
+        # noqa: DAR201
+        """
+        warnings.warn(
+            'Calling from_csv() from a DocumentArray instance does not change the instance in-place. Instead of calling from_csv() from a DocumentArray instance, you should probably use `DocumentArray.from_csv(...)`.'
+        )
         return cls._from_generator('from_csv', *args, **kwargs)
 
     @classmethod
--- a/docarray/array/mixins/io/from_gen.py
+++ b/docarray/array/mixins/io/from_gen.py
@@ -22,7 +22,7 @@ class FromGeneratorMixin:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.from_csv = self.__from_csv
+        self.from_csv = self.from_csv_called_from_instance
 
     @classmethod
     def _from_generator(cls: Type['T'], meth: str, *args, **kwargs) -> 'T':
@@ -134,7 +134,7 @@ class FromGeneratorMixin:
         """
         return cls._from_generator('from_csv', *args, **kwargs)
 
-    def __from_csv(cls: Type['T'], *args, **kwargs) -> 'T':
+    def from_csv_called_from_instance(cls: Type['T'], *args, **kwargs) -> 'T':
         """
         # noqa: DAR101
         # noqa: DAR102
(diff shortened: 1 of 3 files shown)


JoanFM:
I believe this can be solved inside `from_csv` itself checking if `cls` is an Instance or a class.

docarray/array/mixins/io/from_gen.py
```diff
@@ -129,6 +134,17 @@ def from_csv(cls: Type['T'], *args, **kwargs) -> 'T':
         """
         return cls._from_generator('from_csv', *args, **kwargs)
 
+    def __from_csv(cls: Type['T'], *args, **kwargs) -> 'T':
+        """
+        # noqa: DAR101
+        # noqa: DAR102
+        # noqa: DAR201
+        """
+        warnings.warn(
```
JoanFM, line 143: I am not sure if we should raise a `NotExistingMethod` exception or something like that @samsja , something that is the same as what `pandas` do.

samsja: Yes I agree as well that we should raise an error rather than a warning

JohannesMessner:
> Do not add an `init` to the `Mixin`. DO that in the `init` of DocArray Base class

Are we sure about this? It would assign some methods that don't exist there and break separation of concerns.

JoanFM:
> > Do not add an `init` to the `Mixin`. DO that in the `init` of DocArray Base class
> 
> Are we sure about this? It would assign some methods that don't exist there and break separation of concerns. It seems to be accepted practice to let mixins have inits and pass **kwargs between them: https://stackoverflow.com/a/50465583

multiple inheritance is not so easily supported in Python I think. What is ur super when u call super?

JohannesMessner:
> > > Do not add an `init` to the `Mixin`. DO that in the `init` of DocArray Base class
> > 
> > 
> > Are we sure about this? It would assign some methods that don't exist there and break separation of concerns. It seems to be accepted practice to let mixins have inits and pass **kwargs between them: https://stackoverflow.com/a/50465583
> 
> multiple inheritance is not so easily supported in Python I think. What is ur super when u call super?

My understanding is that it picks a certain evaluation order along which it tries to find an __init__ to call, and then it call the first it finds.
But if that __init__ in turn calls `super().__init__(**kwargs)` then it will keep 'walking' along that evaluation order. So as long as every init in a mixin makes that call it should all work as intended. So the mixins just need to be set up to not break this.

samsja:
> > > > Do not add an `init` to the `Mixin`. DO that in the `init` of DocArray Base class
> > > 
> > > 
> > > Are we sure about this? It would assign some methods that don't exist there and break separation of concerns. It seems to be accepted practice to let mixins have inits and pass **kwargs between them: https://stackoverflow.com/a/50465583
> > 
> > 
> > multiple inheritance is not so easily supported in Python I think. What is ur super when u call super?
> 
> My understanding is that `super()` picks a certain evaluation order along which it tries to find an `__init__` to call, and then it call the first it finds. But if that `__init__` in turn calls `super().__init__(**kwargs)` then it will keep 'walking' along that evaluation order. So as long as every init in a mixin makes that call it should all work as intended. The mixins just need to be set up to not break this chain.

I agree with Johannes it would be better to add it in the Mixin `__init__` but if it is not possible because of inheritance problem we can put in in the DA __init__

JoanFM:
> > > > > Do not add an `init` to the `Mixin`. DO that in the `init` of DocArray Base class
> > > > 
> > > > 
> > > > Are we sure about this? It would assign some methods that don't exist there and break separation of concerns. It seems to be accepted practice to let mixins have inits and pass **kwargs between them: https://stackoverflow.com/a/50465583
> > > 
> > > 
> > > multiple inheritance is not so easily supported in Python I think. What is ur super when u call super?
> > 
> > 
> > My understanding is that `super()` picks a certain evaluation order along which it tries to find an `__init__` to call, and then it call the first it finds. But if that `__init__` in turn calls `super().__init__(**kwargs)` then it will keep 'walking' along that evaluation order. So as long as every init in a mixin makes that call it should all work as intended. The mixins just need to be set up to not break this chain.
> 
> I agree with Johannes it would be better to add it in the Mixin `__init__` but if it is not possible because of inheritance problem we can put in in the DA **init**

Do not put it there. Mixins should be stateless

JohannesMessner:
> > > > > > Do not add an `init` to the `Mixin`. DO that in the `init` of DocArray Base class
> > > > > 
> > > > > 
> > > > > Are we sure about this? It would assign some methods that don't exist there and break separation of concerns. It seems to be accepted practice to let mixins have inits and pass **kwargs between them: https://stackoverflow.com/a/50465583
> > > > 
> > > > 
> > > > multiple inheritance is not so easily supported in Python I think. What is ur super when u call super?
> > > 
> > > 
> > > My understanding is that `super()` picks a certain evaluation order along which it tries to find an `__init__` to call, and then it call the first it finds. But if that `__init__` in turn calls `super().__init__(**kwargs)` then it will keep 'walking' along that evaluation order. So as long as every init in a mixin makes that call it should all work as intended. The mixins just need to be set up to not break this chain.
> > 
> > 
> > I agree with Johannes it would be better to add it in the Mixin `__init__` but if it is not possible because of inheritance problem we can put in in the DA **init**
> 
> Do not put it there. Mixins should be stateless

But by moving it to the base class we don't avoid the state, we just move it to somewhere else. The only thing that changes is that the assignment happens somewhere where we don't know the thing that we assign...

But not willing to die in this hill, @anna-charlotte you can choose who to listen to ;)

JoanFM:
> > > > > > > Do not add an `init` to the `Mixin`. DO that in the `init` of DocArray Base class
> > > > > > 
> > > > > > 
> > > > > > Are we sure about this? It would assign some methods that don't exist there and break separation of concerns. It seems to be accepted practice to let mixins have inits and pass **kwargs between them: https://stackoverflow.com/a/50465583
> > > > > 
> > > > > 
> > > > > multiple inheritance is not so easily supported in Python I think. What is ur super when u call super?
> > > > 
> > > > 
> > > > My understanding is that `super()` picks a certain evaluation order along which it tries to find an `__init__` to call, and then it call the first it finds. But if that `__init__` in turn calls `super().__init__(**kwargs)` then it will keep 'walking' along that evaluation order. So as long as every init in a mixin makes that call it should all work as intended. The mixins just need to be set up to not break this chain.
> > > 
> > > 
> > > I agree with Johannes it would be better to add it in the Mixin `__init__` but if it is not possible because of inheritance problem we can put in in the DA **init**
> > 
> > 
> > Do not put it there. Mixins should be stateless
> 
> But by moving it to the base class we don't avoid the state, we just move it to somewhere else. The only thing that changes is that the assignment happens somewhere where we don't know the thing that we assign...
> 
> But not willing to die in this hill, @anna-charlotte you can choose who to listen to ;)

What doesn't need state is the Mixin, not the class.

JoanFM (approved):

JohannesMessner:
> > > > > > > > Do not add an `init` to the `Mixin`. DO that in the `init` of DocArray Base class
> > > > > > > 
> > > > > > > 
> > > > > > > Are we sure about this? It would assign some methods that don't exist there and break separation of concerns. It seems to be accepted practice to let mixins have inits and pass **kwargs between them: https://stackoverflow.com/a/50465583
> > > > > > 
> > > > > > 
> > > > > > multiple inheritance is not so easily supported in Python I think. What is ur super when u call super?
> > > > > 
> > > > > 
> > > > > My understanding is that `super()` picks a certain evaluation order along which it tries to find an `__init__` to call, and then it call the first it finds. But if that `__init__` in turn calls `super().__init__(**kwargs)` then it will keep 'walking' along that evaluation order. So as long as every init in a mixin makes that call it should all work as intended. The mixins just need to be set up to not break this chain.
> > > > 
> > > > 
> > > > I agree with Johannes it would be better to add it in the Mixin `__init__` but if it is not possible because of inheritance problem we can put in in the DA **init**
> > > 
> > > 
> > > Do not put it there. Mixins should be stateless
> > 
> > 
> > But by moving it to the base class we don't avoid the state, we just move it to somewhere else. The only thing that changes is that the assignment happens somewhere where we don't know the thing that we assign...
> > But not willing to die in this hill, @anna-charlotte you can choose who to listen to ;)
> 
> What doesn't need state is the Mixin, not the class.

I don't agree, the only 'state' we have is which implementation of `form_csv` we pick, and this is tied to the mixin, not anything else.

JoanFM (changes requested):
sorry, not approved, let's raise Exception

samsja (changes requested):
Looks good so far ! Next step is to make it a bit more generic so that it work for the all of the `from_xxx` classmethod

JoanFM (approved):

PR #647, opened 2022-10-18
feat: show server error message in push

alaeddine-13:
Show the error message coming from the server in push, if the code is 403

### DIFF
--- a/docarray/array/mixins/io/pushpull.py
+++ b/docarray/array/mixins/io/pushpull.py
@@ -3,6 +3,7 @@ import os
 import os.path
 import warnings
 from collections import Counter
+from http import HTTPStatus
 from pathlib import Path
 from typing import Dict, Type, TYPE_CHECKING, List, Optional, Any
 
@@ -250,7 +251,7 @@ class PushPullMixin:
         if response.ok:
             return response.json()['data']
         else:
-            if response.status_code == 403:
+            if response.status_code == HTTPStatus.FORBIDDEN:
                 response.reason = response.json()['message']
             response.raise_for_status()
 
--- a/docarray/array/mixins/io/pushpull.py
+++ b/docarray/array/mixins/io/pushpull.py
@@ -250,6 +250,8 @@ class PushPullMixin:
         if response.ok:
             return response.json()['data']
         else:
+            if response.status_code == 403:
+                response.reason = response.json()['message']
             response.raise_for_status()
 
     @classmethod
--- a/docarray/array/mixins/io/pushpull.py
+++ b/docarray/array/mixins/io/pushpull.py
@@ -251,8 +251,8 @@ class PushPullMixin:
         if response.ok:
             return response.json()['data']
         else:
-            if response.status_code == HTTPStatus.FORBIDDEN:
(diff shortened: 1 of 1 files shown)


docarray/array/mixins/io/pushpull.py
```diff
@@ -250,6 +250,8 @@ def _get_chunk(_batch):
         if response.ok:
             return response.json()['data']
         else:
+            if response.status_code == 403:
```
JoanFM, line 253: 403 seems a magic number, lets use a constant

JoanFM (changes requested):

docarray/array/mixins/io/pushpull.py
JoanFM: @jina-ai/team-hubble @delgermurun @mapleeit these constants should come from hubble-sdk

mapleeit: Does it make sense to use `http.HTTPStatus.FORBIDDEN` here? 

It's standard http status code and `libarary/http` already provides that.

docarray/array/mixins/io/pushpull.py
```diff
@@ -250,6 +250,8 @@ def _get_chunk(_batch):
...
+                response.reason = response.json()['message']
```
delgermurun, line 254: The `403` error is not the only one. How about trying to get a message from every error response?
All errors should have a message, but checking for its existence is also helpful to have.

docarray/array/mixins/io/pushpull.py
delgermurun: The `403` error is not the only one. How about trying to get a message from every error response?
All errors should have a message, but checking for its existence is also helpful to have.

alaeddine-13: applied

mapleeit: Agree with @delgermurun 

For all the status code `>= 400`, we should adopt the same strategy, which is printing the error message out.

alaeddine-13: done

delgermurun (approved):
LGTM

docarray/array/mixins/io/pushpull.py
```diff
@@ -250,6 +251,8 @@ def _get_chunk(_batch):
         if response.ok:
             return response.json()['data']
         else:
+            if response.status_code >= 400 and 'readableMessage' in response.json():
+                response.reason = response.json()['readableMessage']
```
JoanFM, line 255: u should use HTTPStatus for somehing. There must be some constant from there, 400 seems magic to me

JoanFM (changes requested):

docarray/array/mixins/io/pushpull.py
alaeddine-13: I would've used it if it was an equal check (==).
I think it is more readable to use the actual number in comparison (it just means failed response).
The requests library does the same actually
image

JoanFM (approved):

Issue #629, opened 2022-10-17
Add versioned Documentation


JoanFM:
Add versioned documentation.

Documentation page should have a scroll to show documentation of every patch and the master branch

PR #664, opened 2022-10-23
docs: add versioned documentation

Jackmin801:
Goals:
- [x] Address https://github.com/jina-ai/docarray/issues/629

Docs Changes:
- Added a version selector to the sidebar in `docs/_templates/sidebar/brand.html`.

CI changes:
- Updates the `Manual Docs Build` workflow (`.github/workflows/force-docs-build.yml`).
  - This can be run whenever the docs are updated.
  - Changed some steps to be compatible with the new folder structure introduced by versioned docs
  - Increased modularity of steps to increase separation of concerns as well as increase visibility on time taken for each concern.
  - Decreased time taken to fetch code by changing fetches to shallow (depth 1)
  
- Added the `Build old docs` workflow (`.github/workflows/build-old-docs.yml`)
  - This should be run when a new release is created (because there will be a new old release)
  - The workflow will build the docs for all the versions specified in `docs/_versions.json` of the `main` branch
  - Each version  of the docs is built by checking out to the release, pulling the latest `docs/_templates/sidebar/brand.html` in `main`. Then building with `makedoc.sh`.
  - Each version is then placed in separate folders in `./docs` of the pages_branch.


Jackmin801:
Workflow to build latest docs was tested in https://github.com/Jackmin801/JinaDocArrayTest/actions/runs/3307372739/jobs/5458957225

Workflow to build docs for old releases was tested in https://github.com/Jackmin801/JinaDocArrayTest/actions/runs/3307151516

Deployed site from testing workflows: https://docarray.meows.ninja/

JoanFM (changes requested):
We would need an action when doing a `Release` to add an entry for `docs/_versions.json` right?

.github/workflows/build-old-docs.yml
```diff
@@ -0,0 +1,116 @@
+name: Build old docs
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_token:
+        description: 'Your release token'
+        required: true
+      triggered_by:
+        description: 'CD | TAG | MANUAL'
```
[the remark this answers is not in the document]
Jackmin801: It's always triggered manually. I didn't add this line, it's from the old workflow for Manual Docs Build. Should we take it out? I believe it's not actually used anywhere.

.github/workflows/build-old-docs.yml
```diff
@@ -0,0 +1,116 @@
...
+      matrix:
+        include: ${{ fromJson(needs.token-check.outputs.versions) }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+          ref: ${{ matrix.version }}
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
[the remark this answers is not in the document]
Jackmin801: Ah ok

Jackmin801:
> We would need an action when doing a Release to add an entry for docs/_versions.json right?

Yup. Do we put that in `Release CD` or `Manual Release`?

Release CD: https://github.com/jina-ai/docarray/blob/3a888015277f227c79485dcdf7c5dbe40f9516d2/.github/workflows/tag.yml

Manual Release: https://github.com/jina-ai/docarray/blob/3a888015277f227c79485dcdf7c5dbe40f9516d2/.github/workflows/force-release.yml

.github/workflows/build-old-docs.yml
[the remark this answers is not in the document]
JoanFM: okey good

JoanFM:
> > We would need an action when doing a Release to add an entry for docs/_versions.json right?
> 
> Yup. Do we put that in `Release CD` or `Manual Release`?
> 
> Release CD: https://github.com/jina-ai/docarray/blob/3a888015277f227c79485dcdf7c5dbe40f9516d2/.github/workflows/tag.yml
> 
> Manual Release: https://github.com/jina-ai/docarray/blob/3a888015277f227c79485dcdf7c5dbe40f9516d2/.github/workflows/force-release.yml

in Manual Release

.github/workflows/build-old-docs.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - run: echo "Success!"
         if: "${{ github.event.inputs.release_token }} == ${{ env.release_token }}"
         env:
-          release_token: ${{ secrets.JINA_DEV_RELEASE_TOKEN }}
+          release_token: ${{ secrets.DOCARRAY_RELEASE_TOKEN }}
```
JoanFM, line 39: do not change this for now. This should be changed in another scope

JoanFM (changes requested):

docs/conf.py
```diff
@@ -3,56 +3,56 @@
...
-
-project = 'DocArray'
-slug = re.sub(r'\W+', '-', project.lower())
-author = 'Jina AI'
-copyright = 'Jina AI Limited. All rights reserved.'
-source_suffix = ['.rst', '.md']
-master_doc = 'index'
-language = 'en'
-repo_dir = '../'
+sys.path.insert(0, path.abspath(".."))
```
alaeddine-13, line 6: we usually prefer single quote

scripts/release.sh
```diff
@@ -72,6 +72,16 @@ export RELEASE_VER=$(sed -n '/^__version__/p' $INIT_FILE | cut -d \' -f2)
 LAST_VER=$(git tag -l | sort -V | tail -n1)
 printf "last version: \e[1;32m$LAST_VER\e[0m\n"
 
+# Update new _versions.yml and _versions.json if necessary
+if [[ ! $(head -n1 ./docs/_versions.yml) == "- version: v$RELEASE_VER" ]]; then
+    printf -- "- version: v$RELEASE_VER\n" > ./docs/new_versions.yml
+    cat ./docs/_versions.yml >> ./docs/new_versions.yml
+    mv ./docs/new_versions.yml ./docs/_versions.yml
+    python ./scripts/versions_yaml_2_json.py
+    git add ./docs/_versions.yml
```
alaeddine-13, line 81: Why do we keep json and yaml?

.github/workflows/build-old-docs.yml
```diff
@@ -0,0 +1,126 @@
...
+          pip install --no-cache-dir ".[full]"
+          pip install --no-cache-dir ".[qdrant]"
+          pip install --no-cache-dir ".[annlite]"
+          pip install --no-cache-dir ".[weaviate]"
+          pip install --no-cache-dir ".[elasticsearch]"
+          pip install --no-cache-dir ".[redis]"
+          cd docs
+          pip install -r requirements.txt
+          pip install --pre -U furo
+          pip install sphinx-markdown-tables==0.0.17
```
alaeddine-13, line 83: Why isn't this part of requirements.txt?

docs/_templates/sidebar/brand.html
```diff
@@ -16,26 +16,33 @@
   {%- endif %}
   {% endblock brand_content %}
 </a>
+<script>
+  function setPrefix(prefix){
+    window.sessionStorage.setItem("version-select-index", document.getElementsByClassName("version-select")[0].selectedIndex);
```
alaeddine-13, line 21: Do we have to use session storage?
Can't we just open the versioned page ?

.github/workflows/build-old-docs.yml
Jackmin801: Oh. This is actually a typo 😅 
The token used in docarray repo is `DOCARRAY_RELEASE_TOKEN`:
- https://github.com/jina-ai/docarray/blob/main/.github/workflows/force-docs-build.yml#L21
- https://github.com/jina-ai/docarray/blob/main/.github/workflows/force-release.yml#L20

`JINA_DEV_RELEASE_TOKEN` is from Jina repo

Jackmin801: I didn't catch this in my testing because it doesn't actually work...
The step will still pass and docs will still build even if the release token is wrong

docs/conf.py
Jackmin801: Oh these changes are from black. I can revert them but the next person who tries to black this section of code will have the same changes

Jackmin801: Hrmm but it will conflict with ur PR on redirects. Ill revert them here for now. Can blacken later if u wish

scripts/release.sh
Jackmin801: The yaml is more human readable and the json is more machine readable.
`_versions.yml` can go and` _versions.json` stay and everything will still work.
Just it's a pain to edit the `_versions.json` by hand compared to the `_versions.yml`

.github/workflows/build-old-docs.yml
Jackmin801: No particular reason other than I forgot 😅 . Thanks for pointing it out

Jackmin801: Testing the pipeline with the version pin in requirements.txt now

docs/_templates/sidebar/brand.html
Jackmin801: Ah this is to save the state of the selector between pages. Otherwise, when you switch document versions, you will be brought to the correct page but the selector will default back to `latest`

Jackmin801: By having the session state, we save the item they selected before they changed pages so the selector is consistent

.github/workflows/build-old-docs.yml
Jackmin801: PoC: https://github.com/Jackmin801/JinaDocArrayTest/actions/runs/3318846242/workflow

scripts/release.sh
JoanFM: why does it need to be edited by hand?

Jackmin801: We might want to change the versions we support documentation for

Jackmin801: But should be pretty rare. Maybe it's not worth the added complexity. Just have the person run the Python REPL to edit the JSON if we ever need it.

JoanFM: yes, let's have a single source of truth

JoanFM (changes requested):

.github/workflows/build-old-docs.yml
Jackmin801: Oh wait it doesnt work. Because the requirements.txt for old doc builds will use the old requirements.txt

Jackmin801: Yea its a bit ugly to pin in post like this. We could go back in time and pin the requirements.txt of the individual releases but might be overkill considering this works atm.

JoanFM (approved):

PR #676, opened 2022-10-25
chore: bump annlite version

JoanFM:
Goals:

Bump ANNLite version to see what breaks


setup.py
```diff
@@ -68,7 +68,7 @@
             'qdrant-client==0.8.0',
         ],
         'annlite': [
-            'annlite==0.3.13',
+            'annlite',
```
alaeddine-13, line 71: set correct mimimum version here

setup.py
```diff
@@ -103,7 +103,7 @@
             'jupyterlab',
             'transformers>=4.16.2',
             'weaviate-client~=3.3.0',
-            'annlite>=0.3.12',
+            'annlite',
```
alaeddine-13, line 106: set correct mimimum version here

docarray/array/storage/annlite/backend.py
```diff
@@ -80,7 +81,7 @@ def _init_storage(
         subindex_configs: Optional[Dict] = None,
         **kwargs,
     ):
-
+        config = copy.deepcopy(config)
```
jemmyshin, line 84: why deepcopy here? the `data_path` doesn't change

alaeddine-13: we need it because in some cases you pass the same config object to multiple das. So even though data_path is randomly generated for some DA, it will be re-used for the next one, implicitly.
That's why I try to enforce the same pattern for all backends to avoid this kind of issues

docarray/array/storage/base/getsetdel.py
```diff
@@ -205,9 +205,8 @@ def _update_subindices_set(self, set_index, docs):
         else:  # root level set, update subindices iteratively
             for subindex_selector, subindex_da in subindices.items():
                 old_ids = DocumentArray(self[set_index])[subindex_selector, 'id']
-                with subindex_da:
-                    del subindex_da[old_ids]
-                    subindex_da.extend(DocumentArray(docs)[subindex_selector])
+                del subindex_da[old_ids]
+                subindex_da.extend(DocumentArray(docs)[subindex_selector])
```
JoanFM, line 209: why this change @alaeddine-13 ?

alaeddine-13: this is problematic for annlite. But in general I think subindices should be synced when the parent index is synced

JoanFM: but why not use the context manager?

Issue #678, opened 2022-10-25
Docs text attribute disappears after calling load_uri_to_point_cloud_tensor() on it


anna-charlotte:
When creating a Document instance with a `text` and `uri` attribute, the `text` attribute disappears after calling the method `load_uri_to_point_cloud_tensor()`on the Document instance.

Example:
```python
doc = Document(uri='some_uri', text='hello')
doc.load_uri_to_point_cloud_tensor(samples=100)

assert doc.text == 'hello'
```

Running this results in an assertion error: 
```python 
AssertionError: assert '' == 'hello'
```

PR #680, opened 2022-10-25
fix: keep docs text attr after calling load_uri_to_cloud_point_tensor

anna-charlotte:
Goals:

- Currently, a documents `text` attribute disappears after calling `load_uri_to_point_cloud_tensor` on the Document instance, and one can no longer access it. This has been described in the following issue: https://github.com/jina-ai/docarray/issues/678
- We therefore want to ensure that a documents text attribute is not being removed after calling this method on the Document instance.

### DIFF
--- a/tests/unit/document/test_converters.py
+++ b/tests/unit/document/test_converters.py
@@ -274,3 +274,17 @@ def test_glb_converters(uri, chunk_num):
     doc.load_uri_to_point_cloud_tensor(2000, as_chunks=True)
     assert len(doc.chunks) == chunk_num
     assert doc.chunks[0].tensor.shape == (2000, 3)
+
+
+@pytest.mark.parametrize(
+    'uri',
+    [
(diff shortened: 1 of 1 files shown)


tests/unit/document/test_converters.py
```diff
@@ -274,3 +274,17 @@ def test_glb_converters(uri, chunk_num):
     doc.load_uri_to_point_cloud_tensor(2000, as_chunks=True)
     assert len(doc.chunks) == chunk_num
     assert doc.chunks[0].tensor.shape == (2000, 3)
+
+
+@pytest.mark.parametrize(
+    'uri',
+    [
+        os.path.join(cur_dir, 'toydata/test.glb'),
+        'https://github.com/jina-ai/docarray/raw/main/tests/unit/document/toydata/test.glb',
```
JoanFM, line 283: do not enter such dependency, there is no need and can make the test flaky if github is down or something

JoanFM (changes requested):

PR #683, opened 2022-10-25
feat: add batch operations for elastic

dongxiang123:
Goals:

- ...Add batch operations for elastic backend
- https://docarray.jina.ai/advanced/document-store/extend/
- ...

### DIFF
--- a/docarray/array/storage/elastic/getsetdel.py
+++ b/docarray/array/storage/elastic/getsetdel.py
@@ -118,6 +118,29 @@ class GetSetDelMixin(BaseGetSetDelMixin):
             self._client.delete(index=self._config.index_name, id=_id)
         self._refresh(self._config.index_name)
 
+    def _del_docs_by_ids(self, _ids: Sequence[str]):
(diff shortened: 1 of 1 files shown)


docarray/array/storage/elastic/getsetdel.py
```diff
@@ -118,6 +118,29 @@ def _del_doc_by_id(self, _id: str):
...
+        """
+        body = []
+        for _id in _ids:
+            if self._doc_id_exists(id):
+                body.append({"delete": {"_index": self._config.index_name, "_id": _id}})
+        if len(body) > 0:
+            self._client.bulk(body)
+            self._refresh(self._config.index_name)
+
+    def _set_doc_value_pairs(
```
JohannesMessner, line 134: I think instead of having basically empty implementations here these methods should not be implemented at all. That way the standard way provided by a super-class can kick in.

Or is the plan to implement all of these as part of this PR @dongxiang123 ?

JohannesMessner (changes requested):

samsja:
We closed this PR as we are not working on it anymore

PR #685, opened 2022-10-25
ci: fix token check

Jackmin801:
Goals:

- Fix token check in workflows

Proof that new method works:
- Failed: https://github.com/Jackmin801/JinaDocArrayTest/actions/runs/3321088596
- Succeeded: https://github.com/Jackmin801/JinaDocArrayTest/actions/runs/3321089132

Proof that old method didnt work:
- https://github.com/Jackmin801/JinaDocArrayTest/actions/runs/3318846242/jobs/5483262732

### DIFF
--- a/.github/workflows/build-old-docs.yml
+++ b/.github/workflows/build-old-docs.yml
@@ -33,11 +33,18 @@ jobs:
   token-check:
     runs-on: ubuntu-latest
     steps:
-      - run: echo "Success!"
-        if: "${{ github.event.inputs.release_token }} == ${{ env.release_token }}"
+      - name: Check release token
+        id: token-check
+        run: | 
+          touch SUCCESS
+        if: inputs.release_token == env.release_token
         env:
           release_token: ${{ secrets.DOCARRAY_RELEASE_TOKEN }}
-      - id: get_versions
+      - name: Fail release token
+        run: |
+          [[ -f SUCCESS ]]
+      - name: Get versions
+        id: get_versions
         run: |
           printf "versions=" >> $GITHUB_OUTPUT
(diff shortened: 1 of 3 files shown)


Jackmin801:
We cant use secrets directly in the conditional because of a constraint on `secrets` in Github Actions: https://stackoverflow.com/questions/70249519/how-to-check-if-a-secret-variable-is-empty-in-if-conditional-github-actions

Jackmin801:
We have to have a step that checks that the first step (with the conditional) ran. Otherwise it will just skip the step and still succeed.

.github/workflows/build-old-docs.yml
```diff
@@ -33,11 +33,18 @@ jobs:
...
-        if: "${{ github.event.inputs.release_token }} == ${{ env.release_token }}"
+      - name: Check release token
+        id: token-check
+        run: | 
+          touch SUCCESS
+        if: inputs.release_token == env.release_token
         env:
           release_token: ${{ secrets.DOCARRAY_RELEASE_TOKEN }}
-      - id: get_versions
+      - name: Fail release token
```
JoanFM, line 43: couldn't this be done in the same `step`?

Jackmin801: I don't think so. Because the conditional `if` will skip the step if the condition is false.
Skipping does not fail the job.

Basically we need to turn the skip into a fail. One way to go about it is to have another step that tries to read something that would be done in the skipped step and fails if the skip had occurred

Jackmin801: > We have to have a step (`Fail release token`) that fails if the first step (`Check release token`) wasn't ran. Otherwise it will just skip `Check release token` and still succeed.

JoanFM (approved):

Issue #447, opened 2022-07-22
Bug: initializing Document with dataclass and additional kwargs


JohannesMessner:
# Description

Initializing a Document with a dataclass and other kwargs ignores the dataclass, and only the top level kwargs get set.

# How to reproduce

```python
from docarray import DocumentArray, Document, dataclass
from docarray.typing import Text

@dataclass
class MyDoc:
    chunk_text: Text

d = Document(MyDoc(chunk_text='chunk level text'), text='top level text')
print(d.is_multimodal)  # should be True, but is False
print(d.text)  # should print 'top level text'
print(d.chunk_text.text)  # should print 'chunk level text', but fails
```

# Possible fix

In the constructor the dataclass information gets correctly put into `self._data` in [this line](https://github.com/jina-ai/docarray/blob/main/docarray/base.py#L38), but than `self._data` get overridden [here](https://github.com/jina-ai/docarray/blob/main/docarray/base.py#L45).
We probably need some logic to merge the two.

PR #694, opened 2022-10-26
fix: initialize doc with dataclass obj and kwargs

anna-charlotte:
Goals:

- Allow initialization of a `Document` instance with a dataclass obj as well as additional `kwargs`.
- Currently, when a Document is initialized with those two, the attributes passed with the dataclass object are not being saved in the `Document`instance, due to the fact of overwriting `self._data`, as described here: https://github.com/jina-ai/docarray/issues/447
  
## Example 

```python
from docarray import DocumentArray, Document, dataclass
from docarray.typing import Text

@dataclass
class MyDoc:
    chunk_text: Text

d = Document(MyDoc(chunk_text='chunk level text'), text='top level text')
print(d.is_multimodal)  # should be True, but is False
print(d.text)  # should print 'top level text'
print(d.chunk_text.text)  # should print 'chunk level text', but fails
```


JohannesMessner (changes requested):
Looking good!

But can we have a test where we see how conflicting fields are handled?
I think if you use a basic type like `str` in the data class it will be put in the `tags` of the top level Document. What happens if I then also set tags as **kwargs? What if the keys in these two different tag dictionaries clash, what if they don't?

JohannesMessner (approved):

JoanFM (approved):

PR #699, opened 2022-10-27
fix: add none value in document construction

dongxiang123:
Goals:

The None values in the column will cause errors in document construction.
- ...The following snippet can be executed

```python

DA_CONFIG = {
    "storage": "weaviate",
    "config": {
        "name": "Persisted",
        "host": "localhost",
        "port": 8080,
        "vector_cache_max_objects": 3000000,
        "columns": [("rank", "int"), ("year", "int")]
    }
}

da = DocumentArray(**DA_CONFIG)
docs = [Document(text="Some text here", rank=2, year=2020),
        Document(text="Another kind of text here", rank=1, year=2022),
        Document(text="This document is bad", rank=2, year=3)]

with da:
    da.clear()
    da.extend(docs)

da[:,'tags__rank']
But when we have a None value then we can't index to a Document containing a None in a column that was specified with a certain datatype.

DA_CONFIG = {
    "storage": "weaviate",
    "config": {
        "name": "Persisted",
        "host": "localhost",
        "port": 8080,
        "vector_cache_max_objects": 3000000,
        "columns": [("rank", "int"), ("year", "int")]
    }
}

da = DocumentArray(**DA_CONFIG)
docs = [Document(text="Some text here", rank=2, year=2020),
        Document(text="Another kind of text here", rank=1, year=2022),
        Document(text="This document is bad", rank=None, year=3)]

with da:
    da.clear()
    da.extend(docs)
```
# this will break
da[:,'tags__rank']
- ...


docarray/array/storage/base/backend.py
```diff
@@ -75,7 +75,10 @@ def _map_id(self, _id: str) -> str:
         return _id
 
     def _map_column(self, value, col_type) -> str:
-        return self.TYPE_MAP[col_type].converter(value)
+        if value is not None:
+            return self.TYPE_MAP[col_type].converter(value)
+        else:
+            raise ValueError('Column values should not be None.')
```
alaeddine-13, line 81: there's no other backend that supports nullable fields ?

dongxiang123: Actually other backend also don't support nullbale fields, like None for n_dim in other backends, I think it is better to keep consistent for all the backends.

JoanFM: we have to aim to enable `nullable` fields wherever we can

JoanFM (changes requested):

docarray/array/storage/base/backend.py
samsja: yes I agree I would say allow `nullable` should be the default and we deactivate it for some given backend that don't support it

dongxiang123:
Because we need to switch to tasks with higher priority, this one is temporarily on hold.

dongxiang123:
Here is the summary for backends and whether they are supporting the none value configs:
1. Redis: 
    1. When sending questions, it is not working because dataclass checking.
2. Elastic:
    1. Support by default
3. qdrant:
    1.  not working because of the Qdrantic learning. 
4. annlite: Support by default
5. sqlite:
    1. No founding cases.
6. weaviate: has the room to improve

samsja:
We closed this PR as we are not working on it anymore

PR #718, opened 2022-11-03
feat: add flag to dis-able offsettoid

dongxiang123:
Goals:

- ...
- ...DocArray allows list-like behavior by adding offset2id structure to storage backends. This means the database stores, along with documents, meta information about document order.
This information is only updated/synced when a DA is destroyed or quit the context manager. Furthermore, offset2id is updated in memory in all update operations.
However, offset2id is not useful in indexers where concurrent usage is possible and users do not need information about document location. Besides, updating offset2id is not efficient.
Therefore, it's important to have an option that disables list-like behavior (and offset2id with it) and use it in DocumentArray indexers.
```python
@dataclass
class QdrantConfig:
    n_dim: int
    distance: str = 'cosine'
    enable_offset2id: bool = True
    collection_name: Optional[str] = None
    host: Optional[str] = field(default="localhost")
    port: Optional[int] = field(default=6333)
    serialize_config: Dict = field(default_factory=dict)
    scroll_batch_size: int = 64
    ef_construct: Optional[int] = None
    full_scan_threshold: Optional[int] = None
    m: Optional[int] = None
    columns: Optional[Union[List[Tuple[str, str]], Dict[str, str]]] = None
```


docarray/array/document.py
```diff
@@ -80,6 +80,7 @@ def __new__(
         cls,
         _docs: Optional['DocumentArraySourceType'] = None,
         copy: bool = False,
+        offset2id: bool = True,
```
JoanFM, line 83: I think we should remove

JoanFM (changes requested):

samsja:
btw to fix the CI commit message you should rename offset2id to offsettoid (yes we don't allow number in the commit message :sweat_smile:

samsja:
@dongxiang123 Can you put a python snippet of the features in the docs description ?

docarray/array/storage/memory/getsetdel.py
```diff
@@ -24,22 +24,31 @@ def _del_docs_by_slice(self, _slice: slice):
         del self._data[_slice]
 
     def _del_doc_by_id(self, _id: str):
-        self._del_doc_by_offset(self._id2offset[_id])
+        if self._enable_offset2id:
+            self._del_doc_by_offset(self._id2offset[_id])
+        else:
+            raise ValueError("Id2offset is disabled")
```
JoanFM, line 30: I think in the `DocumentArrayInMemory` this should not be applied. For the `InMemory` one the list-like behavior should always work

docarray/array/storage/memory/getsetdel.py
```diff
@@ -57,14 +66,18 @@ def _get_doc_by_offset(self, offset: int) -> 'Document':
...
+        else:
+            raise ValueError("ID2Offset is disabled.")
 
     def _get_docs_by_slice(self, _slice: slice) -> Iterable['Document']:
         return self._data[_slice]
 
     def _clear_storage(self):
         self._data.clear()
-        self._id2offset.clear()
+        if self._enable_offset2id:
```
JoanFM, line 79: Do not touch the `InMemory` implementation

JoanFM (changes requested):

JohannesMessner (changes requested):
Since this is a feature that we expect all backends to handle, we should adjust the documentation about "Add new Document Store" accordingly.

samsja:
Cannot we have this behavior shared across all of the backend ?

JohannesMessner:
> Cannot we have this behavior shared across all of the backend ?

Good point, I think we can. I think we can pass a flag to the `Offset2id` class that tells it to not do any syncing. Then `append` etc would just be no-ops that can be called from the backend with no effect. That way the backend would only have to pass the flag that it receives in its config to offset2id, once.

alphinside:
Hi, I see that the no offset2id flag is passed around to disable sync. Maybe you guys also can consider adding method to detect if the persisted data is having offset2id meta or not. This way it will avoid offset2id inconsistency issues when a Docarray is re-initialized again but somehow forgot to set the config of `enable_offset2id` to False, it will automatically assign to False rather than True

dongxiang123:
@JohannesMessner @samsja This way sounds hacky to me, even if it is cleaner in this way, let's discuss this way in today's Daily.

dongxiang123:
I created a new PR with new approach here. https://github.com/docarray/docarray/pull/730, and this one will be closed.

PR #735, opened 2022-11-08
chore(document): change type hint for tags in docarray.document.data.…

marcosbodio:
…DocumentData

Goals:

- ...
- ...


docarray/document/mixins/_property.py
```diff
@@ -110,23 +110,14 @@ def uri(self, value: str):
...
         self._data._set_default_value_if_none('tags')
         return self._data.tags
 
     @tags.setter
-    def tags(self, value: Dict[str, 'StructValueType']):
+    def tags(self, value: Dict[str, Any]):
         self._data.tags = value
 
-    @property
-    def _metadata(self) -> Optional[Dict[str, 'StructValueType']]:
```
JoanFM, line 122: why are you changing this here?

docarray/document/mixins/_property.py
```diff
@@ -1,12 +1,12 @@
...
+from typing import TYPE_CHECKING, Dict, List, Optional, Union, Any
 
-if TYPE_CHECKING:  # pragma: no cover
-    from docarray.score import NamedScore
-    from docarray.array.match import MatchArray
-    from docarray.array.chunk import ChunkArray
-    from docarray import DocumentArray
-    from docarray.typing import ArrayType, StructValueType, DocumentContentType
+if TYPE_CHECKING:
+    from ...score import NamedScore
```
JoanFM, line 5: please keep the absolute imports

docarray/document/mixins/_property.py
```diff
@@ -1,12 +1,12 @@
-# auto-generated from /Users/hanxiao/Documents/docarray/scripts/gen_doc_property_mixin.py
-from typing import TYPE_CHECKING, Dict, List, Optional
+# auto-generated from /Users/marco/code/opensource/docarray/scripts/gen_doc_property_mixin.py
```
JoanFM, line 1: remove this auto generated line

JoanFM (changes requested):

docarray/document/mixins/_property.py
marcosbodio: `_property.py` is auto-generated using `scripts/gen_doc_property_mixin.py`. This line is the same as the code in the main branch https://github.com/docarray/docarray/blob/main/docarray/document/mixins/_property.py.

docarray/document/mixins/_property.py
marcosbodio: I can do that: it is a simple change to the script `scripts/gen_doc_property_mixin.py`

docarray/document/mixins/_property.py
marcosbodio: I suggest to keep the comment, because it tells that `_property.py` is auto-generated using `scripts/gen_doc_property_mixin.py`. I can change `scripts/gen_doc_property_mixin.py`, so that the comment will look like `# auto-generated from docarray/scripts/gen_doc_property_mixin.py`. Is that OK?

JoanFM: okey is good

docarray/document/mixins/_property.py
JoanFM: okey,

marcosbodio: @JoanFM sorry, my mistake. I double-checked, and getter and setter for `_metadata` are missing in the version of `_property.py` in my branch because the script `cripts/gen_doc_property_mixin.py` has the following code

```python
    for f in fields(DocumentData):
        if f.name.startswith('_'):
            continue
   ....
```

and therefore it skips all fields starting with `_`. I do not know why you have those getter and setter in the `main` branch. I can remove the `if` in the script, but in that case the script would generate also getter and setter for other fields of `Document` starting with `_`, such as `_reference_doc`. Alternatively, I may had the getter and setter for `_metadata` manually.

What do you prefer? Any other idea?

JoanFM: Let's add them manually

JoanFM (approved):

Issue #738, opened 2022-11-09
bug: elasticsearch query dsl is not supported correctly


girishc13:
The [find](https://github.com/docarray/docarray/blob/main/docarray/array/mixins/find.py#L88) method throws a [ValueError](https://github.com/docarray/docarray/blob/main/docarray/array/mixins/find.py#L153) when the `query` such as Elasticsearch query DSL is provided and the `filter` is not provided. Elasticsearch supports search and filter queries in a single DSL and should be forwarded to the client if the query DSL is provided. Otherwise this blocks users from implementing features available by default.

PR #740, opened 2022-11-09
feat: support find by text and filter

AnneYang720:
This PR is related to #738 

Goals:

- [x] support find by text and filter for ElasticSearch
- [x] also support that for Redis (if needed)
- [x] add tests


tests/unit/array/mixins/test_find.py
```diff
@@ -145,6 +145,65 @@ def test_find_by_text(storage, config, start_storage):
...
+        da.extend(
+            [
+                Document(id=f'{i+10}', tags={'i': i}, text=f'noodles {i}')
+                for i in range(10)
+            ]
+        )
+
+    results = da.find('pizza', filter=filter)
+
+    assert all([int(r.id) < 10 for r in results]) == True
```
JoanFM, line 203: first u need to assrt length of results is larger than 0, otherwise I think the next asserts would still work

tests/unit/array/mixins/test_find.py
```diff
@@ -145,6 +145,65 @@ def test_find_by_text(storage, config, start_storage):
...
+    assert all([r.tags['i'] < 10 for r in results]) == True
```
JoanFM, line 204: no need to do == True

JoanFM (changes requested):

JoanFM (approved):

PR #756, opened 2022-11-11
feat: add tensor type

JohannesMessner:
This adds the typ `Tensor` that wraps `np.ndarray`. It adds protobuf serialization and is a valid Pydantic type, other than that it is just a normal ndarray.


docarray/typing/ndarray.py
```diff
@@ -1,4 +1,4 @@
 import numpy as np
+from .tensor import Tensor
```
JoanFM, line 2: please use global import

docarray/typing/tensor/tensor.py
```diff
@@ -0,0 +1,72 @@
...
+    from pydantic.fields import ModelField
+    from pydantic import BaseConfig, PydanticValueError
+
+from docarray.document.base_node import BaseNode
+from docarray.proto import DocumentProto, NdArrayProto, NodeProto
+
+T = TypeVar('T', bound='Tensor')
+
+
+class Tensor(np.ndarray, BaseNode):
```
JoanFM, line 14: why does it inherit from `np.ndarray`?

PR #762, opened 2022-11-14
docs: add docs about list-like feature

dongxiang123:

Goals:
Since this [PR](https://github.com/docarray/docarray/pull/730/files#diff-ed0be5f19866ce0e879e0ad77f486223c573998977944684dc03499250e5089b) is merge and we need to add document about constructing a new backend in the document
- ...
- ...


JoanFM (changes requested):

docs/advanced/document-store/extend.md
```diff
@@ -84,7 +84,16 @@ You will need to implement the above five functions, which correspond to the log
...
 inherited from the `BaseGetSetDelMixin`. Therefore, you need to implement methods to persist this mapping, in case you 
-want to also persist the ordering of Documents inside the storage.
+want to also persist the ordering of Documents inside the storage. Considered the performance issue brought by list-like
+structure implemented by Offset2id, you can disable this feature by passing a flag when constructing the backend.
+After so, the Offset2id will become a shell and do nothing.
+
+```Python
+from docarray import DocumentArray, Document
+
+da = DocumentArray(storage='qdrant', config={'n_dim':3, 'list_like':False})
```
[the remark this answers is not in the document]
dongxiang123: Make sense

JohannesMessner:
I think we need to explain that this `list_like` parameter has to be part of the `config`, and then how it is passed to the `__init__` of `Offsed2ID`.

dongxiang123:
> I think we need to explain that this `list_like` parameter has to be part of the `config`, and then how it is passed to the `__init__` of `Offsed2ID`.

added examples.

docs/advanced/document-store/extend.md
```diff
@@ -84,7 +84,19 @@ You will need to implement the above five functions, which correspond to the log
 
 Note that DocumentArray maintains an `offset2ids` mapping to allow a list-like behaviour. This mapping is 
 inherited from the `BaseGetSetDelMixin`. Therefore, you need to implement methods to persist this mapping, in case you 
-want to also persist the ordering of Documents inside the storage.
+want to also persist the ordering of Documents inside the storage. Considered the performance cost brought by list-like
+structure implemented by Offset2id, you can disable this feature by passing a flag when constructing the backend.
+After so, the Offset2id will become a shell and do nothing. Then in the `getsetdel.py` you can construct the _offset2ids 
+member variable by passing list_like flag as follows:
```
JohannesMessner, line 90:
```suggestion
want to also persist the ordering of Documents inside the storage. However, the list-like
structure implemented by Offset2id can introduce performance bottlenecks, so you can disable this feature by passing a flag when constructing the backend.

In your backend implementations, in `getsetdel.py`, you have to construct the _offset2ids 
member variable by passing list_like flag as follows:
```

docs/advanced/document-store/extend.md
```diff
@@ -203,6 +225,17 @@ class BackendMixin(BaseBackendMixin):
 
 `MyDocStoreConfig` is a dataclass for containing the configs. You can expose arguments of your document store to this data class and allow users to customize them. In `init_storage` function, you need to parse `config` either from `MyDocStoreConfig` object or a `Dict`.
 
+The `list-like` flag is passed into the config file as follows:
```
JohannesMessner, line 228:
```suggestion
To enable the disabling of list-like features, your configuration should accept the flag `list_like` as follows:
```

docs/advanced/document-store/extend.md
```diff
@@ -203,6 +225,17 @@ class BackendMixin(BaseBackendMixin):
...
+```python
+@dataclass
+class MyDocStoreConfig:
+    config1: str
+    config2: str
+    list - like: True
+    config3: Dict
+    ...
+```
+By default, it is set to be `True`.
```
JohannesMessner, line 238:
```suggestion
By default, this should be set to be `True`.

Further, you have to store the value of this flag in `self._list_like`. Some methods that are handled outside of your control will take the value form there and use it appropriately.
```

JohannesMessner (changes requested):

docs/advanced/document-store/extend.md
```diff
@@ -203,6 +225,17 @@ class BackendMixin(BaseBackendMixin):
 
 `MyDocStoreConfig` is a dataclass for containing the configs. You can expose arguments of your document store to this data class and allow users to customize them. In `init_storage` function, you need to parse `config` either from `MyDocStoreConfig` object or a `Dict`.
 
+The `list-like` flag is passed into the config file as follows:
+```python
+@dataclass
+class MyDocStoreConfig:
+    config1: str
+    config2: str
+    list - like: True
```
JoanFM, line 234:
```suggestion
    list_like: bool
```

JoanFM (changes requested):

docs/advanced/document-store/extend.md
alexcg1, line 90:
```suggestion
want to also persist the ordering of Documents inside the storage. Considering the performance cost brought by the list-like structure implemented by Offset2id, you can disable this feature by passing a flag when constructing the backend.
After this, the Offset2id will become a shell and do nothing. Then in the `getsetdel.py` you can construct the `_offset2ids` 
member variable by passing the `list_like` flag as follows:
```

docs/advanced/document-store/extend.md
```diff
@@ -156,6 +168,16 @@ Most of the interfaces come from Python standard [MutableSequence](https://docs.
 As a reference, to see how we implement for SQLite, check out {class}`~docarray.array.storage.sqlite.seqlike.SequenceLikeMixin`.
 ```
 
+For the reason to support the list-like feature, the list-like APIs should do flag checking only when the offset2id structure is called as follows:
```
alexcg1, line 171:
```suggestion
To support the list-like feature, the list-like APIs should perform flag checking only when the offset2id structure is called as follows:
```

docs/advanced/document-store/extend.md
```diff
@@ -203,6 +225,17 @@ class BackendMixin(BaseBackendMixin):
...
+```python
+@dataclass
+class MyDocStoreConfig:
+    config1: str
+    config2: str
+    list_like: bool
+    config3: Dict
+    ...
+```
+By default, it is set to be `True`.
```
alexcg1, line 238:
```suggestion
By default, it is set to `True`. 
```

alexcg1 (changes requested):

docs/advanced/document-store/extend.md
```diff
@@ -84,7 +84,22 @@ You will need to implement the above five functions, which correspond to the log
 
 Note that DocumentArray maintains an `offset2ids` mapping to allow a list-like behaviour. This mapping is 
 inherited from the `BaseGetSetDelMixin`. Therefore, you need to implement methods to persist this mapping, in case you 
-want to also persist the ordering of Documents inside the storage.
+want to also persist the ordering of Documents inside the storage. However, the list-like
+structure implemented by Offset2id can introduce performance bottlenecks, so you can disable this feature by passing a flag when constructing the backend.
```
JohannesMessner, line 88:
```suggestion
structure implemented by Offset2id can introduce performance bottlenecks, so you can disable this feature by passing a flag when constructing the backend.

In step 4 you will see how you should read in this flag (`list_like`) from the user. Here you have to use it do adapt some operations.
```

JoanFM (approved):

PR #766, opened 2022-11-14
fix: allow backends not to implement list_like disabling

JohannesMessner:
A simple fix that makes it so storage backend implementations don't necessarily have to have a `_list_like` property.
We still want to encourage all backends to implement that feature, but it won't outright fail.
Besides, I think it is a better pattern not to assume the existence of a certain field that is outside the Mixin's control.


docarray/array/storage/base/getsetdel.py
```diff
@@ -129,7 +130,7 @@ def _del_docs_by_mask(self, mask: Sequence[bool]):
     def _del_all_docs(self):
         self._clear_subindices()
         self._clear_storage()
-        self._offset2ids = Offset2ID(list_like=self._list_like)
+        self._offset2ids = Offset2ID(list_like=getattr(self, '_list_like', None))
```
JoanFM, line 133:
```suggestion
        self._offset2ids = Offset2ID(list_like=getattr(self, '_list_like', True))
```

JoanFM: or False

JoanFM (changes requested):

JohannesMessner:
CI seems flaky (something going wrong during pip installing when preparing the tests); failures do not seem related to this PR

JoanFM (approved):

PR #769, opened 2022-11-14
chore(ci): readd commint lint

samsja:

# What this PR do

* commit name lint was disable when the refactoring was in PoC we need to enable it now
* fix some ruff ci part

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


JohannesMessner (commented):
Should we start to not care about bad commit messages anymore as long as the final one that we merge is ok?

samsja:
> Should we start to not care about bad commit messages anymore as long as the final one that we merge is ok?

lets take this decision after we merge this branch in master, my fear is that we will not be able to merge otherwise

JohannesMessner (approved):

PR #772, opened 2022-11-14
refactor: change embedding and base node

samsja:
# context

some small refactoring

### DIFF
--- a/docarray/array/documentarray.py
+++ b/docarray/array/documentarray.py
@@ -1,8 +1,7 @@
 from typing import Iterable, Type
 
-from docarray.document import AnyDocument, BaseDocument
+from docarray.document import AnyDocument, BaseDocument, BaseNode
 from docarray.document.abstract_document import AbstractDocument
-from docarray.typing import BaseNode
 
 from .abstract_array import AbstractDocumentArray
 from .mixins import ProtoArrayMixin
--- a/docarray/document/__init__.py
+++ b/docarray/document/__init__.py
@@ -1,4 +1,5 @@
 from docarray.document.any_document import AnyDocument
+from docarray.document.base_node import BaseNode
 from docarray.document.document import BaseDocument
 
-__all__ = ['AnyDocument', 'BaseDocument']
+__all__ = ['AnyDocument', 'BaseDocument', 'BaseNode']
--- a/docarray/predefined_document/text.py
+++ b/docarray/predefined_document/text.py
@@ -1,7 +1,7 @@
 from typing import Optional
 
 from docarray.document import BaseDocument
-from docarray.typing.ndarray import Embedding, Tensor
+from docarray.typing.embedding import Embedding, Tensor
 
 
 class Text(BaseDocument):
--- a/docarray/typing/__init__.py
+++ b/docarray/typing/__init__.py
@@ -1,6 +1,5 @@
-from docarray.document.base_node import BaseNode
-
-from docarray.typing.ndarray import Embedding, Tensor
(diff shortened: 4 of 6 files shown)


docarray/document/abstract_document.py
```diff
@@ -1,7 +1,16 @@
...
 
+if TYPE_CHECKING:
+    from docarray.document.mixins.proto import ProtoMixin
+
 
 class AbstractDocument(Iterable):
     __fields__: Dict[str, ModelField]
+
+    @classmethod
+    @abstractmethod
```
JoanFM, line 14: this looks very weird. `AbstractDocument` caring about this. Why is this needed for every Mixin?

JoanFM (changes requested):

docarray/document/abstract_document.py
samsja: I believe this is the exact same that having the __fields__ value at this level. This is a common behavior of all document to have this class and it is not only related to proto

docarray/document/abstract_document.py
```diff
@@ -1,7 +1,16 @@
...
+    def _get_nested_document_class(cls, field: str) -> Type['ProtoMixin']:
```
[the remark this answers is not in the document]
samsja: but the return type should be Proto compatible since I called proto related method. Maybe I should move back this one to the proto mixin.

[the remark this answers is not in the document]
JoanFM: I believe so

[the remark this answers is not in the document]
samsja: okay solved now @JoanFM

docarray/document/abstract_document.py
samsja: resolved now

docarray/proto/docarray.proto
```diff
@@ -46,6 +46,15 @@ message NodeProto {
...
     // a sub DocumentArray
     DocumentArrayProto chunks = 5;
+
+    NdArrayProto embedding = 6;
+
+    string any_url = 7;
+
+    string image_url = 8;
+
+    string id = 9;
```
JoanFM, line 56: Id should not be a oneof, it should be an attribute next to the map

docarray/proto/docarray.proto
```diff
@@ -46,6 +46,15 @@ message NodeProto {
 
     // a sub DocumentArray
     DocumentArrayProto chunks = 5;
+
+    NdArrayProto embedding = 6;
+
+    string any_url = 7;
```
JoanFM, line 52: I am not convinced by this. This is going to explode.

JoanFM: I believe the meaning of the Document should be provided by the class, not by the Proto

JoanFM (changes requested):

docarray/proto/docarray.proto
samsja: i am not convince. At the Document level the ID is in the dict as well

samsja: let me think about it

docarray/proto/docarray.proto
samsja: Lets do it for now, in the future we might have to encode this type in a schema field in the proto but for now it is not needed

JohannesMessner (approved):

JohannesMessner:
Let's keep in mind that we might have to move the `id` out of the map at some point, to cleanly differentiate it from other id's that the user might manually put inside the schema/dataclass

PR #776, opened 2022-11-14
chore: fix ci not running on refactoring dev branch

samsja:
# context

CI was set to be triggered on pr or in master. But we are working on feat-rewritte-v2 so we need to add this branch for now to the ci check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - feat-rewritte-v2
   pull_request:
     types: [opened, synchronize]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - feat-rewritte-v2
```
JohannesMessner, line 7:
```suggestion
      - feat-rewrite-v2
```

JohannesMessner (changes requested):

JohannesMessner (approved):

PR #803, opened 2022-11-16
feat: allow da bulk access to return da for document

samsja:
# Context

since https://github.com/docarray/docarray/pull/790 we can already do 

```python

class Mmdoc(BaseDocument):
    text: str

da = DocumentArray[Mmdoc](
    (Mmdoc(text=f'hello') for _ in range(10))
)

assert da.text == 10*['hello']
```

never the less it always return a list of item for the field, even if the field is suppose to contain `Document`. In this case this should return a `DocumentArray` instead of a `List` of `Document`

# What this pr do
When the nested field is a Document it returns a `DocumentArray` instead of a `List` of `Document`

```python

from docarray import Text

class Mmdoc(BaseDocument):
    text: Text

da = DocumentArray[Mmdoc](
    (Mmdoc(text=Text(text='hello')) for _ in range(10))
)

assert da.text == DocumentArray[Text](10*[Text(text='hello'))])
```

### DIFF
--- a/docarray/array/abstract_array.py
+++ b/docarray/array/abstract_array.py
@@ -2,7 +2,6 @@ from abc import abstractmethod
 from typing import Iterable, Type
 
 from docarray.document import BaseDocument
-from docarray.document.abstract_document import AbstractDocument
 
 
 class AbstractDocumentArray(Iterable):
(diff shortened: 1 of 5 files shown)


JohannesMessner:
What happens in the following case?

```python

from docarray import Text

class Mmdoc(BaseDocument):
    text: DocumentArray[Text]

da = DocumentArray[Mmdoc](
    (Mmdoc(text=DocumentArray([Text(text='hello') for _ in range(5)]) for _ in range(10)))
)

assert da.text == ???
```
From reading the code it looks like it would return a list of DocumentArrays.

Is this already supported? If yes, can we have a test for it?

samsja:
Yes this will return a list of document array. Is this the behavior we want to have ? I believe yes

JohannesMessner (approved):

PR #806, opened 2022-11-17
Feat type implement protobuf

dongxiang123:
Goals:

- ...
- ...


JohannesMessner:
@dongxiang123 could you make sure that all the pre-commit "things" (lack, ruff, mypy, etc) are applied?

docarray/typing/chunks.py
```diff
@@ -0,0 +1,21 @@
+from docarray.document.base_node import BaseNode
+from docarray.proto import NodeProto
+
+
+class Chunks(BaseNode):
```
JohannesMessner, line 5: Why do we need this new class?

docarray/typing/nested.py
```diff
@@ -0,0 +1,21 @@
+from docarray.document.base_node import BaseNode
+from docarray.proto import NodeProto
+
+
+class Nested(BaseNode):
```
JohannesMessner, line 5: Same here, is it needed?

docarray/typing/text.py
```diff
@@ -0,0 +1,33 @@
+import string
+from typing import Type
+
+from docarray.document.base_node import BaseNode
+from docarray.proto import NodeProto
+
+
+class Text(BaseNode):
```
JohannesMessner, line 8: I don't think we want this as a type, `Text` is supposed to be a pre-defined Document, not a type

JohannesMessner (changes requested):

docarray/typing/id.py
```diff
@@ -43,3 +41,12 @@ def _to_node_protobuf(self) -> NodeProto:
         :return: the nested item protobuf message
         """
         return NodeProto(id=self)
+
+    @classmethod
+    def from_protobuf(cls: Type[T], pb_msg: 'str') -> T:
+        """
+        read ndarray from a proto msg
```
samsja, line 48:
```suggestion
        read id from a proto msg
```

docarray/typing/tensor/torch_tensor.py
```diff
@@ -82,7 +82,7 @@ def _to_node_protobuf(self: T, field: str = 'torch_tensor') -> NodeProto:
         return NodeProto(**{field: nd_proto})
 
     @classmethod
-    def _read_from_proto(cls: Type[T], pb_msg: 'NdArrayProto') -> 'T':
+    def from_protobuf(cls: Type[T], pb_msg: 'NdArrayProto') -> 'T':
         """
         read ndarray from a proto msg
```
samsja, line 87:
```suggestion
        read torch tensor from a proto msg
```

samsja (commented):
minor comments

samsja (approved):
lgtm, nice refactoring !

JohannesMessner (approved):

PR #807, opened 2022-11-17
hotfix: pull docarray with "username/da-name" format

delgermurun:
Goals:

- To fix `FileNotFoundError: [Errno 2] No such file or directory: '.../username/test_name'` error when pulling docarray with "username/da-name" formatted name

### DIFF
--- a/docarray/array/mixins/io/pushpull.py
+++ b/docarray/array/mixins/io/pushpull.py
@@ -303,10 +303,12 @@ class PushPullMixin:
             from docarray.array.mixins.io.binary import LazyRequestReader
 
             _source = LazyRequestReader(r)
-            if local_cache and os.path.exists(f'{__cache_path__}/{name}'):
-                _cache_len = os.path.getsize(f'{__cache_path__}/{name}')
+
+            cache_file = f'{__cache_path__}/{name.replace("/", "_")}.da'
+            if local_cache and os.path.exists(cache_file):
+                _cache_len = os.path.getsize(cache_file)
                 if _cache_len == _da_len:
-                    _source = f'{__cache_path__}/{name}'
+                    _source = cache_file
 
             r = cls.load_binary(
                 _source,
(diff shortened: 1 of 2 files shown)


docarray/array/mixins/io/pushpull.py
```diff
@@ -303,10 +303,12 @@ def pull(
             from docarray.array.mixins.io.binary import LazyRequestReader
 
             _source = LazyRequestReader(r)
-            if local_cache and os.path.exists(f'{__cache_path__}/{name}'):
-                _cache_len = os.path.getsize(f'{__cache_path__}/{name}')
+
+            cache_path = f'{__cache_path__}/{name.replace("/", "_")}.da'
```
samsja, line 307: so this `cache_path` is a filename right ? Can we rename it to `cache_file` ?

samsja: can we put the replace("/", "_") behavior into a seperate private static method ? 

so that the line looks like

```python
            name = self._migrate_deprecated_url(name)
            cache_path = f'{__cache_path__}/{name}.da'

```

samsja (changes requested):
minor comment

docarray/array/mixins/io/pushpull.py
delgermurun: I don't like replacing the `name` parameter. It could be used somewhere below this section in the future.

Creating static method for it, feels like overkill for me.

Also it is not that old name causing problem, new name causes problem because it includes `/`.
So `migrate deprecated name/url` is not exactly true.

I have following variations:

1)
```python
cache_file_name = name.replace('/', '_')
cache_file = f'{__cache_path__}/{cache_file_name}.da'
```

2)
```python
cache_file = f'{__cache_path__}/{name.replace("/", "_")}.da'
```

3)
```python
cache_file_name = cls._normalize_file_name(name.replace('/', '_')) # or some better name for this
cache_file = f'{__cache_path__}/{cache_file_name}.da'
```

Which one do you prefer? My preference is 2), however, I am open to other ideas as well.

samsja: as u want go for 2 them

delgermurun: okay. So I'll keep it as it is.

samsja (approved):

Issue #775, opened 2022-11-14
Support root_id for storage backends


JohannesMessner:
Some of our users work with deeply nested data, where they perform vector search on some nesting level, but are actually interested in retrieving the root level documents.

In memory this can be solved by traversing the nested structure on the fly, but with a database backend it is not doable: nested levels are only present in serialized form, so one would have to load everything into memory in order to be able to traverse the structure.

To tackle this, we propose the following:

- we create a function `get_root_doc(da, doc)` that returns the root document of `doc`. The implementation could be something similar to this:
```python
def get_root_doc(da, doc):
    root_da_flat = da[...]
    result = doc
    while result.parent_id:
        result = root_da_flat[result.parent_id]
    return result
```
- For storage backends we expose an api that allows you to search by some nesting level, but retrieve documents on the root level: `da.find(..., return_root=True)`
- It works the following way:
   1. when inserting a (batch of) Document(s), it calls `get_root_doc()` on that
   2. It stores the root document's `id` as a separate column in the database
   3. when searching with `return_root=True` it performs a search, then take the result's stored root_id, and returns the root document based on that
   4. The level the user searches on needs to exist as a subindex (this is already the case), and the root level is always properly indexed anyways. The intermediate nesting levels can stay serialized.

PR #808, opened 2022-11-17
feat: support root_id for storage backends

AnneYang720:
This PR implements issue #775 

Goals:

- [ ] codes for supporting root_id
- [ ] add related tests
- [ ] add docs if needed

### DIFF
--- a/docarray/array/mixins/find.py
+++ b/docarray/array/mixins/find.py
@@ -99,6 +99,7 @@ class FindMixin:
         filter: Optional[Dict] = None,
         only_id: bool = False,
         index: str = 'text',
+        return_root: Optional[bool] = False,
         on: Optional[str] = None,
         **kwargs,
     ) -> Union['DocumentArray', List['DocumentArray']]:
@@ -131,9 +132,11 @@ class FindMixin:
 
         :return: a list of DocumentArrays containing the closest Document objects for each of the queries in `query`.
         """
+        from docarray import Document, DocumentArray
+
         index_da = self._get_index(subindex_name=on)
         if index_da is not self:
-            return index_da.find(
+            results = index_da.find(
                 query,
                 metric,
                 limit,
@@ -144,7 +147,16 @@ class FindMixin:
                 index,
                 on=None,
             )
-        from docarray import Document, DocumentArray
+
+            # print(results[:, 'id'])
+
+            if return_root:
+                da = self[results[:, 'tags__root_id']]
+                return da
+
+            # print(da[:,'id'])
+
+            return results
 
         if isinstance(query, dict):
             if filter is None:
--- a/docarray/array/storage/base/seqlike.py
+++ b/docarray/array/storage/base/seqlike.py
@@ -10,6 +10,13 @@ class BaseSequenceLikeMixin(MutableSequence[Document]):
     def _update_subindices_append_extend(self, value):
         if getattr(self, '_subindices', None):
             for selector, da in self._subindices.items():
+
+                if getattr(da, '_config', None):
+                    if da._config.root_id:
+                        for i in range(len(value)):
+                            for doc in DocumentArray(value[i])[selector]:
+                                doc.tags['root_id'] = value[i].id
+
                 docs_selector = DocumentArray(value)[selector]
                 if len(docs_selector) > 0:
                     da.extend(docs_selector)
--- a/tests/unit/array/mixins/test_find.py
+++ b/tests/unit/array/mixins/test_find.py
@@ -949,3 +949,43 @@ def test_find_subindex_multimodal(storage, config):
     assert (closest_docs[0].embedding == np.array([3, 3])).all()
     for d in closest_docs:
         assert d.id.endswith('_2')
+
+
+def test_find_return_root(start_storage):
+    with DocumentArray(
+        storage='redis',
+        config={
+            'n_dim': 128,
+            'index_name': 'idx',
+        },
+        subindex_configs={'@c': {'n_dim': 3}, '@cc': {'n_dim': 3, 'root_id': True}},
+    ) as da:
+        da.extend(
+            [
+                Document(
+                    id=f'{i}',
+                    chunks=[
+                        Document(
+                            id=f'sub{i}_0',
+                            chunks=[
+                                Document(
+                                    id=f'sub2_{i}_0', embedding=np.random.random(3)
+                                )
+                            ],
+                        ),
+                        Document(
+                            id=f'sub{i}_1',
+                            chunks=[
+                                Document(
+                                    id=f'sub2_{i}_1', embedding=np.random.random(3)
+                                )
+                            ],
+                        ),
+                    ],
+                )
+                for i in range(10)
+            ]
+        )
+        res = da.find(np.random.random(3), on='@cc', return_root=True, limit=5)
+        assert len(res) == 5
+        assert all(d.id in [f'{i}' for i in range(10)] for d in res)
--- a/docarray/array/storage/redis/backend.py
+++ b/docarray/array/storage/redis/backend.py
@@ -35,6 +35,7 @@ class RedisConfig:
     block_size: Optional[int] = None
     initial_cap: Optional[int] = None
     columns: Optional[Union[List[Tuple[str, str]], Dict[str, str]]] = None
+    root_id: bool = False
 
 
 class BackendMixin(BaseBackendMixin):


docarray/array/mixins/find.py
```diff
@@ -144,7 +147,16 @@ def find(
...
             )
-        from docarray import Document, DocumentArray
+
+            # print(results[:, 'id'])
+
+            if return_root:
+                da = self[results[:, 'tags__root_id']]
+                return da
+
+            # print(da[:,'id'])
```
JohannesMessner, line 157: commented prints

docarray/array/storage/base/seqlike.py
```diff
@@ -10,6 +10,13 @@ class BaseSequenceLikeMixin(MutableSequence[Document]):
     def _update_subindices_append_extend(self, value):
         if getattr(self, '_subindices', None):
             for selector, da in self._subindices.items():
+
+                if getattr(da, '_config', None):
+                    if da._config.root_id:
+                        for i in range(len(value)):
+                            for doc in DocumentArray(value[i])[selector]:
+                                doc.tags['root_id'] = value[i].id
```
JohannesMessner, line 18: I thank here you can use a built-in notation, something like
```python
subindex_docs = DocumentArray(value[i])[selector]
subindex_docs[:, 'tags__root_id'] = [value[i].id for _ in range(len(subindex_docs))]
```
Not 100% sure, but I think something like that would work so you don't have to loop manually

docarray/array/storage/elastic/backend.py
```diff
@@ -46,6 +46,7 @@ class ElasticConfig:
     ef_construction: Optional[int] = None
     m: Optional[int] = None
     columns: Optional[Union[List[Tuple[str, str]], Dict[str, str]]] = None
+    root_id: bool = False
```
JohannesMessner, line 49: Maybe we should just always store the root_id and not even ask for this config. The overhead seems small (compared to the complete serialized docs that we always store), and it would improve the user experience. What do you think @AnneYang720 @alaeddine-13 @JoanFM ?

JohannesMessner (changes requested):
Looking good so far!
I think there are a fw more thing we have to consider.

- The users might also be interested in the `score` of a retrieved Document. I think if `return_root=True` we should just copy the score from the chunk Document to the root Document before returning it.

- What if the user modifies the subindex directly, like `da['@c'].extend(...)`? Then we can't know the parent id. How do we want to handle this case?

- What about operations like `da[x] = doc`? Then we should also set the root_id, not only for extend and append, so we will have to handle this in `_update_subindex_set()`.

docarray/array/storage/base/seqlike.py
AnneYang720: This cannot work because here https://github.com/docarray/docarray/blob/main/docarray/array/storage/base/getsetdel.py#L301

docarray/array/storage/redis/seqlike.py
```diff
@@ -59,6 +60,17 @@ def _upload_batch(self, batch_of_docs: Iterable['Document']):
...
+        if (
+            self._config.root_id
+            and self._config.index_name.find('_subindex_') != -1
+            and not all([doc.tags.get('root_id', None) for doc in da])
+        ):
+            warnings.warn(
+                "Not all documents have root_id set. This may cause unexpected behavior.",
+                UserWarning,
+            )
+
```
JohannesMessner, line 73: If think to handle this properly, and for all backends at once, this should be moved into `extend()` of the base `seqlike`.
If I understand correctly, the challenge here is to detect if a subindex is extended directly. To make this generally possible we might have to introduce a flag `._is_subindex` or similar; I think that is the better trade-off than doing this check in all backends separately

JohannesMessner (changes requested):

docarray/array/storage/elastic/backend.py
JoanFM: Yes, store by default

docarray/array/storage/base/getsetdel.py
```diff
@@ -200,6 +200,12 @@ def _update_subindices_set(self, set_index, docs):
             _check_valid_values_nested_set(self[set_index], docs)
             if set_index in subindices:
                 subindex_da = subindices[set_index]
+
+                if getattr(subindex_da, '_config', None):
+                    if subindex_da._config.root_id:
+                        for i in range(len(docs)):
+                            docs[i].tags['root_id'] = subindex_da[i].tags['root_id']
```
JoanFM, line 207: use something more private, if possible

JoanFM: like `__docarray__root_id__` or something like this?

JoanFM (changes requested):

docarray/array/storage/base/getsetdel.py
```diff
@@ -200,6 +200,12 @@ def _update_subindices_set(self, set_index, docs):
...
+
```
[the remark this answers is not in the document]
AnneYang720: This branch of the if-block has check `_check_valid_values_nested_set`, which makes sure the new docs has the same series of `[:, id]` as the previous docs. So I think we can know the `root_id` in the case of `da['@c'] = new_da`.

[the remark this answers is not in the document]
AnneYang720: But I think the subindex being deleted is raised by root_id. In the case of `da['@c'] = new_da`, Redis and ElasticSearch will delete the subindex in their database when `_clear_storage()`.

docarray/array/mixins/find.py
```diff
@@ -144,7 +148,37 @@ def find(
...
+                    root_da_flat = da[...]
+                    da = DocumentArray()
+                    for doc in docs:
+                        result = doc
+                        while getattr(result, 'parent_id', None):
+                            result = root_da_flat[result.parent_id]
+                        da.append(result)
+                    return da
+
+                if not isinstance(self, DocumentArrayInMemory) and not all(
```
JohannesMessner, line 165: Can we maybe split the special treatment of in-memory docarray into a separate function/method? No sure if it can be done easily, but if it can, I would do it

JohannesMessner (changes requested):

docarray/array/storage/base/getsetdel.py
```diff
@@ -200,13 +200,27 @@ def _update_subindices_set(self, set_index, docs):
...
+                value = DocumentArray(docs)
+
+                if (
+                    getattr(subindex_da, '_config', None)
+                    and subindex_da._config.root_id
+                ):
+                    for v in value:
+                        for doc in DocumentArray(v)[subindex_selector]:
+                            doc.tags['root_id'] = (
+                                doc.tags['root_id'] if 'root_id' in doc.tags else v.id
```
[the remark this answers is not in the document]
AnneYang720: I think the logic is what you expected.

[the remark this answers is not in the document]
AnneYang720: But I think it's better to have a "private" name.

docarray/array/storage/base/getsetdel.py
```diff
@@ -200,13 +200,27 @@ def _update_subindices_set(self, set_index, docs):
...
         else:  # root level set, update subindices iteratively
             for subindex_selector, subindex_da in subindices.items():
                 old_ids = DocumentArray(self[set_index])[subindex_selector, 'id']
                 del subindex_da[old_ids]
-                subindex_da.extend(DocumentArray(docs)[subindex_selector])
+
+                value = DocumentArray(docs)
+
+                if (
+                    getattr(subindex_da, '_config', None)
```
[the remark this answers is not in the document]
AnneYang720: `DocumentArrayInMemory` doesn't have `_config`.

AnneYang720:
We now face some limitations:

1.  Bool flag `root_id` is added to each storage backend config with default value `True`. But `in-memory` it is always True because it doesn't have a config or flag.
2. With `root_id=True` by default, users may receive many warnings when they want to add or change Documents on nested level directly without setting `_root_id_` or `parent_id` manually. Like `da['@c'].extend(Document())`
3. If we remove the `root_id` flag and always store this infomation by default, and only raise warning or error when user do `da.find(query, on='@c', return_root=True)`, user may need to re-index the database.

docarray/array/storage/base/seqlike.py
```diff
@@ -63,6 +72,12 @@ def __bool__(self):
         return len(self) > 0
 
     def extend(self, values: Iterable['Document'], **kwargs) -> None:
+
+        from docarray.helper import check_root_id
+
+        if getattr(self, '_is_subindex', False):
```
JohannesMessner, line 78: I think this `getattr` is not necessary anymore, we should be able to just to `self._is_subindex()`, right?

docarray/helper.py
```diff
@@ -491,6 +491,42 @@ def _get_array_info(da: 'DocumentArray'):
...
+def check_root_id(da: 'DocumentArray', value: Union['Document', Sequence['Document']]):
+
+    from docarray import Document
+    from docarray.array.memory import DocumentArrayInMemory
+
+    if isinstance(value, Document):
+        value = [value]
+
+    if isinstance(da, DocumentArrayInMemory):
+        if not all([getattr(doc, 'parent_id', None) for doc in value]):
```
JohannesMessner, line 516: this should look into `tags` and not check if the document itself has the root id, right?

JohannesMessner (changes requested):

tests/unit/array/mixins/test_find.py
```diff
@@ -987,3 +987,116 @@ class MMDoc:
     assert (closest_docs[0].embedding == np.array([3, 3])).all()
     for d in closest_docs:
         assert d.id.endswith('_2')
+
```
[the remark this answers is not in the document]
AnneYang720: I added test `test_subindex_root_id` for all the 5 cases I can think of that raise a warning.

docarray/array/storage/base/seqlike.py
```diff
@@ -10,7 +11,15 @@ class BaseSequenceLikeMixin(MutableSequence[Document]):
...
         if getattr(self, '_subindices', None):
             for selector, da in self._subindices.items():
-                docs_selector = DocumentArray(value)[selector]
+
+                value = DocumentArray(value)
+
+                if getattr(da, '_config', None) and da._config.root_id:
+                    for v in value:
+                        for doc in DocumentArray(v)[selector]:
+                            doc.tags['_root_id_'] = v.id
```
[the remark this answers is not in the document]
AnneYang720: The private key can't begin with `__` if we want to use `da[:, 'tags____sth__']` to get the attributes because [here](https://github.com/docarray/docarray/blob/main/docarray/helper.py#L86). So I changed `root_id` to `_root_id_`.

docarray/helper.py
AnneYang720: `DocumentArrayInMemory` can use `parent_id` recursively to get there root level Document, so I didn't store extra `root_id` information in tags for in-memory. [current code](https://github.com/docarray/docarray/pull/808/files#diff-d1f35e3f3298a94b90e3a354c069cf641edad1a1374551b14befde43666c370d)

docarray/array/mixins/find.py
```diff
@@ -144,7 +146,26 @@ def find(
                 index,
                 on=None,
             )
-        from docarray import Document, DocumentArray
+
+            if return_root:
+                from docarray.array.memory import DocumentArrayInMemory
+                from docarray.helper import get_root_docs
+
+                if isinstance(self, DocumentArrayInMemory):
```
JoanFM, line 154: This should be implemented using inheritance.

All `DocumentStores` should inherit from the same Mixin that implements `get_root_docs` in a way, while `inMemory` implements its own way.

This here includes weird cyclic imports

JoanFM (changes requested):

JohannesMessner:
> We now face some limitations:
> 
>     1. Bool flag `root_id` is added to each storage backend config with default value `True`. But `in-memory` it is always True because it doesn't have a config or flag.
> 
>     2. With `root_id=True` by default, users may receive many warnings when they want to add or change Documents on nested level directly without setting `_root_id_` or `parent_id` manually. Like `da['@c'].extend(Document())`
> 
>     3. If we remove the `root_id` flag and always store this infomation by default, and only raise warning or error when user do `da.find(query, on='@c', return_root=True)`, user may need to re-index the database.

To add a little bit more context to 2. and 3.:

The intended use of this feature is for someone to add Documents to the root-level DocumentArray, so that Documents in a subindex can automatically be 'tagged' with a root_id: `da.extend(...)`.
But if a user modifies a subindex dicrectly, we can't know the roots of the documents that they add (there might not even be a root): `da['@c'].extend(...)`. So, we cannot support this pattern.

The question becomes, when do we warn about this behaviour? At insert time, or at `find()` time when the missing `root_id` actually starts to matter? That is what 2. and 3. above refer to.

JoanFM:
> > We now face some limitations:
> > ```
> > 1. Bool flag `root_id` is added to each storage backend config with default value `True`. But `in-memory` it is always True because it doesn't have a config or flag.
> > 
> > 2. With `root_id=True` by default, users may receive many warnings when they want to add or change Documents on nested level directly without setting `_root_id_` or `parent_id` manually. Like `da['@c'].extend(Document())`
> > 
> > 3. If we remove the `root_id` flag and always store this infomation by default, and only raise warning or error when user do `da.find(query, on='@c', return_root=True)`, user may need to re-index the database.
> > ```
> 
> To add a little bit more context to 2. and 3.:
> 
> The intended use of this feature is for someone to add Documents to the root-level DocumentArray, so that Documents in a subindex can automatically be 'tagged' with a root_id: `da.extend(...)`. But if a user modifies a subindex dicrectly, we can't know the roots of the documents that they add (there might not even be a root): `da['@c'].extend(...)`. So, we cannot support this pattern.
> 
> The question becomes, when do we warn about this behaviour? At insert time, or at `find()` time when the missing `root_id` actually starts to matter? That is what 2. and 3. above refer to.

If u have `root_id` True and you are extending in the wrong way, add the `warning`, to me this makes sense. And warn the earliest the better

docarray/array/storage/base/getsetdel.py
```diff
@@ -200,13 +200,25 @@ def _update_subindices_set(self, set_index, docs):
...
```
JohannesMessner, line 214:
```suggestion
                    getattr(subindex_da, '_config', None)  # checks if in-memory da
```

docs/fundamentals/documentarray/subindex.md
```diff
@@ -217,8 +217,7 @@ Document(embedding=np.random.rand(512)).match(da, on='@c')
 ```
 ````
 
-Such a search will return Documents from the subindex. If you are interested in the top-level Documents associated with
-a match, you can retrieve them using `parent_id`:
+Such a search will return Documents from the subindex. If you are interested in the top-level Documents associated with a match, you can retrieve them using `parent_id`:
```
JohannesMessner, line 220: I would remove this sentence, I would say that this is not the recommended usage anymore. Instead people should use the new feature.

docs/fundamentals/documentarray/subindex.md
```diff
@@ -232,3 +231,27 @@ top_image_matches = da.find(query=np.random.rand(512), on='@c')
...
+````
+````{tab} Subindex with chunks
+```python
+top_level_matches = da.find(query=np.random.rand(512), on='@c', return_root=True)
+```
+````
+
+````{admonition} Note
+:class: note
+When you add or change Documents on nested level directly, the `_root_id_` (or `parent_id` for DocumentArrayInMemory) of new Documents should be set manually for `return_root=True` to work:
```
JohannesMessner, line 250:
```suggestion
When you add or change Documents directly on a subindex, the `_root_id_` (or `parent_id` for DocumentArrayInMemory) of new Documents should be set manually for `return_root=True` to work:
```

JohannesMessner (changes requested):

tests/unit/array/mixins/test_find.py
```diff
@@ -987,3 +987,116 @@ class MMDoc:
...
+        ),
+        ('annlite', {'n_dim': 3}, {'@c': {'n_dim': 3}}),
+        ('sqlite', dict(), {'@c': dict()}),
+        ('qdrant', {'n_dim': 3}, {'@c': {'n_dim': 3}}),
+        ('elasticsearch', {'n_dim': 3}, {'@c': {'n_dim': 3}}),
+        ('redis', {'n_dim': 3}, {'@c': {'n_dim': 3}}),
+        ('milvus', {'n_dim': 3}, {'@c': {'n_dim': 3}}),
+    ],
+)
+def test_find_return_root(storage, config, subindex_configs, start_storage):
```
JohannesMessner, line 1011: I would like to see two more test cases (can be in the same test): That the returned root-level documents are actually in the right order, and that the associated scores are copied correctly.

JohannesMessner (approved):

docarray/array/storage/base/backend.py
```diff
@@ -41,6 +42,7 @@ def _init_subindices(
                     config, config_subindex, config_joined, name
                 )
                 self._subindices[name] = self.__class__(config=config_joined)
+                self._subindices[name]._is_subindex = True
```
JoanFM, line 45: this should be handled somehow in the constructor

JoanFM: this shoul dbe another parameter in `__init_storage` with private name `_is_subindex` with `False` as default

JoanFM (changes requested):

PR #811, opened 2022-11-17
feat: image url

JohannesMessner:
Goals:

This adds the type `ImageUrl`.

For now it can:
- load the uri to tensor
- load the uri to bytes

TODO:
- [x] more tests
- [x] appropriate validation
- [x] some small TODOs in the code comments

PR #821, opened 2022-11-21
feat: image url (#811)

dongxiang123:
Goals:

- 
The abstract class should contain the following methods, 1. _to_node_protobuf (Convert Document into a NodeProto protobuf message). 2. from_protobuf (read value from a proto msg) 3. validate (inherited from pydantic about validating data.)
- ...


docarray/typing/abstract_typing.py
```diff
@@ -0,0 +1,33 @@
...
+        cls: Type[T],
+        value: Union[str, int, UUID],
+        field: Optional['ModelField'] = None,
+        config: Optional['BaseConfig'] = None,
+    ) -> T:
+        return
+
+    @abstractmethod
+    def _to_node_protobuf(self) -> NodeProto:
+        return NodeProto(id=self)
```
samsja, line 29: this one is not needed IMO. We should rather make the `AbstractTyping` inherit from `BaseNode`

docarray/typing/abstract_typing.py
```diff
@@ -0,0 +1,33 @@
...
+        yield cls.validate
+
+    @abstractmethod
+    def validate(
+        cls: Type[T],
+        value: Union[str, int, UUID],
+        field: Optional['ModelField'] = None,
+        config: Optional['BaseConfig'] = None,
+    ) -> T:
+        return
```
samsja, line 25:
```suggestion
        ...
```

docarray/typing/abstract_typing.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @abstractmethod
+    def _to_node_protobuf(self) -> NodeProto:
+        return NodeProto(id=self)
+
+    @abstractmethod
+    def from_protobuf(cls: Type[T], pb_msg: T) -> T:
+        return parse_obj_as(cls, pb_msg)
```
samsja, line 33: this cannot be an abstractmethod and define some code. I believe we just need to make is abstract
```suggestion
    @abstractmethod
    def from_protobuf(cls: Type[T], pb_msg: T) -> T:
        ...
```

PR #824, opened 2022-11-21
refactor: add typing abstract class

dongxiang123:

Goals:

We need to create an common interface via an abstract class for our docarray.typing obj

The abstract class should contain the following methods, 1. _to_node_protobuf (Convert Document into a NodeProto protobuf message). 2. from_protobuf (read value from a proto msg) 3. validate (inherited from pydantic about validating data.)

- ...
- ...

### DIFF
--- a/docarray/typing/abstract_type.py
+++ b/docarray/typing/abstract_type.py
@@ -1,0 +1,30 @@
+from abc import abstractmethod
+from typing import Any, Type, TypeVar
+
+from pydantic import BaseConfig
+from pydantic.fields import ModelField
+
+from docarray.document.base_node import BaseNode
+
+T = TypeVar('T')
+
+
+class AbstractType(BaseNode):
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    @abstractmethod
+    def validate(
+        cls: Type[T],
+        value: Any,
+        field: 'ModelField',
+        config: 'BaseConfig',
+    ) -> T:
+        ...
+
+    @classmethod
+    @abstractmethod
+    def from_protobuf(cls: Type[T], pb_msg: T) -> T:
+        ...
--- a/docarray/typing/id.py
+++ b/docarray/typing/id.py
@@ -1,16 +1,15 @@
-from typing import Optional, Type, TypeVar, Union
+from typing import Type, TypeVar, Union
 from uuid import UUID
-
 from pydantic import BaseConfig, parse_obj_as
 from pydantic.fields import ModelField
 
-from docarray.document.base_node import BaseNode
 from docarray.proto import NodeProto
+from docarray.typing.abstract_type import AbstractType
 
 T = TypeVar('T', bound='ID')
 
 
-class ID(str, BaseNode):
+class ID(str, AbstractType):
     """
     Represent an unique ID
     """
@@ -23,8 +22,8 @@ class ID(str, BaseNode):
     def validate(
         cls: Type[T],
         value: Union[str, int, UUID],
-        field: Optional['ModelField'] = None,
-        config: Optional['BaseConfig'] = None,
+        field: 'ModelField',
+        config: 'BaseConfig',
     ) -> T:
 
         try:
--- a/docarray/typing/tensor/tensor.py
+++ b/docarray/typing/tensor/tensor.py
@@ -2,17 +2,18 @@ from typing import TYPE_CHECKING, Any, Dict, List, Tuple, Type, TypeVar, Union, 
 
 import numpy as np
 
+from docarray.typing.abstract_type import AbstractType
+
 if TYPE_CHECKING:
     from pydantic.fields import ModelField
     from pydantic import BaseConfig
 
-from docarray.document.base_node import BaseNode
 from docarray.proto import NdArrayProto, NodeProto
 
 T = TypeVar('T', bound='Tensor')
 
 
-class Tensor(np.ndarray, BaseNode):
+class Tensor(np.ndarray, AbstractType):
     @classmethod
     def __get_validators__(cls):
         # one or more validators may be yielded which will be called in the
--- a/docarray/typing/tensor/torch_tensor.py
+++ b/docarray/typing/tensor/torch_tensor.py
@@ -4,6 +4,8 @@ from typing import TYPE_CHECKING, Any, Dict, Type, TypeVar, Union, cast
 import numpy as np
 import torch  # type: ignore
 
+from docarray.typing.abstract_type import AbstractType
+
 if TYPE_CHECKING:
     from pydantic.fields import ModelField
     from pydantic import BaseConfig
@@ -22,7 +24,7 @@ class metaTorchAndNode(torch_base, node_base):
     pass
 
 
-class TorchTensor(torch.Tensor, BaseNode, metaclass=metaTorchAndNode):
+class TorchTensor(AbstractType, torch.Tensor, metaclass=metaTorchAndNode):
     # Subclassing torch.Tensor following the advice from here:
     # https://pytorch.org/docs/stable/notes/extending.html#subclassing-torch-tensor
     @classmethod
--- a/docarray/typing/url/any_url.py
+++ b/docarray/typing/url/any_url.py
@@ -3,8 +3,8 @@ from typing import TYPE_CHECKING, Type, TypeVar
 from pydantic import AnyUrl as BaseAnyUrl
 from pydantic import errors, parse_obj_as
 
-from docarray.document.base_node import BaseNode
 from docarray.proto import NodeProto
+from docarray.typing.abstract_type import AbstractType
 
 if TYPE_CHECKING:
     from pydantic.networks import Parts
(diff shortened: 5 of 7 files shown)


docarray/typing/abstract_typing.py
```diff
@@ -0,0 +1,26 @@
+from abc import ABC, abstractmethod
+from typing import Any, Optional, Type, TypeVar, Union
+
+from pydantic import BaseConfig
+from pydantic.fields import ModelField
+
+T = TypeVar("T")
+
+
+class AbstractTyping(ABC):
```
samsja, line 10: can we inhereit from BaseNode as well ?

docarray/typing/url/any_url.py
```diff
@@ -26,6 +29,17 @@ def _to_node_protobuf(self) -> NodeProto:
...
+    def validate(
+        cls: Type[T],
+        value: Union[Any, Any, Any],
+        field: Optional['ModelField'] = None,
+        config: Optional['BaseConfig'] = None,
+    ) -> T:
+        return super().validate(
+            value, cast(ModelField, field), cast(BaseConfig, config)
+        )
+
```
samsja, line 42: we don't need to define this method here, the inheritance is enough

docarray/typing/abstract_type.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC, abstractmethod
+from typing import Any, Optional, Type, TypeVar, Union
+
+from pydantic import BaseConfig
+from pydantic.fields import ModelField
+
+from docarray.document import BaseNode
+
+T = TypeVar("T")
```
samsja, line 9:
```suggestion
T = TypeVar('T', bound='AbstractType')
```

docarray/typing/abstract_type.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class AbstractType(BaseNode):
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
```
JohannesMessner, line 15: If it is abstract it shouldn't offer this implementation, right? And inherit from `ABC`?

docarray/typing/url/image_url.py
```diff
@@ -36,9 +36,9 @@ def _to_node_protobuf(self) -> NodeProto:
...
         cls: Type[T],
-        value: Union[T, np.ndarray, Any],
-        field: 'ModelField',
-        config: 'BaseConfig',
+        value: Union[Any, Any, Any],
+        field: ModelField,
+        config: BaseConfig
     ) -> T:
 
         url = super().validate(value, field, config)  # basic url validation
```
samsja, line 44: we don't need this one I believe

samsja (changes requested):

docarray/typing/abstract_type.py
samsja: I think it is okay to offer this implementation since this is a common pattern even though it is abstract. An abstract class mean only one of method is abstract not all of them. ABC makes sense though

docarray/typing/tensor/tensor.py
```diff
@@ -23,9 +24,9 @@ def __get_validators__(cls):
     @classmethod
     def validate(
         cls: Type[T],
-        value: Union[T, np.ndarray, List[Any], Tuple[Any], Any],
-        field: 'ModelField',
-        config: 'BaseConfig',
+        value: Union[Any, Any, Any],
```
samsja, line 27: same here lets keep the old line other we lose info

dongxiang123: Because if do so, the mypy will raise the error about the incompatibility between abstract class and inherided type.

docarray/typing/tensor/torch_tensor.py
```diff
@@ -35,9 +37,9 @@ def __get_validators__(cls):
     @classmethod
     def validate(
         cls: Type[T],
-        value: Union[T, np.ndarray, Any],
-        field: 'ModelField',
-        config: 'BaseConfig',
+        value: Union[Any, Any, Any],
```
samsja, line 40: same here

docarray/typing/abstract_type.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @classmethod
+    @abstractmethod
+    def validate(
+        cls: Type[T],
+        value: Union[Any, Any, Any],
```
samsja, line 21:
```suggestion
        value: Any
```

docarray/typing/url/image_url.py
```diff
@@ -33,23 +31,6 @@ def _to_node_protobuf(self) -> NodeProto:
...
-
-        url = super().validate(value, field, config)  # basic url validation
-        has_image_extension = any(url.endswith(ext) for ext in IMAGE_FILE_FORMATS)
-        if not has_image_extension:
-            raise ValueError(
-                f'Image URL must have one of the following extensions:'
-                f'{IMAGE_FILE_FORMATS}'
-            )
-        return cls(str(url), scheme=None)
-
```
JohannesMessner, line 52: @dongxiang123 we cannot just remove this code, it is functional and needed. It validates that the image url actually points to a valid image file

JohannesMessner (changes requested):

docarray/typing/abstract_type.py
```diff
@@ -0,0 +1,30 @@
+from abc import abstractmethod
+from typing import Any, Type, TypeVar, Union
+
+from pydantic import BaseConfig
+from pydantic.fields import ModelField
+
+from docarray.document import BaseNode
```
samsja, line 7:
```suggestion
from docarray.document.base_node import BaseNode
```

This will fix the circular import problem. Otherwise you need to import Document at this level and this create the problem

docarray/typing/abstract_type.py
```diff
@@ -0,0 +1,30 @@
...
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    @abstractmethod
+    def validate(
+        cls: Type[T],
+        value: Any,
+        field: ModelField,
+        config: BaseConfig,
```
samsja, line 23:
```suggestion
        field: 'ModelField',
        config: 'BaseConfig',
```

these two object are only imported for type testing therefore we need to put them into ' ' otherwise python complain

docarray/typing/id.py
```diff
@@ -23,8 +22,8 @@ def __get_validators__(cls):
     def validate(
         cls: Type[T],
         value: Union[str, int, UUID],
-        field: Optional['ModelField'] = None,
-        config: Optional['BaseConfig'] = None,
+        field: ModelField,
+        config: BaseConfig,
```
samsja, line 26:
```suggestion
        field: 'ModelField',
        config: 'BaseConfig',
```

same here

docarray/typing/tensor/tensor.py
```diff
@@ -24,8 +25,8 @@ def __get_validators__(cls):
     def validate(
         cls: Type[T],
         value: Union[T, np.ndarray, List[Any], Tuple[Any], Any],
-        field: 'ModelField',
-        config: 'BaseConfig',
+        field: ModelField,
+        config: BaseConfig,
```
samsja, line 29:
```suggestion
        field: 'ModelField',
        config: 'BaseConfig',
```

docarray/typing/tensor/torch_tensor.py
```diff
@@ -36,8 +38,8 @@ def __get_validators__(cls):
     def validate(
         cls: Type[T],
         value: Union[T, np.ndarray, Any],
-        field: 'ModelField',
-        config: 'BaseConfig',
+        field: ModelField,
+        config: BaseConfig,
```
samsja, line 42:
```suggestion
        field: 'ModelField',
        config: 'BaseConfig',
```

docarray/typing/url/image_url.py
samsja: probably a mixup with removing the validate from any_url :thinking: lets bring back this function

docarray/typing/url/image_url.py
```diff
@@ -1,17 +1,18 @@
...
+from pydantic.fields import ModelField
+
 from docarray.proto import NodeProto
 from docarray.typing.url.any_url import AnyUrl
 from docarray.typing.url.helper import _uri_to_blob
 
 if TYPE_CHECKING:
     import PIL
-    from pydantic import BaseConfig
-    from pydantic.fields import ModelField
```
samsja, line 14: we need to bring back does line for the TYPE_CHECKING

docarray/typing/url/image_url.py
```diff
@@ -1,17 +1,18 @@
 import io
 import struct
-from typing import TYPE_CHECKING, Any, Optional, Tuple, Type, TypeVar, Union
+from typing import TYPE_CHECKING, Optional, Tuple, TypeVar, Union, Type, Any
 
 import numpy as np
 
+from pydantic import BaseConfig
+from pydantic.fields import ModelField
+
```
samsja, line 9: lets move this to TypeChecking (where it was before) as we don't need them at runtime but only when doing type checking

samsja (approved):
lgtm

PR #826, opened 2022-11-22
feat: add max_rel_per_label to support recall for labeled data

guenthermi:
Signed-off-by: Michael Guenther <guenthermi50@gmail.com>

Goals:

- Support recall@k and F1 measure@k for labeled datasets.

---

For labeled datasets it is not trivial to calculate metrics like recall and F1 measure, which require the knowledge of the number of relevant documents in the document collection for each query since neither the whole set of relevant documents nor the number of documents with a specific label is provided to the `evaluate` function.

To enable the calculation of recall and F1 measure, this PR
- adds a parameter `max_rel_per_label: Dict` to the evaluate function which provides the number of relevant documents for each label, i.e., the number of documents in the collection with this label.
- calculates those label counts for `max_rel_per_label` in the `embed_and_evaluate_function`.

---

**Code Example**:

```python
# example DocumentArray with matches and labels for evaluation 
da = DocumentArray([Document(text=str(i), tags={'label': i}) for i in range(3)])
for d in da:
  d.matches = da
# each label occurs one time in the data collection (not provided to the evaluate function)
max_rel_per_label = {i: 1 for i in range(3)}
# evaluate matches
metrics = da.evaluate(['recall_at_k'], max_rel_per_label=max_rel_per_label)
print(metrics)
```
```text
{'recall_at_k': 1.0}
```


docarray/array/mixins/evaluation.py
```diff
@@ -161,7 +166,16 @@ def evaluate(
         results = defaultdict(list)
         caller_max_rel = kwargs.pop('max_rel', None)
         for d, gd in zip(self, ground_truth):
-            max_rel = caller_max_rel or len(gd.matches)
+            if caller_max_rel:
```
gmastrapas, line 169: I think you need to refactor the if else logic here a bit
```
for d, gd in zip(self, ground_truth):
    if caller_max_rel:
        max_rel = caller_max_rel

    if ground_truth_type == 'labels':
        if max_rel_per_label:
            max_rel = max_rel_per_label.get(d.tags[label_tag], None)
            if max_rel is None:
                raise ValueError(
                    '`max_rel_per_label` misses the label ' + str(d.tags[label_tag])
                )
        else:
            raise ValueError('max_rel is required or something')
    else:
        max_rel = len(gd.matches)
```

guenthermi: I think it is correct, that `caller_max_rel` is used if the user provides a `max_rel` attribute explicitly.

guenthermi: This exception when `max_rel` is not set should also not be their because most of the metrics do not require `max_rel`, but setting it to None might be better than setting it to `len(gd.matches)`. I will change this.

LMMilliken (approved):

docarray/array/mixins/evaluation.py
```diff
@@ -438,12 +456,20 @@ def fuse_matches(global_matches: DocumentArray, local_matches: DocumentArray):
                     new_matches.append(m)
                 query_data[doc.id, 'matches'] = new_matches
 
+        if ground_truth and label_tag in ground_truth[0].tags:
+            max_rel_per_label = dict(Counter([d.tags[label_tag] for d in ground_truth]))
+        elif not ground_truth and label_tag in query_data[0].tags:
+            max_rel_per_label = dict(Counter([d.tags[label_tag] for d in query_data]))
+        else:
+            max_rel_per_label = None
```
bwanglzu, line 464: i don't understand, `max_rel_per_label` is a variable you passed into the function, then what is this `max_rel_per_label`?

bwanglzu: okay i see, these are two functions

bwanglzu: can you elaberate a bit the naming, `max_rel_per_label`? why not `num_relevant_documents_per_label` or something like that?

bwanglzu (changes requested):
left some comment

docarray/array/mixins/evaluation.py
guenthermi: ok done

JoanFM (changes requested):
I would like to see some changes in Documentation

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -215,6 +215,50 @@ da_prediction.evaluate(
 
 In this case, the keyword argument `k` is passed to all metric functions, even though it does not fulfill any specific function for the calculation of the reciprocal rank.
 
+### The max-rel parameter
```
alexcg1, line 218:
```suggestion
### The max_rel parameter
```

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -215,6 +215,50 @@ da_prediction.evaluate(
...
+
+Some metric functions shown in the table above require a `max_rel` parameter.
+This parameter should be set to the number of relevant documents in the document collection.
```
alexcg1, line 221:
```suggestion
This parameter should be set to the number of relevant Documents in the Document collection.
```

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -215,6 +215,50 @@ da_prediction.evaluate(
...
+Without the knowledge of this number, metrics like `recall_at_k` and `f1_score_at_k` can be calculated.
```
alexcg1, line 222: Can or cannot?

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -215,6 +215,50 @@ da_prediction.evaluate(
...
+
+In the `evaluate` function, one can provide a keyword argument `max_rel`, which is then used for all queries.
```
alexcg1, line 224: Change all "one" to "you"

docs/fundamentals/documentarray/evaluation.md
```diff
@@ -215,6 +215,50 @@ da_prediction.evaluate(
...
+
+```python
+da_prediction.evaluate(ground_truth=da_original, metrics=['recall_at_k'], max_rel=9)
+```
+
+```text
+{'recall_at_k': 1.0}
+```
+
+Since all relevant documents are in the matches, the recall is one.
```
alexcg1, line 236: Capitalize Documents throughout text

alexcg1 (changes requested):

alexcg1 (approved):
LGTM :+1:

JoanFM (approved):

Issue #829, opened 2022-11-22
Subindex find is broken by da['@c'] = new_da 


AnneYang720:
After using `da['@c'] = new_da` with Redis or ElasticSearch storage backends, find on subindex level will no longer work. MRE is below.

```python
from docarray import Document, DocumentArray
import numpy as np

with DocumentArray(
    storage='elasticsearch', # or redis
    config={
        'n_dim': 128,
    },
    subindex_configs={'@c': {'n_dim': 3}},
) as da:
    da.extend(
        [
            Document(
                id=f'{i}',
                chunks=[
                    Document(id=f'sub{i}_0', embedding=np.random.random(3)),
                    Document(id=f'sub{i}_1', embedding=np.random.random(3)),
                ],
            )
            for i in range(1)
        ]
    )
    res = da.find(np.random.random(3), on='@c') # this works
    
    da['@c'] = [Document(id='sub0_0', embedding=np.random.random(3)), Document(id='sub0_1', embedding=np.random.random(3))]
    res = da.find(np.random.random(3), on='@c') # this fails
```

This reason is `da['@c'] = new_da`  will call `subindex_da.clear()` in `_update_subindices_set` and then `_clear_storage`. The `_clear_storage` of Redis and ElasticSearch deletes the index in database.

PR #837, opened 2022-11-23
fix: rebuild index when clear storage

AnneYang720:

Goals:

This PR is related to issue #829  and will rebuild index for Redis and ElasticSearch when `_clear_storage` is called.

- [x] change for Redis
- [x] change for ElasticSearch
- [x] add related test

### DIFF
--- a/docarray/array/storage/elastic/backend.py
+++ b/docarray/array/storage/elastic/backend.py
@@ -93,8 +93,14 @@ class BackendMixin(BaseBackendMixin):
         self._config.columns = self._normalize_columns(self._config.columns)
 
         self.n_dim = self._config.n_dim
-        self._client = self._build_client()
         self._list_like = self._config.list_like
+
+        self._client = Elasticsearch(
+            hosts=self._config.hosts,
+            **self._config.es_config,
+        )
+
+        self._build_index()
         self._build_offset2id_index()
 
         # Note super()._init_storage() calls _load_offset2ids which calls _get_offset2ids_meta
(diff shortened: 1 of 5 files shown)


docarray/array/storage/elastic/backend.py
```diff
@@ -166,22 +172,15 @@ def _build_schema_from_elastic_config(self, elastic_config):
             ] = index_options
         return da_schema
 
-    def _build_client(self):
```
JohannesMessner, line 169: I think we can keep this method and do `self._client = self._build_client()`, while still keeping `_build_index()`? That way we don't need to repeat these lines down in line 289

docarray/array/storage/redis/backend.py
```diff
@@ -84,7 +84,14 @@ def _init_storage(
         self._doc_prefix = config.index_name + ':'
         self._config.columns = self._normalize_columns(self._config.columns)
 
-        self._client = self._build_client()
+        self._client = Redis(
```
JohannesMessner, line 87: same here, i think we can keep the _buid_client() method

docarray/array/storage/redis/backend.py
```diff
@@ -94,25 +101,19 @@ def _init_storage(
...
-        client = Redis(
-            host=self._config.host,
-            port=self._config.port,
-            **self._config.redis_config,
-        )
-
-        if self._config.update_schema:
-            if self._config.index_name.encode() in client.execute_command('FT._LIST'):
-                client.ft(index_name=self._config.index_name).dropindex()
+    def _build_index(self, rebuild: bool = False):
```
JohannesMessner, line 104: I am not familiar with the inner workings of redis, could you very briefly explain why this `rebuild` argument is needed?

AnneYang720: RedisConfig has boolean `update_schema` to decide whether to update Redis Search index schema during initialization. I want the function `_build_index` to be used by both `_init_storage` and rebuild index of `_clear_storage`. This `rebuild` argument is for the latter situation.

JohannesMessner (approved):

JoanFM (approved):

PR #840, opened 2022-11-24
fix: add support independent len calculation

dongxiang123:

Goals:

- ...
As reported by users, after disabling the list-like flag in Docarray backends, the len should still work, because otherwise the redis backend will fail. Therefore it is necessary to add this functionality.

- ...

### DIFF
--- a/docarray/array/storage/redis/seqlike.py
+++ b/docarray/array/storage/redis/seqlike.py
@@ -26,8 +26,12 @@ class SequenceLikeMixin(BaseSequenceLikeMixin):
 
         :return: the length of this :class:`DocumentArrayRedis` object
         """
-        try:
+        if self._list_like:
             return len(self._offset2ids)
+        try:
+            lua_script = f'return #redis.pcall("keys", "{self._config.index_name}:*")'
+            cmd = self._client.register_script(lua_script)
+            return cmd()
         except:
             return 0
 
--- a/tests/unit/document/test_disable_offset2id.py
+++ b/tests/unit/document/test_disable_offset2id.py
@@ -25,9 +25,14 @@ def docs():
 )
 def test_disable_offset2id(docs, storage, config, start_storage):
     if config:
-        da = DocumentArray(docs, storage=storage, config=config)
+        da = DocumentArray(storage=storage, config=config)
     else:
-        da = DocumentArray(docs, storage=storage)
+        da = DocumentArray(storage=storage)
+
+    assert len(da) == 0
+
+    da.extend(docs)
+    assert len(da) == 3
 
     with pytest.raises(ValueError):
         da[0]
--- a/docarray/array/storage/milvus/getsetdel.py
+++ b/docarray/array/storage/milvus/getsetdel.py
@@ -40,6 +40,8 @@ class GetSetDelMixin(BaseGetSetDelMixin):
                 )
             sorted_res = sorted(res, key=lambda k: int(k['offset']))
             self._offset2ids = Offset2ID([r['document_id'] for r in sorted_res])
+        else:
+            self._offset2ids = Offset2ID([], list_like=self._list_like)
 
     def _save_offset2ids(self):
         if self._list_like:


docarray/array/storage/base/helper.py
```diff
@@ -17,10 +18,14 @@ def get_id(self, idx):
...
         if self._list_like:
             self.ids.append(data)
+        else:
+            self.ids_count += 1
 
     def extend(self, data):
         if self._list_like:
             self.ids.extend(data)
+        else:
+            self.ids_count += 1
```
AnneYang720, line 28:
```suggestion
            self.ids_count += len(da)
```

AnneYang720 (commented):
You can change the previous test to
```python
@pytest.fixture(scope='function')
def docs():
    d1 = Document(id='0', embedding=np.array([10, 0]))
    d2 = Document(id='1', embedding=np.array([0, 10]))
    d3 = Document(id='2', embedding=np.array([-10, -10]))
    yield d1, d2, d3


@pytest.mark.parametrize(
    'storage,config',
    [
        ('weaviate', {'n_dim': 2, 'list_like': False}),
        ('annlite', {'n_dim': 2, 'list_like': False}),
        ('qdrant', {'n_dim': 2, 'list_like': False}),
        ('elasticsearch', {'n_dim': 2, 'list_like': False}),
        ('redis', {'n_dim': 2, 'list_like': False}),
        ('milvus', {'n_dim': 2, 'list_like': False}),
    ],
)
def test_disable_offset2id(docs, storage, config, start_storage):
    if config:
        da = DocumentArray(storage=storage, config=config)
    else:
        da = DocumentArray(storage=storage)
    
    assert len(da) == 0
    
    da.extend(docs)
    assert len(da) == 3
    
    del da['0']
    assert len(da) == 2

    with pytest.raises(ValueError):
        da[0]
```

And since you have changed the extend function of `Offset2ID`, the `_extend` function for storage backends should also be changed, like [for Redis](https://github.com/docarray/docarray/blob/main/docarray/array/storage/redis/seqlike.py#L64), `if self._list_like:` should be removed.

AnneYang720 (commented):
There are three more things:
- Since you have changed the extend function of `Offset2ID`, the `_extend` function for storage backends should also be changed, like [for Redis](https://github.com/docarray/docarray/blob/main/docarray/array/storage/redis/seqlike.py#L64), `if self._list_like:` should be removed.
- The `_del_doc` and `_del_docs` [function](https://github.com/docarray/docarray/blob/main/docarray/array/storage/base/getsetdel.py#L105) will call `self._offset2ids.delete_by_id(_id)` and `self._del_doc_by_id(_id)`. `self._del_doc_by_id(_id)` won't delete the document in storage if the key doesn't exist, but the `self.ids_count -= 1` of offset2id is executed anyway. You can notice this issue if you change the `del da['0']` to `del da['any_id']` in the script above. 
- The test for milvus fails with `AttributeError: 'DocumentArrayMilvus' object has no attribute '_offset2ids'`

docarray/array/storage/redis/seqlike.py
```diff
@@ -27,7 +27,7 @@ def __len__(self):
         :return: the length of this :class:`DocumentArrayRedis` object
         """
         try:
-            return len(self._offset2ids)
```
JoanFM, line 30: u can try to see if `offset2ids` exists and then default to this, it may be faster

docarray/array/storage/redis/seqlike.py
```diff
@@ -27,7 +27,7 @@ def __len__(self):
...
+            return len(self._client.ft(index_name=self._config.index_name).info())
```
JoanFM, line 30: how do u know everyone has this method?

JoanFM (changes requested):

docarray/array/storage/redis/seqlike.py
```diff
@@ -26,8 +26,12 @@ def __len__(self):
 
         :return: the length of this :class:`DocumentArrayRedis` object
         """
-        try:
+        if self._list_like:
             return len(self._offset2ids)
+        try:
+            lua_script = "return #redis.pcall('keys', '*')"
```
AnneYang720, line 32:
```suggestion
            lua_script = f'return #redis.pcall("keys", "{self._config.index_name}:*")'
```

AnneYang720 (changes requested):

tests/unit/document/test_disable_offset2id.py
```diff
@@ -37,20 +42,6 @@ def test_offset2ids(docs, start_storage):
     list_like = False
     ids = [str(i) for i in range(3)]
     offset2id = Offset2ID(ids, list_like)
-    len_before = len(offset2id)
-    offset2id.append("4")
```
JoanFM, line 41: why is this removed?

JoanFM (changes requested):

tests/unit/document/test_disable_offset2id.py
dongxiang123: Because the Offset2id class is not responsible for keeping the total length of Docarray list, it is no use to keep the functionality here.

JoanFM (changes requested):

JoanFM (approved):

PR #845, opened 2022-11-24
Add support to load multi page tiff files into chunks

anna-charlotte:
A tiff file can contain multiple images. Currently, only the first image is being stored to a Documents `.tensor` when calling `.load_uri_to_image_tensor()`. We want to add support of multi-page tiff files now. In the case of a multi page tiff file, we want to save them to the Documents `.chunks` so that each image is wrapped in its own Document.

- [ ] add support for multi page tiff files


samsja:
this looks good.  Should the method still be named `.load_uri_to_image_tensor()` ? Technically we don't put it into the tensor in your case

samsja:
> > this looks good. Should the method still be named `.load_uri_to_image_tensor()` ? Technically we don't put it into the tensor in your case
> 
> Yes, I agree, this way the name doesn't really fit. Is changing the functions name not too much of a change though? Otherwise, I can change it to something like `.load_uri_to_images`

changing the function name is too big of a change. Maybe it is okay like this, Lets see if the test pass and we can rediscuss the API

docarray/document/mixins/image.py
```diff
@@ -359,26 +374,53 @@ def _to_image_tensor(
...
-        return np.array(raw_img)
+        n_frames = raw_img.n_frames
+    except AttributeError:
+        n_frames = 1
+
+    if n_frames > 1:
+
+        frames = []
+        for i in range(raw_img.n_frames):
+            raw_img.seek(i)
```
samsja, line 404: what is this doing ?

samsja: okay makes sense

samsja (approved):

docs/datatypes/image/index.md
```diff
@@ -33,6 +33,26 @@ print(d.tensor, d.tensor.shape)
 (618, 641, 3)
 ```
 
+DocArray also supports loading multi-page tiff files. In this case, the image tensors are being stored to the `.tensor` attributes at chunk level instead of top level.
```
alexcg1, line 36:
```suggestion
DocArray also supports loading multi-page tiff files. In this case, the image tensors are stored to the `.tensor` attributes at the chunk-level instead of the top-level.
```

alexcg1 (changes requested):

alexcg1 (approved):
LGTM :+1:

samsja (approved):

Issue #612, opened 2022-10-12
Add full text search support to qdrant backend


alaeddine-13:
Qdrant added support for full text search in [v0.10.0](https://github.com/qdrant/qdrant/releases/tag/v0.10.0) : https://qdrant.tech/documentation/indexing/#full-text-index
Support this feature in DocArray `find` interface

JoanFM:
Note, this is not real text search, this is text based filtering.

JoanFM:
This is blocked by the update on `qdrant versioning`

PR #849, opened 2022-11-25
feat: support find_by_text for qdrant

AnneYang720:

Goals:

This PR is related to issue #612 

- [ ] `_find_by_text`
- [ ] add related codes

### DIFF
--- a/docarray/array/storage/qdrant/backend.py
+++ b/docarray/array/storage/qdrant/backend.py
@@ -14,6 +14,7 @@ from typing import (
 
 import numpy as np
 from qdrant_client import QdrantClient
+from qdrant_client.http import models
 from qdrant_client.http.models.models import (
     Distance,
     CreateCollection,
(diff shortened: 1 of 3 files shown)


JoanFM (changes requested):
I think it needs to be clarified that the `full_text_search` here is not the same as in Elastic Search. I am not sure this implementation is correct actually.

According to `Qdrant`, this is just another `filter` conditions but not a full text search powered by `BM25`. 

I would reject this PR and reimplement this as just raising the version and documenting that these kind of filters can be used

JoanFM:
> > I think it needs to be clarified that the `full_text_search` here is not the same as in Elastic Search. I am not sure this implementation is correct actually.
> > According to `Qdrant`, this is just another `filter` conditions but not a full text search powered by `BM25`.
> > I would reject this PR and reimplement this as just raising the version and documenting that these kind of filters can be used
> 
> I think the way this implementation does the indexing etc. is correct, from briefly looking at the qdrant documentation it seems like this kind of filter requires a dedicated index to be built for it. But I agree that we should probably expose this feature through our filtering API, not expose it on the same level as vector similarity `find()`. Additional reference: https://qdrant.tech/documentation/filtering/#full-text-match

Exactly, I think it should be expose via `filter`

JohannesMessner (approved):

Issue #857, opened 2022-11-28
Milvus _get_docs_by_ids cannot deal with duplicated id


AnneYang720:
MRE is below:
```python
from docarray import Document, DocumentArray

with DocumentArray(
    storage='milvus',
    config={'n_dim': 3,},
) as da:
    da.extend([Document(id=f'{i}',) for i in range(10)])

ids = ['1', '2', '3', '1', '2', '3']
print(da[ids]) # this will print <DocumentArray (length=3) at xxx>
print(da[ids][:, 'id']) # this will print ['1', '2', '3']
print(len(da[ids])) # this will print 3
```
The possible reason is the milvus [api](https://github.com/docarray/docarray/blob/main/docarray/array/storage/milvus/getsetdel.py#L67) in `_get_docs_by_ids`.

PR #859, opened 2022-11-28
fix: milvus _get_docs_by_ids

AnneYang720:
Goals:

This PR is related to issue #857 

- [x] codes
- [x] add related test


docarray/array/storage/milvus/getsetdel.py
```diff
@@ -72,8 +72,7 @@ def _get_docs_by_ids(self, ids: 'Iterable[str]', **kwargs) -> 'DocumentArray':
                 raise KeyError(f'No documents found for ids {ids}')
             docs.extend(self._docs_from_query_response(res))
         # sort output docs according to input id sorting
-        id_to_index = {id_: i for i, id_ in enumerate(ids)}
-        return DocumentArray(sorted(docs, key=lambda d: id_to_index[d.id]))
+        return DocumentArray([docs[d] for d in ids])
```
JohannesMessner, line 75: Can we keep the dict-based approach, but have every `id_` point to a list of positions, and leverage that in `key=...` somehow?
The reason it was done this way is performance, we want to gather the id-to-position mapping only once, and then delegate everything else to `sorted()`, which leverages a fist implementation in C.

JohannesMessner (changes requested):
Good catch - just one small performance concern.

docarray/array/storage/milvus/getsetdel.py
AnneYang720: I can't think of a python built-in function. `sort` doesn't make the list longer than the original one.

JohannesMessner: Ok I see, then let's keep it.

JohannesMessner (approved):

PR #880, opened 2022-11-30
feat: store key frame indices when loading video tensor from uri

anna-charlotte:
Goals:

When calling `.load_uri_to_video_tensor` we want to additionally store the indices of the key frames in the Document's `.tags['key_frame_indices']`. This makes it easier to access selected frames, e.g.:
```python
from docarray import Document

d = Document(uri='toy.mp4')
d.load_uri_to_video_tensor()
first_scene = d.tensor[d.tags['key_frame_indices'][0]: d.tags['key_frame_indices'][1]]

print(first_scene.shape)
```
```
(95, 320, 176, 3)
``` 
The indices will be stored independently of the value of the `key_frames` parameter.

- [x] store key frame indices in `.tags`

### DIFF
--- a/docarray/document/mixins/video.py
+++ b/docarray/document/mixins/video.py
@@ -100,7 +100,7 @@ class VideoDataMixin:
     def load_uri_to_video_tensor(
         self: 'T', only_keyframes: bool = False, **kwargs
     ) -> 'T':
-        """Convert a :attr:`.uri` to a video ndarray :attr:`.tensor` and store the indices of the key frames in :attr:`.tags` if only_keyframes=False.
+        """Convert a :attr:`.uri` to a video ndarray :attr:`.tensor` and store the indices of the keyframes in :attr:`.tags` if only_keyframes=False.
 
         :param only_keyframes: only keep the keyframes in the video
         :param kwargs: supports all keyword arguments that are being supported by av.open() as
@@ -123,7 +123,7 @@ class VideoDataMixin:
                 self.tensor = np.moveaxis(np.stack(frames), 1, 2)
                 return self
 
-        key_frame_indices = []
+        keyframe_indices = []
         input = av.open(self.uri, **kwargs)
         for i, packet in enumerate(input.demux(video=0)):
 
@@ -131,10 +131,10 @@ class VideoDataMixin:
                 continue
 
             if packet.stream.type == 'video' and packet.is_keyframe:
-                key_frame_indices.append(i)
+                keyframe_indices.append(i)
 
         self.tensor = np.moveaxis(np.stack(frames), 1, 2)
-        self.tags['key_frame_indices'] = key_frame_indices
+        self.tags['keyframe_indices'] = keyframe_indices
         return self
 
     def save_video_tensor_to_file(
--- a/docarray/document/mixins/video.py
+++ b/docarray/document/mixins/video.py
@@ -100,7 +100,7 @@ class VideoDataMixin:
     def load_uri_to_video_tensor(
         self: 'T', only_keyframes: bool = False, **kwargs
     ) -> 'T':
-        """Convert a :attr:`.uri` to a video ndarray :attr:`.tensor`.
+        """Convert a :attr:`.uri` to a video ndarray :attr:`.tensor` and store the indices of the key frames in :attr:`.tags`.
 
         :param only_keyframes: only keep the keyframes in the video
         :param kwargs: supports all keyword arguments that are being supported by av.open() as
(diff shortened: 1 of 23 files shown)


docarray/document/mixins/video.py
```diff
@@ -110,16 +110,36 @@ def load_uri_to_video_tensor(
...
+            stream = container.streams.video[0]
+            stream.codec_context.skip_frame = 'NONKEY'
 
-            frames = []
+            key_frames = []
             for frame in container.decode(video=0):
+                img = frame.to_image()
+                key_frames.append(img)
+
+        with av.open(self.uri, **kwargs) as container:
```
samsja, line 121: if we only want key frame we don't need to repopen the files again

docarray/document/mixins/video.py
```diff
@@ -110,16 +110,36 @@ def load_uri_to_video_tensor(
...
+            i = 0
+            curr_key_frame = key_frames[i]
+            key_frame_indices = []
+            frames = []
+            for j, frame in enumerate(container.decode(video=0)):
                 img = frame.to_image()
                 frames.append(np.asarray(img))
 
-        self.tensor = np.moveaxis(np.stack(frames), 1, 2)
+                if (np.asarray(img) == np.asarray(curr_key_frame)).all():
```
samsja, line 130: I am not a big fan of doing this. 
From what I understand av is able to tell when it load the video which frame is a key frame. It means that the info is either store in the file or that av can determine it. In both case we should do smth like this.

```python
for j, frame in enumerate(container.decode(video=0)):
    frames.append(np.asarray(img))
    if is_key_frame(frame):
        key_frame_indices.append(j)

```
the question is how de we do the `is_key_frame` but I believe av must have it

docarray/document/mixins/video.py
samsja: No, rather it is when `only_keyframes=true` a part of your current code is not needed it is executed but you return a result that you compute earlier so you can just skip it when you only want the keyframes

JoanFM (approved):

docarray/document/mixins/video.py
```diff
@@ -100,7 +100,7 @@ def generator_from_webcam(
     def load_uri_to_video_tensor(
         self: 'T', only_keyframes: bool = False, **kwargs
     ) -> 'T':
-        """Convert a :attr:`.uri` to a video ndarray :attr:`.tensor`.
+        """Convert a :attr:`.uri` to a video ndarray :attr:`.tensor` and store the indices of the keyframes in :attr:`.tags` if only_keyframes=False.
```
JohannesMessner, line 103: Maybe we can mention the key in tags that is used for this?

docs/datatypes/video/index.md
```diff
@@ -81,6 +81,22 @@ for idx, c in enumerate(d.tensor):
 
 Makes sense, right?
```
JohannesMessner, line 82:
```suggestion
```

Not related to this PR, but there are always some people to whom things don't make sense, let's not mock them lol

samsja (approved):
lgtm

docarray/document/mixins/video.py
```diff
@@ -102,7 +102,8 @@ def load_uri_to_video_tensor(
     ) -> 'T':
         """Convert a :attr:`.uri` to a video ndarray :attr:`.tensor`.
 
-        :param only_keyframes: only keep the keyframes in the video
+        :param only_keyframes: if true keep only the keyframes in the video, if false keep all frames and store the
+            indices of the keyframes in :attr:`.tags`
```
alexcg1, line 106:
```suggestion
        :param only_keyframes: if True keep only the keyframes, if False keep all frames and store the
            indices of the keyframes in :attr:`.tags`
```

docs/datatypes/video/index.md
```diff
@@ -79,7 +79,22 @@ for idx, c in enumerate(d.tensor):
 :width: 40%
 ```
 
-Makes sense, right?
+
+Additionally, when `only_keyframes=False` the indices of the keyframes are being stored in `.tags['keyframe_indices']` when calling `.load_uri_to_video_tensor`. This way, one can easily access selected scenes:
```
alexcg1, line 83:
```suggestion
Additionally, if `only_keyframes=False` the keyframe indices will be stored in `.tags['keyframe_indices']` when you call `.load_uri_to_video_tensor()`. This way, you can easily access selected scenes: 
```

alexcg1 (changes requested):

alexcg1 (approved):
LGTM :+1:

PR #899, opened 2022-12-06
chore(v2): add pytest mark slow to tests

samsja:
# Context

the goal of this pr is to mark some test are slow and mark which test require internet.

On my laptop test time is 8 sec and 1.7 without the slow test


JohannesMessner (commented):
Nice one

JohannesMessner (approved):

PR #900, opened 2022-12-06
refactor(v2):  add namespace docarray to tensor private method

samsja:
# Context

Our tensor type is built on top of torch tensor and numpy array. We inherit from both of this class and we add some custom prive function like `__validate_shape__` since that in a future we want to extend the tensor types to jax and tensorflow the likelihood that our custom private method clash with other framework private method increase. Therefore in this PR we rename all of the private method to `__docarray_XXX__`


JohannesMessner (approved):
Good idea

PR #920, opened 2022-12-08
fet(v2): proto stacked mode

samsja:
# Context

this pr allow to dump a document array into a proto in stacked mode


docarray/array/array.py
```diff
@@ -182,13 +182,37 @@ def __getitem__(self, item):  # note this should handle slices
         else:
             return self.__list_getitem__(item)
 
+    def __get_stacked_document__(self, item):
+        """
+        In stacked mode this return the Document where the filed stacked are put to None
+        In unstacked mode (default) this is the same as __getitem__
```
JohannesMessner, line 188:
```suggestion
        In stacked mode this returns the Document where the stacked fields are put to None
        In unstacked mode (default) this is the same as __getitem__
```

docarray/array/array.py
```diff
@@ -182,13 +182,37 @@ def __getitem__(self, item):  # note this should handle slices
...
+        :param item:
+        :return: Document
+        """
+        if self.is_stacked():
+            doc = super().__getitem__(item)
+            for field in self._columns.keys():
+                setattr(doc, field, None)
+            return doc
+        else:
+            return self.__list_getitem__(item)
```
JohannesMessner, line 199: What is the difference between this and `super().__getitem__(item)`? Why does line 194 do it the other way?
Could this be simplified?:
```python
doc = super().__getitem__(item)
if self.is_stacked():
    for field in self._columns.keys():
        setattr(doc, field, None)
return doc
```

docarray/array/mixins/proto.py
```diff
@@ -13,7 +14,18 @@ def from_protobuf(
     ) -> AbstractDocumentArray:
         """create a Document from a protobuf message"""
 
-        return cls(cls.document_type.from_protobuf(od) for od in pb_msg.docs)
+        content_type = pb_msg.WhichOneof('content')
+
+        if content_type == 'list_':
+            return cls(cls.document_type.from_protobuf(od) for od in pb_msg.list_.docs)
```
JohannesMessner, line 20: What does `od` stand for here?

docarray/array/mixins/proto.py
```diff
@@ -24,12 +36,35 @@ def to_protobuf(self) -> 'DocumentArrayProto':
...
 
-        from docarray.proto import DocumentArrayProto
+        from docarray.proto import (
+            DocumentArrayListProto,
+            DocumentArrayProto,
+            DocumentArrayStackedProto,
+            UnionArrayProto,
+        )
+
+        if self.is_stacked() and self._columns is not None:
```
JohannesMessner, line 46: I think this means that if it is stacked, but columns are non, then serializing and de-serializing will yield a da that is now not stacked, right? is that what we want?

JohannesMessner (changes requested):

docarray/array/array.py
```diff
@@ -182,13 +182,37 @@ def __getitem__(self, item):  # note this should handle slices
...
+
     def __list_iter__(self):
         return super().__iter__()
 
     def __iter__(self):
         for i in range(len(self)):
             yield self[i]
 
+    def __iter_over_stacked_documents__(self):
```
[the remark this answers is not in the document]
samsja: yes it should only be called in stacked mode. Technically it could be usefull to call in outside of proto as well maybe in the future.

I will refactor this part anyway lets not do a blocker now technically it will not failed anyway. I really dislike this pattern of fusing both implementation so that is why I want to refactor

docarray/array/mixins/proto.py
samsja: actually `self._columns is not None` is only here because of mypy. self._columns cannot be None if we are in stacked mode but mypy cannot know.

docarray/array/array.py
samsja: yes it can be

JohannesMessner (approved):

Issue #911, opened 2022-12-08
query operator $exists does not work correctly with tags


marcosbodio:
**Describe the bug**
The query operator `$exists` does not work correctly with `tags`.

**To Reproduce**
```python
from docarray import DocumentArray, Document

da = DocumentArray([Document(id=f'd{x}', x=x) for x in range(10)])
print(f"documents with key 'x' in tags: {da.find({'tags__x': {'$exists': True}})[:, 'id']}")
```

**Expected behavior**
The code above should print
```
documents with key 'x' in tags: ['d0', 'd1', 'd2', 'd3', 'd4', 'd5', 'd6', 'd7', 'd8', 'd9']
```
but instead it prints
```
documents with key 'x' in tags: ['d1', 'd2', 'd3', 'd4', 'd5', 'd6', 'd7', 'd8', 'd9']
```

**Desktop**
 - OS: macOS Ventura 13.0.1 (22A400)
 - Browser Firefox
 - Version Firefox 107.0.1 (64-bit)

PR #923, opened 2022-12-09
fix: query operator $exists does not work correctly with tags (#911)

marcosbodio:
See #911.
- Fix behaviour of `$exists` operator when applied to fields in `tags`: `{'tags__x': {'$exists': True}}` returns `True` when `tags.x is not None`
- Update documentation accordingly.


docs/fundamentals/documentarray/find.md
```diff
@@ -1,10 +1,16 @@
 (find-documentarray)=
+
 # Query by Conditions
 
-We can use {meth}`~docarray.array.mixins.find.FindMixin.find` to select Documents from a DocumentArray based the conditions specified in a `query` object. One can use `da.find(query)` to filter Documents and get nearest neighbours from `da`:
+We can use {meth}`~docarray.array.mixins.find.FindMixin.find` to select Documents from a DocumentArray based the
+conditions specified in a `query` object. One can use `da.find(query)` to filter Documents and get nearest neighbours
+from `da`:
```
JohannesMessner, line 7:
```suggestion
You can use {meth}`~docarray.array.mixins.find.FindMixin.find` to select Documents from a DocumentArray based the
conditions specified in a `query` object. You can use `da.find(query)` to filter Documents and get nearest neighbors
from `da`:
```

docs/fundamentals/documentarray/find.md
```diff
@@ -1,10 +1,16 @@
...
 
-- To filter Documents, the `query` object is a Python dictionary object that defines the filtering conditions using a [MongoDB](https://docs.mongodb.com/manual/reference/operator/query/)-like query language.
-- To find nearest neighbours, the `query` object needs to be a NdArray-like, a Document, or a DocumentArray object that defines embedding. One can also use `.match()` function for this purpose, and there is a minor interface difference between these two functions, which will be described {ref}`in the next chapter<match-documentarray>`.
+- To filter Documents, the `query` object is a Python dictionary object that defines the filtering conditions using
+  a [MongoDB](https://docs.mongodb.com/manual/reference/operator/query/)-like query language.
+- To find nearest neighbours, the `query` object needs to be a NdArray-like, a Document, or a DocumentArray object that
+  defines embedding. One can also use `.match()` function for this purpose, and there is a minor interface difference
+  between these two functions, which will be described {ref}`in the next chapter<match-documentarray>`.
```
JohannesMessner, line 13:
```suggestion
  defines embedding. You can also use `.match()` function for this purpose, and there is a minor interface difference
  between these two functions, which is described {ref}`in the next chapter<match-documentarray>`.
```

docs/fundamentals/documentarray/find.md
```diff
@@ -82,9 +88,10 @@ A query filter document can use the query operators to specify conditions in the
 { <field1>: { <operator1>: <value1> }, ... }
 ```
 
-Here `field1` is {ref}`any field name<doc-fields>` of a Document object.  To access nested fields, one can use the dunder expression. For example, `tags__timestamp` is to access `doc.tags['timestamp']` field.
+Here `field1` is {ref}`any field name<doc-fields>` of a Document object. To access nested fields, one can use the dunder
```
JohannesMessner, line 91:
```suggestion
Here `field1` is {ref}`any field name<doc-fields>` of a Document object. To access nested fields, you can use the dunder
```

docs/fundamentals/documentarray/find.md
```diff
@@ -131,37 +151,57 @@ r = da.find({'tags__h': {'$gt': 10}})
...
+      "w": 30.0
+    },
+    "text": "planner",
+    "weight": 75.0
+  }
+]
 ```
 
-Beside using a predefined value, one can also use a substitution with `{field}`, notice the curly brackets there. For example,
+Beside using a predefined value, one can also use a substitution with `{field}`, notice the curly brackets there. For
```
JohannesMessner, line 180:
```suggestion
Besides using a predefined value, you can also use a substitution with `{field}`, notice the curly brackets there. For
```

JohannesMessner (changes requested):
Looks good! just some minor tweaks to the documentation wording.

PR #931, opened 2022-12-12
feat: find function

JohannesMessner:
## Goals:

Create `find()` function:

```python
from docarray.utility import find, find_batched

class MyDoc(Document):
    tensor: TorchTensor

da = DocumentArray[MyDoc](MyDoc(tensor=torch.rand(128)) for _ in range(10))

matches, scores = find(da, MyDoc(tensor=torch.rand(128)), embedding_field='tensor', metric='cosine_sim')

batched_query = DocumentArray[MyDoc](
    [MyDoc(tensor=torch.rand(128)) for _ in range(3)]
)
results = find_batched(da, batched_query, embedding_field='tensor', metric='cosine_sim')
assert len(results) == 3
for matches_i, socres_i in results:
    ...

```

## TODO

- [x] implement find
  - [x] for torch
  - [x] for numpy
  - [x] allow docarray as query
- [ ] refactor: Create classes for backend operations. Edit: Separate PR
- [ ] refactor: let types define the backend they belong to. Edit: Separate PR
- [x] consider having `find` and `find_batched` explicitly
- [ ] user defined callable as distance function. Edit: not doing in first iteration
- [x] tests
- [x] Documentation (docstrings)
- [ ] Optional: batching Edit: not in this PR
- [ ] Optional: nested find.Edit: not doing in first iteration
- [ ] Optional: other features from current docarray. Edit: not doing in first iteration


docarray/utility/find.py
```diff
@@ -0,0 +1,321 @@
...
+
+def find_batched(
+    index: DocumentArray,
+    query: Union[Tensor, DocumentArray],
+    embedding_field: str = 'embedding',
+    metric: str = 'cosine_sim',
+    limit: int = 10,
+    device: Optional[str] = None,
+    descending: Optional[bool] = None,
+) -> List[FindResult]:
```
JohannesMessner, line 115: Is this a good return format?

JohannesMessner:
Note that the file structure is not great atm, this will also change with the PR about computational backends

docarray/utility/distance.py
```diff
@@ -0,0 +1 @@
+
```
samsja, line 1: we don't need this file I guess

docarray/utility/find.py
```diff
@@ -0,0 +1,321 @@
...
+    top_scores, top_indices = top_k_fn(
+        dists, k=limit, device=device, descending=descending
+    )
+
+    results = []
+    for indices_per_query, scores_per_query in zip(top_indices, top_scores):
+        docs_per_query = []
+        for idx in indices_per_query:  # workaround until #930 is fixed
+            docs_per_query.append(index[idx])
+        docs_per_query = DocumentArray(docs_per_query)
```
samsja, line 212: why not just add a this in a DocumentArray directly since DocumentArray is just a list of document

samsja: don't we need to do DocumentArray[smth] here ?

JohannesMessner: >why not just add a this in a DocumentArray directly since DocumentArray is just a list of document
Because `DocumentArray()` doesn't work. Maybe `DocumentArray.empty(0)`?
>don't we need to do DocumentArray[smth] here ?
Yes, good catch

samsja (approved):

PR #935, opened 2022-12-13
chore: add different python version to ci

samsja:

# What this pr do

* bump to python 3.8 for testing
* check import and install on different python version

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7]
+        python-version: [3.8]
         test-path: ${{fromJson(needs.prep-testbed.outputs.matrix)}}
     steps:
       - uses: actions/checkout@v2.5.0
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7]
+        python-version: [3.8]
```
JohannesMessner, line 35: Shouldn't this include all versions >= 3.8? Or only for CI?

.github/workflows/ci.yml
```diff
@@ -124,12 +124,15 @@ jobs:
 
   import-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-env: [3.8, 3.9, '3.10', '3.11']
```
JohannesMessner, line 129: Some have `''`, other don't, does it make a difference?

JohannesMessner (changes requested):

.github/workflows/ci.yml
alaeddine-13, line 129: why some are strings and some are not

samsja: fixed

samsja: fixed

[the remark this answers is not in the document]
samsja: well we don't support it so no. If this breaks because of numpy we cannot do anything

alaeddine-13 (approved):

PR #940, opened 2022-12-14
feat(v2): add audio url and predefined document

anna-charlotte:
Add support for audio files to docarray v2

Goals:

- [x] `AudioUrl`
- [x] `AudioTensor` with `AudioNdarray` and `AudioTorchTensor`
- [x] `Audio` predefined doc (with optional attrs `Audio.tensor`, `Audio.url` and `Audio.embedding`)
- [x] tests


docarray/typing/tensor/audio/audio_tensor.py
```diff
@@ -0,0 +1,13 @@
+from typing import Union
+
+from docarray.typing.tensor.audio.audio_ndarray import AudioNdArray
+
+try:
+    import torch  # noqa: F401
+except ImportError:
+    AudioTensor = Union[AudioNdArray]
```
JoanFM, line 8: why a Union?

docarray/typing/url/audio_url.py
```diff
@@ -0,0 +1,114 @@
...
+            audio_tensor = doc.audio_url.load()
+            assert isinstance(audio_tensor, np.ndarray)
+
+        :return: np.ndarray representing the audio file content
+        """
+        import io
+
+        file: Union[io.BytesIO, T]
+
+        if self.startswith('http'):
```
JoanFM, line 81: how is this `startswith` implemented on `self`? how is `T` guaranteed to implement this method?

tests/units/typing/tensor/test_audio_tensor.py
```diff
@@ -0,0 +1,95 @@
...
+        (AudioTorchTensor, torch.zeros(1000, 2)),
+        (AudioNdArray, np.zeros((1000, 2))),
+    ],
+)
+def test_save_audio_tensor_to_wav_file(cls_tensor, tensor):
+    tmp_file = str(TOYDATA_DIR / 'tmp.wav')
+    audio_tensor = parse_obj_as(cls_tensor, tensor)
+    audio_tensor.save_to_wav_file(tmp_file)
+    assert os.path.isfile(tmp_file)
+    os.remove(tmp_file)
```
JoanFM, line 95: u do not need to this, or if you do this, use a `fixture` or `context` manager. If `assert` fails it should be removed as well right?

tests/units/typing/url/test_audio_url.py
```diff
@@ -0,0 +1,78 @@
...
+    [
+        *[('wav', file) for file in AUDIO_FILES],
+        ('wav', REMOTE_AUDIO_FILE),
+        ('illegal', 'illegal'),
+        ('illegal', 'https://www.google.com'),
+        ('illegal', 'my/local/text/file.txt'),
+        ('illegal', 'my/local/text/file.png'),
+    ],
+)
+def test_validation(file_format, path_to_file):
```
JoanFM, line 60: I'd rathewr have a `test_illegal_validation` and a `test_validation` separated as different tests

tests/units/typing/url/test_audio_url.py
```diff
@@ -0,0 +1,78 @@
...
+
+@pytest.mark.slow
+@pytest.mark.internet
+@pytest.mark.parametrize(
+    'file_url',
+    [*AUDIO_FILES, REMOTE_AUDIO_FILE],
+)
+def test_proto_audio_url(file_url):
+    uri = parse_obj_as(AudioUrl, file_url)
+    uri._to_node_protobuf()
```
JoanFM, line 78: no assertion?

JoanFM (changes requested):

docarray/predefined_document/audio.py
```diff
@@ -0,0 +1,72 @@
...
+        audio.tensor = audio.url.load()
+        model = MyEmbeddingModel()
+        audio.embedding = model(audio.tensor)
+
+    You can extend this Document:
+
+    .. code-block:: python
+
+        from docarray import Audio
+        from docarray.typing import Embedding
```
alaeddine-13, line 36: looks like this is not needed

docarray/typing/tensor/audio/abstract_audio_tensor.py
```diff
@@ -0,0 +1,25 @@
...
+    ) -> None:
+        """
+        Save tensor to a .wav file. Mono/stereo is preserved.
+
+        :param file_path: path to a .wav file. If file is a string, open the file by
+            that name, otherwise treat it as a file-like object.
+        :param sample_rate: sampling frequency
+        :param sample_width: sample width in bytes
+        """
+        ...
```
alaeddine-13, line 25: I suggest the following:
* put all audio-specific logic (save_to_wav_file without tensor to byte conversion and _to_node_protobuf) into 1 class, ideally this class. I think we can make these 2 methods concrete and therefore this class can be turned into a mixin rather than an abstract class.
* move the tensor-to-bytes logic into tensor classes as a method `def tobytes()`. Then use this common method when implementing the common `save_to_wav_file()` method. Alternatively, and if you believe this logic is more related to audio, you can define `to_audio_bytes` in concrete audio tensor classes, but still the implementation of `save_to_wav_file` should stay common in the parent audio tensor class.
* A concrete audio tensor class is just a class that combines the base audio tensor class and the concrete.
I will be making suggestions in this direction

docarray/typing/tensor/audio/abstract_audio_tensor.py
```diff
@@ -0,0 +1,25 @@
+from abc import ABC, abstractmethod
+from typing import BinaryIO, TypeVar, Union
+
+from docarray.typing.tensor.abstract_tensor import AbstractTensor
+
+T = TypeVar('T', bound='AbstractAudioTensor')
+
+
+class AbstractAudioTensor(AbstractTensor, ABC):
+    @abstractmethod
```
alaeddine-13, line 10: rename to AudioTensorMixin and remove the abstract method. Move the common logic of save_to_wav_file in the other classes to this one.
For the logic related to converting tensor to bytes, you use the method `self.tobytes()` which we should define for tensor classes

docarray/typing/tensor/audio/audio_ndarray.py
```diff
@@ -0,0 +1,83 @@
...
+
+from docarray.typing import NdArray
+
+T = TypeVar('T', bound='AudioNdArray')
+
+if TYPE_CHECKING:
+    from docarray.proto import NodeProto
+
+
+class AudioNdArray(NdArray):
```
alaeddine-13, line 12: remove the implementation of both methods and just apply this modification
```suggestion
class AudioNdArray(AudioTensorMixin, NdArray):
```

docarray/typing/tensor/audio/audio_torch_tensor.py
```diff
@@ -0,0 +1,84 @@
...
+
+from docarray.typing.tensor.torch_tensor import TorchTensor, metaTorchAndNode
+
+T = TypeVar('T', bound='AudioTorchTensor')
+
+if TYPE_CHECKING:
+    from docarray.proto import NodeProto
+
+
+class AudioTorchTensor(TorchTensor, metaclass=metaTorchAndNode):
```
alaeddine-13, line 12: remove the implementation of both methods and apply this change
```suggestion
class AudioTorchTensor(AudioTensorMixin, TorchTensor, metaclass=metaTorchAndNode):
```

docarray/typing/tensor/audio/audio_ndarray.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def save_to_wav_file(
+        self: 'T',
+        file_path: Union[str, BinaryIO],
+        sample_rate: int = 44100,
+        sample_width: int = 2,
+    ) -> None:
+
+        max_int16 = 2**15
+        tensor = (self * max_int16).astype('<h')
```
alaeddine-13, line 76: this logic combined with tensor.tobytes() I believe belongs to the NdArray class. We can add a method for it `tobytes()`

docarray/typing/tensor/audio/audio_torch_tensor.py
```diff
@@ -0,0 +1,84 @@
...
+        self: 'T',
+        file_path: Union[str, BinaryIO],
+        sample_rate: int = 44100,
+        sample_width: int = 2,
+    ) -> None:
+        import torch
+
+        max_int16 = 2**15
+        tensor = self * max_int16
+        tensor = tensor.to(dtype=torch.int16)
```
alaeddine-13, line 77: this logic and `tensor.cpu().detach().numpy().tobytes()` also I believe belongs to TorchTensor and should be moved there `def tobytes`

docarray/predefined_document/audio.py
```diff
@@ -0,0 +1,71 @@
...
+    EXAMPLE USAGE:
+
+    You can use this Document directly:
+
+    .. code-block:: python
+
+        from docarray import Audio
+
+        # use it directly
+        audio = Audio(url='https://www.kozco.com/tech/piano2.wav')
```
alaeddine-13, line 26: Maybe we should rely on a file from github rather than a file from the internet ?
how about using the files here ? https://github.com/docarray/docarray/tree/feat-add-audio-v2/tests/toydata

tests/integrations/predefined_document/test_audio.py
```diff
@@ -0,0 +1,85 @@
...
+from docarray import Audio
+from docarray.typing import AudioUrl
+from docarray.typing.tensor.audio import AudioNdArray, AudioTorchTensor
+from tests import TOYDATA_DIR
+
+LOCAL_AUDIO_FILES = [
+    str(TOYDATA_DIR / 'hello.wav'),
+    str(TOYDATA_DIR / 'olleh.wav'),
+]
+REMOTE_AUDIO_FILE = 'https://www.kozco.com/tech/piano2.wav'
```
alaeddine-13, line 18: same suggestion

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -116,3 +117,8 @@ def __docarray_from_native__(cls: Type[T], value: Any) -> T:
     def get_comp_backend() -> Type[AbstractComputationalBackend]:
         """The computational backend compatible with this tensor type."""
         ...
+
+    @abc.abstractmethod
+    def n_dim(self) -> int:
+        """Get number of tensor dimensions."""
+        ...
```
alaeddine-13, line 124: not sure this is necessary, since (for now) tensors implement ndim in the same way. My concern is that we're adding an unnecessary method to the public interface. @JoanFM what do you think ?

docarray/typing/tensor/audio/audio_tensor.py
```diff
@@ -0,0 +1,13 @@
...
+
+try:
+    import torch  # noqa: F401
+except ImportError:
+    AudioTensor = AudioNdArray
+
+else:
+    from docarray.typing.tensor.audio.audio_torch_tensor import AudioTorchTensor
+
+    AudioTensor = Union[AudioNdArray, AudioTorchTensor]  # type: ignore
```
samsja, line 13: why is the type ignore required here ?

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -15,6 +15,7 @@
 
 class AbstractTensor(AbstractType, Generic[ShapeT], ABC):
 
+    TENSOR_FIELD_NAME: str
```
samsja, line 18:
```suggestion
    _PROTO_FIELD_NAME: str
```

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -116,3 +117,8 @@ def __docarray_from_native__(cls: Type[T], value: Any) -> T:
     def get_comp_backend() -> Type[AbstractComputationalBackend]:
         """The computational backend compatible with this tensor type."""
         ...
+
+    @abc.abstractmethod
+    def n_dim(self) -> int:
```
samsja, line 122: I fear that this name can clash with future backend. I would rather call it `__docarray_n_dim__`

tests/units/typing/tensor/test_audio_tensor.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@pytest.mark.parametrize(
+    'tensor,cls_audio_tensor,cls_tensor',
+    [
+        (torch.zeros(1000, 2), AudioTorchTensor, torch.Tensor),
+        (np.zeros((1000, 2)), AudioNdArray, np.ndarray),
+    ],
+)
+def test_set_audio_torch_tensor_audio_ndarray(tensor, cls_audio_tensor, cls_tensor):
```
samsja, line 21:
```suggestion
def test_set_audio_tensor(tensor, cls_audio_tensor, cls_tensor):
```

tests/units/typing/tensor/test_audio_tensor.py
```diff
@@ -0,0 +1,99 @@
...
+    class MyAudioDoc(Document):
+        tensor: cls_audio_tensor
+
+    doc = MyAudioDoc(tensor=tensor)
+    assert isinstance(doc.tensor, cls_audio_tensor)
+    assert isinstance(doc.tensor, cls_tensor)
+    assert (doc.tensor == tensor).all()
+
+
+def test_set_audio_tensor():
```
samsja, line 31: this test is redundant with the one before

samsja (commented):
This pr looks really nice. I like how the idea of having different tensor types for the different modality work :) I added some comments

docarray/typing/tensor/abstract_tensor.py
JohannesMessner: I agree, let's not add it here. If anything this should go into the computational backend. The tensor interface itself should stay as lean as possible, we don't want to run the risk of interfering with the "actual" tensors that we inherit from (torch, numpy, etc).

docarray/typing/tensor/audio/audio_torch_tensor.py
```diff
@@ -0,0 +1,59 @@
...
+from docarray.typing.tensor.torch_tensor import TorchTensor, metaTorchAndNode
+from docarray.typing.url.audio_url import MAX_INT_16
+
+T = TypeVar('T', bound='AudioTorchTensor')
+
+
+class AudioTorchTensor(AbstractAudioTensor, TorchTensor, metaclass=metaTorchAndNode):
+    """
+    Subclass of TorchTensor, to represent an audio tensor.
+    Additionally, this allows storing such a tensor as a .wav audio file.
```
JohannesMessner, line 13:
```suggestion
    Adds audio-specific features to the tensor.
```
This way it won't get outdated when/if we add more stuff in the future. The specific docstring should go to the methods themselves

docarray/typing/tensor/abstract_tensor.py
JohannesMessner, line 18: good move

docarray/typing/tensor/audio/audio_ndarray.py
```diff
@@ -0,0 +1,58 @@
...
+from docarray.typing.tensor.audio.abstract_audio_tensor import AbstractAudioTensor
+from docarray.typing.url.audio_url import MAX_INT_16
+
+T = TypeVar('T', bound='AudioNdArray')
+
+
+class AudioNdArray(AbstractAudioTensor, NdArray):
+    """
+    Subclass of NdArray, to represent an audio tensor.
+    Additionally, this allows storing such a tensor as a .wav audio file.
```
JohannesMessner, line 13: same here with the docstring

docarray/typing/url/audio_url.py
```diff
@@ -0,0 +1,114 @@
...
+        url = super().validate(value, field, config)  # basic url validation
+        has_audio_extension = any(ext in url for ext in AUDIO_FILE_FORMATS)
+        if not has_audio_extension:
+            raise ValueError(
+                f'Audio URL must have one of the following extensions:'
+                f'{AUDIO_FILE_FORMATS}'
+            )
+        return cls(str(url), scheme=None)
+
+    def load(self: T) -> np.ndarray:
```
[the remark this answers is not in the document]
JohannesMessner: I would say `AudioNdArray`.
Normally we just load the np.ndarray, bc the framework specific conversion can happen when putting it back into the document, and having it as np.ndarray makes the least assumptions about the sorrounding code.
In this case, since there are actual audio features that come with the array, I would go with `AudioNdArray`. It can still be treated like a normal np.ndarray, but bring the aforementioned features.

But just verify in a test that setting a `AudioNdArray` to a field with type `AudioTorchArray` actually works without issue?

docarray/typing/url/audio_url.py
```diff
@@ -0,0 +1,114 @@
...
+            file = self
+
+        # note wave is Python built-in mod. https://docs.python.org/3/library/wave.html
+        with wave.open(file) as ifile:
+            samples = ifile.getnframes()
+            audio = ifile.readframes(samples)
+
+            # Convert buffer to float32 using NumPy
+            audio_as_np_int16 = np.frombuffer(audio, dtype=np.int16)
+            audio_as_np_float32 = audio_as_np_int16.astype(np.float32)
```
JohannesMessner, line 100: Should we give (optional) control over the dtype?

JohannesMessner (changes requested):
Great PR, love the audio tensor stuff, this is a good pattern to use moving forward. just some things to consider

docarray/typing/tensor/audio/audio_tensor.py
samsja: okay we do the same actually with `Tensor`

docarray/typing/url/audio_url.py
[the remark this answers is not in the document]
samsja: great I like it better like this

JohannesMessner (approved):
Just resolve the conflicts and we're good to go, great work!

samsja (approved):

JoanFM (approved):

PR #941, opened 2022-12-14
refactor(v2): computational backends

JohannesMessner:
Goals:

This introduces "computational backends" for organizing and grouping functionality that is framework specific, e.g. numpy or torch


docarray/computation/abstract_comp_backend.py
```diff
@@ -0,0 +1,127 @@
...
+    @abstractmethod
+    def stack(
+        tensors: Union[List['Tensor'], Tuple['Tensor']], dim: int = 0
+    ) -> 'Tensor':
+        """
+        Stack a list of tensors along a new axis.
+        """
+        ...
+
+    class Retrieval(ABC):
```
samsja, line 26: why is this inside the other class ?

JohannesMessner: just for further structuring/namespacing things. so all the metrics can go in one place, all the retrieval related stuff in another place, etc

docarray/typing/abstract_type.py
```diff
@@ -13,6 +13,8 @@
 
 
 class AbstractType(BaseNode):
+    is_tensor = False  # change for tensor-like subclasses
```
samsja, line 16: I am not sure to understand why we need this. Only object of type AbstractTensor will have this `is_tensor=True` right ? So why not just do
```python
if issubclass(type_, AbstractType):
    if issubclass(type_, AbstractTensor):
        smth_with_tensor()
    else:
        smth_without_tensor

```

JohannesMessner: Yeah true, will revert

samsja (approved):

PR #946, opened 2022-12-14
refactor(v2): seperate stacked and not stacked document array

samsja:
# context

refactor the stacked mode to have two separate class

```python
    class Image(Document):
        tensor: TorchTensor[3, 224, 224]

    da = DocumentArray[Image](
        [Image(tensor=torch.zeros(3, 224, 224)) for _ in range(10)]
    )

    da = da.stack()
    # do stacked stuff
    da = da.unstack()

```

or with context manager

```python
with da.stacked_mode():
   ...
```


docarray/array/abstract_array.py
```diff
@@ -1,37 +1,94 @@
...
+    from docarray.proto import DocumentArrayProto, NodeProto
     from docarray.typing import NdArray, TorchTensor
 
 
-class AbstractDocumentArray(Sequence):
+T = TypeVar('T', bound='AbstractDocumentArray')
+T_doc = TypeVar('T_doc', bound=BaseDocument)
 
+
+class AbstractDocumentArray(Sequence[BaseDocument], Generic[T_doc], AbstractType):
```
JohannesMessner, line 16: What are the implications of not inheriting from list?

docarray/utility/find.py
```diff
@@ -16,7 +17,7 @@ class FindResult(NamedTuple):
 
 
 def find(
-    index: DocumentArray,
+    index: Union[DocumentArray, DocumentArrayStacked],
```
JohannesMessner, line 20: This seems a bit smelly to me. On the one hand, the user doesn't know that stacked and non stacked are two different classes, they just call stack() and unstack() on on object. On the other hand, they need to know about this distinction in cases like this

JohannesMessner (changes requested):
I am not sure about some of the details of the user interaction here. It seems like in some places we hide these two different classes, in other places we don't. I think this might not be clear to the user.

docarray/array/abstract_array.py
samsja: I decided not to inherit from list because we cannot really follow the list interface, example the __class_getitem__ is clashing with the list one etcc. Overall it was a pain to follow the list interface because we slightly need to change it

docarray/utility/find.py
samsja: I see your point probably we need to revisit the different concept that we expose to the user. Maybe what we need is a to expose the Abstract class under the name DocumentContainer

docarray/array/abstract_array.py
JohannesMessner: Makes sense, agree

docarray/utility/find.py
JohannesMessner: Something like this, yes. Or go all the way and own `DocumentArray` and `DocumentList` with `stack()/unstack()` to go between them, but also telling this to the user. The way it currently is I don't think we should merge the PR, since it makes things more confusing on the outside (to our users) with the only benefit being on the inside (a refactoring)

samsja: Actually I think we don't have a choice but to expose the Union. Because you need to call stack anyway inside your method if this is not yet stacked. And the stack method would only be available at the `DocuentArray` level so the Union will always be needed

samsja: okay I have an idea let me try

JohannesMessner (approved):

PR #949, opened 2022-12-15
fix: unwanted coercion in to_pydantic_model

JohannesMessner:
Goals:

This raises the required pydantic version to fix unwanted type coercion in `doc.to_pydantic_model()`, and by extension, in `doc.to_dict()`.

Before:
```python
from jina import Document

d = Document()
d.tags['title'] = 'Infinity'

print(d.to_pydantic_model())
```
```text
id='f3fbf26990ef007fe577994b634140db' parent_id=None granularity=None adjacency=None blob=None tensor=None mime_type=None text=None weight=None uri=None tags={'title': 'inf'} offset=None location=None embedding=None modality=None evaluations=None scores=None chunks=None matches=None
```
Note `tags={'title': 'inf'}`, it coerced `"Infinity"` to `inf`, not what we want.
After:
```python
from jina import Document

d = Document()
d.tags['title'] = 'Infinity'

print(d.to_pydantic_model())
```
```text
id='f3fbf26990ef007fe577994b634140db' parent_id=None granularity=None adjacency=None blob=None tensor=None mime_type=None text=None weight=None uri=None tags={'title': 'Infinity'} offset=None location=None embedding=None modality=None evaluations=None scores=None chunks=None matches=None
```
Note `tags={'title': 'Infinity'}`, no unwanted coercion.


JoanFM (changes requested):
Can we add a test showing the behavior is correct?

JoanFM (approved):

PR #950, opened 2022-12-15
fix: calculate relevant docs on index instead of queries

guenthermi:
Goals:

- The `num_relevant_documents_per_label` is calculated in the `embed_and_evaluate` method on the documents in `self`. However, this is only correct if `self` is matched against itself. Instead it should be calculated on the `index_data` attribute if it is provided.

### DIFF
--- a/tests/unit/array/mixins/oldproto/test_eval_class.py
+++ b/tests/unit/array/mixins/oldproto/test_eval_class.py
@@ -529,6 +529,54 @@ def test_embed_and_evaluate_single_da(storage, config, start_storage):
 
 
 @pytest.mark.parametrize(
+    'exclude_self, expected_results',
+    [
+        (
+            True,
+            {
+                'r_precision': 1,
+                'precision_at_k': 1.0,
+                'hit_at_k': 1.0,
+                'average_precision': 1.0,
+                'reciprocal_rank': 1.0,
+                'recall_at_k': 5.0 / 9.0,
+                'f1_score_at_k': (10.0 / 9.0) / (5.0 / 9.0 + 1),
+                'ndcg_at_k': 1.0,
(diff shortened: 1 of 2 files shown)


tests/unit/array/mixins/oldproto/test_eval_class.py
```diff
@@ -567,6 +567,21 @@ def test_embed_and_evaluate_two_das(storage, config, sample_size, start_storage)
...
+        [Document(text=str(i), label=i % 10) for i in range(10, 110)]
+    )
+    res = queries_da.embed_and_evaluate(
+        index_data=index_da,
+        metrics=['precision_at_k', 'reciprocal_rank', 'recall_at_k', 'f1_score_at_k'],
+        embed_funcs=dummy_embed_function,
+        match_batch_size=1,
+        limit=10,
+    )
+    print(res)
```
JoanFM, line 582: remove print, there should be assertions

JoanFM (changes requested):
Put it as draft until really ready for review

tests/unit/array/mixins/oldproto/test_eval_class.py
```diff
@@ -639,7 +654,7 @@ def emb_func(da):
     ],
 )
 def test_embed_and_evaluate_on_real_data(two_embed_funcs, kwargs):
-    metric_names = ['precision_at_k', 'reciprocal_rank']
+    metric_names = ['precision_at_k', 'reciprocal_rank', 'recall_at_k']
```
LMMilliken, line 657: Maybe add a test case for `exclude_self` = True/False

JoanFM (approved):

gmastrapas (approved):

PR #951, opened 2022-12-16
fix(v2): handle union and optional types

### LINKED ISSUE #939: Figure out handling of Union types
Many features of our current implementation rely on the type hints under the hood to infer behaviour.
This break whenever the type hint is of form `Union` or `Optional`, since those cannot be used in instance and subclass checks.

This is especially problematic, since our type `Tensor` is a Union under the hood, and it is the type used by our pre buit documents. For example, `find()` and `stack()` don't currently work with `Image`.

The simplest workaround is probably to check if a type is union or optional, extract the types inside them, and perform subclass/instance checks on those

JohannesMessner:
Signed-off-by: Johannes Messner <messnerjo@gmail.com>

Goals:

Properly handle `Union` types (such as `Tensor = Union[TorchTensor, NdArray]`) and `Optional` types in stack, find, and bulk access.
Before, all of these cases would throw an exception.

Closes #939


docarray/array/array.py
```diff
@@ -13,6 +13,8 @@
     Union,
 )
 
+from typing_inspect import is_union_type  # type: ignore
```
samsja, line 16: you don't need the type ignore if you add to pyprojet toml

docarray/util/find.py
```diff
@@ -1,4 +1,6 @@
-from typing import List, NamedTuple, Optional, Type, Union
+from typing import List, NamedTuple, Optional, Type, Union, get_args
+
+from typing_inspect import is_optional_type, is_union_type  # type: ignore
```
samsja, line 3: same not needed

docarray/array/mixins/attribute.py
```diff
@@ -21,6 +23,11 @@ def _get_array_attribute(
         in the array like container
         """
         field_type = self.__class__.document_type._get_nested_document_class(field)
+        if is_optional_type(field_type):
+            field_type = get_args(field_type)[0]  # get type inside Optional
+        elif is_union_type(field_type):
+            # determine type based on the first element
+            field_type = type(getattr(self[0], field))
```
samsja, line 30: so this looks at the data to determine the type right ? Smart idea

JohannesMessner: exactly, I don't like it so much actually, but I think it's the only way

JohannesMessner:
Converting to draft bc there are two scenarios still untackled:
- Combination of Optional and Union, which is the case for pre-defined Documents
- stacking Union types should be supported, since pre-defined docs use union type

tests/units/array/test_mixins/test_attribute.py
```diff
@@ -0,0 +1,168 @@
+from typing import Optional, Union
```
samsja, line 1: we don't have mixin in array anymore. we should move this file to somewhere else

samsja: did u write all of these tests? I don't understand how they are all relevant.

JohannesMessner: moved to `test_array`

samsja (approved):

Issue #952, opened 2022-12-19
Is Chinese retrieval supported in Redis backend


SuiJiGuoChengSuiJiGuo:
RediSearch support chinese retrieval https://redis.io/docs/stack/search/chinese/.  But I can't find a place to configure in 'config' or 'Redis config'.  
I want to ask for help about this feature. 
Thanks.

JoanFM:
Hey @SuiJiGuoChengSuiJiGuo ,

We are working on a PR to expose this possibility

SuiJiGuoChengSuiJiGuo:
Thank you for your help!!!

PR #953, opened 2022-12-19
feat: add language attribute to RedisConfig

AnneYang720:

Goals:

Attribute `language` is added to RedisConfig, so that user can search in certain language, especially chinese.
#952 

- [x] add code
- [x] add test


docarray/array/storage/opensearch/find.py
```diff
@@ -166,7 +166,7 @@ def _find(
         :return: DocumentArray containing the closest documents to the query if it is a single query, otherwise a list of DocumentArrays containing
            the closest Document objects for each of the queries in `query`.
         """
-        query = np.array(query).astype(np.float)
+        query = np.array(query).astype(np.float32)
```
JoanFM, line 169: why this?

JoanFM: why touching opensearch at all?

JoanFM (changes requested):

docarray/array/storage/opensearch/find.py
AnneYang720: `numpy` has a new release three days ago and its version is updated to `1.24.0` in ci, where `np.float` and `np.int` is no longer available

JoanFM (approved):

PR #956, opened 2022-12-19
feat: add nested access for document array

anna-charlotte:
 
Nested access in V2 is possible by calling the nested attributes you want to access on a DocumentArray, e.g. `da.image.title`. This does not work for nested DocumentArrays though:

```python
class Author(Document):
    name: str


class Book(Document):
    authors: Optional[DocumentArray[Author]]
    content: str


da = DocumentArray[Book](
    [
        Book(
            authors=DocumentArray[Author]([Author(name='Jenny'), Author(name='Ben')]),
            content=f'hello{i}'
        ) for i in range(10)
    ]
)

authors = da.authors.name  # does not work since da.authors is a list of DocumentArrays and does not have

# instead do this:
authors = da.traverse_flat(access_path='authors.name')  # returns List of strings, with len = 20

```

`traverse_flat` takes the access path and returns a list of those attributes, flattened on the first level. 
`da.traverse_flat(access_path='authors.name')` therefore returns a list of 20 strings instead of list of ten lists of 2 names each.

- [ ] add traverse_flat + test


docarray/array/abstract_array.py
```diff
@@ -92,3 +91,119 @@ def _to_node_protobuf(self) -> 'NodeProto':
...
+                x = getattr(node, curr_attr)
+                yield from AnyDocumentArray._traverse(x, path_attrs)
+        else:
+            yield node
+
+    @staticmethod
+    def _flatten(sequence) -> List[Any]:
+        from docarray import DocumentArray
+
+        if isinstance(sequence[0], (list, DocumentArray)):
```
JoanFM, line 206: please add `typehints`. Also what happens if `sequence` has a length of `0`?

JoanFM (changes requested):

docarray/array/array.py
```diff
@@ -77,7 +79,10 @@ def __len__(self):
         return len(self._data)
 
     def __getitem__(self, item):
-        return self._data[item]
+        if isinstance(item, (int, np.generic)):
+            return self._data[item]
+        elif isinstance(item, str):
+            pass
```
JoanFM, line 85: in which case does this happen?

JoanFM:
In the v1, we decided to hide traverse flat behind the [@] accessors, are we expecting to expose this now?

JoanFM (approved):

PR #961, opened 2022-12-27
feat: add color to point cloud display

anna-charlotte:

When displaying a point cloud, we want to additionally consider its colors, if those are available. 
The point cloud tensor is stored in a Documents `.tensor` field. If colors are available store them in a chunk `Document` with `tags['name'] = 'point_cloud_colors'`. The chunks tensor must be of shape (n_points, 3) or (n_points, 4), with a corresponding color for each point.

- [x] check for point_colors in `display_point_cloud_tensor()`


JoanFM (changes requested):
add a test to show that at least it does not break

docarray/document/mixins/plot.py
```diff
@@ -185,15 +185,24 @@ def display_point_cloud_tensor(self) -> None:
         from IPython.display import display
         from hubble.utils.notebook import is_notebook
 
+        colors = np.tile(np.array([0, 0, 0]), (len(self.tensor), 1))
+        for chunk in self.chunks:
+            if (
+                'name' in chunk.tags.keys()
+                and chunk.tags['name'] == 'point_colors'
```
JoanFM, line 192: this point_colors string should come from some common variable, otherwise is just a magic name

JoanFM (changes requested):

docarray/document/mixins/plot.py
JoanFM: yes

tests/unit/document/test_summary.py
```diff
@@ -69,3 +69,22 @@ def test_plot_embedding():
...
     d.summary()
+
+
+@pytest.mark.parametrize(
+    'colors', [None, np.random.rand(100, 3), np.random.rand(100, 4)]
+)
+def test_plot_point_cloud(colors):
+    import os
+
+    os.environ['DISPLAY'] = ':1'
```
JoanFM, line 80: see in jina how in tests some env vars are set. This should be set in a fixture so that later the value can be reset to the old value

JoanFM (changes requested):

docarray/document/mixins/mesh.py
```diff
@@ -17,6 +17,10 @@ class Mesh:
     FACES = 'faces'
 
 
+class PointCloud:
```
samsja, line 20: lets use proper python ENUM here : https://docs.python.org/3/library/enum.html

samsja: and lets call it PointCloud Enum

docarray/document/mixins/plot.py
```diff
@@ -185,15 +185,24 @@ def display_point_cloud_tensor(self) -> None:
         from IPython.display import display
         from hubble.utils.notebook import is_notebook
 
+        colors = np.tile(np.array([0, 0, 0]), (len(self.tensor), 1))
+        for chunk in self.chunks:
+            if (
+                'name' in chunk.tags.keys()
+                and chunk.tags['name'] == PointCloud.COLORS
+                and (chunk.tensor.shape[-1] == 3 or chunk.tensor.shape[-1] == 4)
```
samsja, line 193:
```suggestion
                and chunk.tensor.shape[-1] in [3, 4]

```

docs/datatypes/mesh/index.md
```diff
@@ -2576,6 +2576,16 @@ init();</script></body>
 </html>" width="100%" height="500px" style="border:none;"></iframe
 
 
+To display a colored point cloud, store the corresponding colors in the `.tensor` of a chunk `Document` with the name tag `'point_cloud_colors'`. The colors have to be of shape (n_samples, 3) or (n_samples, 4).
+
+```python
```
samsja, line 2581: is there a way we can load both color and data from some files ?

setup.py
```diff
@@ -60,7 +60,7 @@
             'requests',
             'matplotlib',
             'Pillow',
-            'trimesh',
+            'trimesh[easy]',
```
samsja, line 63: what is the difference with easy ?

docarray/document/mixins/plot.py
```diff
@@ -185,15 +185,24 @@ def display_point_cloud_tensor(self) -> None:
         from IPython.display import display
         from hubble.utils.notebook import is_notebook
 
+        colors = np.tile(np.array([0, 0, 0]), (len(self.tensor), 1))
+        for chunk in self.chunks:
+            if (
+                'name' in chunk.tags.keys()
```
JohannesMessner, line 191: Is `'name`' a key that is already established in other parts of the code / other features? If not I think I would prefer a more descriptive key here

docs/datatypes/mesh/index.md
```diff
@@ -2576,6 +2576,16 @@ init();</script></body>
...
+n_samples = 1000
+colors = np.random.rand(n_samples, 3)
+doc = Document(
+    uri='mesh_man.glb', chunks=[Document(tensor=colors, name='point_cloud_colors')]
+).load_uri_to_point_cloud_tensor(samples=n_samples)
+```
```
JohannesMessner, line 2587: Maybe we can switch this code snippet around to first load the point coordinate like always, and then add the color chunk in a second step

JohannesMessner (changes requested):
Did you remove the test that you mention in the comment here?

docs/datatypes/mesh/index.md
samsja: okay I see and in practice how would I load the color with trimesh ? How do I get this colors array ?

samsja: okay I see makes sense then

setup.py
```diff
@@ -61,6 +61,8 @@
             'matplotlib',
             'Pillow',
             'trimesh',
+            'ipython',
+            'pyglet<=1.5.27',  # required by trimesh
```
samsja, line 65: Why is the reason not to let trimesh handle this dependency capping ?

docarray/document/mixins/plot.py
JohannesMessner: no sounds good, let's keep it consistent

JohannesMessner (approved):

JoanFM (approved):

PR #963, opened 2023-01-02
fix: replace newline with whitespace to fix display in plot embeddings

anna-charlotte:

If a Documents fields contain '\n' this messes up the display in `.plot_embeddings()`. 
Therefore we want to replace all '\n' with whitespaces for display.

E.g. for the following Document:

image

The information of that Document looks like this:
Screenshot 2023-01-02 at 09 27 03

But should look like this:
Screenshot 2023-01-02 at 09 06 41

- [x] replace all '\n' with whitespaces in `.plot_embeddings()`

### DIFF
--- a/docarray/array/mixins/io/csv.py
+++ b/docarray/array/mixins/io/csv.py
@@ -78,17 +78,17 @@ class CsvIOMixin:
                 writer.writeheader()
 
             for d in self:
-                pd = d.to_dict(
+                doc_dict = d.to_dict(
                     protocol='jsonschema',
                     exclude=set(exclude_fields) if exclude_fields else None,
                     exclude_none=True,
                 )
                 if flatten_tags:
-                    t = pd.pop('tags')
-                    pd.update({f'tag__{k}': v for k, v in t.items()})
+                    t = doc_dict.pop('tags')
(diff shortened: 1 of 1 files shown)


docarray/array/mixins/io/csv.py
```diff
@@ -86,6 +86,8 @@ def save_csv(
                 if flatten_tags:
                     t = pd.pop('tags')
                     pd.update({f'tag__{k}': v for k, v in t.items()})
+
+                pd = {k: str(v).replace('\n', ' ') for k, v in pd.items()}
```
samsja, line 90: lets rename this to something else than `pd` since this is usually use for pandas

JohannesMessner (approved):
LGTM, love to see PR descriptions like this one! 👍

docarray/array/mixins/io/csv.py
samsja: What I don't understand is why the parser don't just do a return to the line by itself ? Is it because it does not handle it ?

JoanFM:
have we checked if this is a documented feature/behavior of the dependency we use to plot embeddings? if not, we could in parallel open an issue to them to see if we can remove the hack in the future

samsja (approved):

PR #966, opened 2023-01-02
feat: add tensor types in da

samsja:
# Context

This pr allow to precise a tensor type at the DocumentArray level. This allow to stacked document array even if the Document define tensor with`AnyTensor` which is a union.

```python
class Image(Document):
    tensor: AnyTensor # Union[TorchTensor, NdArray]

da = DocumentArray[Image](
    [Image(tensor=torch.zeros(3, 224 ,224)) for _ in range(10)],
    tensor_type=TorchTensor,
).stack()


assert 'tensor' in da._columns.keys()
assert isinstance(da._columns['tensor'], TorchTensor)
```

### DIFF
--- a/docarray/array/array_stacked.py
+++ b/docarray/array/array_stacked.py
@@ -47,9 +47,9 @@ class DocumentArrayStacked(AnyDocumentArray):
     calculation, deep learning forward pass)
 
     A DocumentArrayStacked is similar to {class}`~docarray.array.DocumentArray`
-    but the field of the Document that are {class}`~docarray.typing.AnyTensor` are stacked
-    into a batches of AnyTensor. Like {class}`~docarray.array.DocumentArray`  you
-    can be precise a Document schema by using the `DocumentArray[MyDocument]`
+    but the field of the Document that are {class}`~docarray.typing.AnyTensor` are 
+    stacked into a batches of AnyTensor. Like {class}`~docarray.array.DocumentArray`  
+    you can be precise a Document schema by using the `DocumentArray[MyDocument]`
     syntax where MyDocument is a Document class  (i.e. schema).
     This creates a DocumentArray that can only contains Documents of
     the type 'MyDocument'.
--- a/docarray/array/array_stacked.py
+++ b/docarray/array/array_stacked.py
@@ -47,9 +47,9 @@ class DocumentArrayStacked(AnyDocumentArray):
     calculation, deep learning forward pass)
 
     A DocumentArrayStacked is similar to {class}`~docarray.array.DocumentArray`
-    but the field of the Document that are {class}`~docarray.typing.AnyTensor` are
-    stacked into a batches of AnyTensor. Like {class}`~docarray.array.DocumentArray`
-    you can be precise a Document schema by using the `DocumentArray[MyDocument]`
+    but the field of the Document that are {class}`~docarray.typing.Tensor` are stacked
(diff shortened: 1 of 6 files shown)


docarray/array/array_stacked.py
```diff
@@ -59,40 +59,43 @@ class DocumentArrayStacked(AnyDocumentArray):
...
         for field_name, field in cls.document_type.__fields__.items():
             field_type = field.type_
             if is_union_type(field_type):
-                # cannot stack union fields (might be different types)
-                continue
-
-            is_torch_subclass = (
-                issubclass(field_type, torch.Tensor) if torch_imported else False
-            )
+                if field.type_ == AnyTensor:
```
JohannesMessner, line 86: what about other Union types that are not `AnyTensor`? I think it should still work, we just need to check that the provided `tensor_type` is on of the args of the Union

docarray/array/array_stacked.py
```diff
@@ -110,12 +113,17 @@ def _create_columns(
...
+            if isinstance(type_, type):
+                if issubclass(type_, BaseDocument):
+                    columns[field_to_stack] = DocumentArray.__class_getitem__(type_)(
+                        to_stack
+                    ).stack()
+
+                elif issubclass(type_, (NdArray, TorchTensor)):
+                    columns[field_to_stack] = type_.__docarray_stack__(to_stack)  # type: ignore # noqa: E501
+
+            elif type_ == AnyTensor:  # should be properly handled with union
```
JohannesMessner, line 125: I agree with that comment you left ;)
I think we can, at the beginning, do a check for `is_union_type()` (from the typing_inspect lib), and what is currently inside of `if isinstance(type_, type):` can just come after the `else`.
IMO that is cleaner than having the `isinstance(type_, type)` check

JohannesMessner (changes requested):

docarray/array/array_stacked.py
samsja: You mean if I had a document like this:
```python
class MyDoc(BaseDocument):
    tensor: Union[TorchTensor, NdArray, JaxTensor, SamiTensor]
```

the tensor field should be stacked if I do `DocumentArray(..., tensor_types=TorchTensor)` ?

docarray/array/array_stacked.py
samsja: shit I added this comment to remind me to handle it properly once I rebase on your PR and I forgot to do it

docarray/array/array_stacked.py
samsja: actually I don't agree

samsja: lets say I have a document like this

```python
class MyDoc(BaseDocument):
    data: Union[TorchTensor, bytes]
```

here I really want to express that my data can either be a tensor or a bytes, therefore I should not be able to stack since the data is not homogeneous. The case of `AnyTensor` is arguably different because it is more of a "mode" and user will either use torch or numpy not both at the same time, but I believe it cannot be extended to every Union otherwise we lose the meaning of what is a Union.

samsja: but only the templating solution will fix this problem actually

JohannesMessner: ok let's keep it like this for now, and revisit the templating idea at some point in the future

JohannesMessner (approved):

PR #972, opened 2023-01-03
feat(v2): add video support

anna-charlotte:
Add support for video files to docarray v2

Goals:

- [x] `VideoUrl`
- [x] `VideoTensor` with `VideoNdarray` and `VideoTorchTensor`
- [x] `Video` predefined doc (with optional attrs `Video.tensor`, `Video.key_frame_indices`, `Video.url` and `Video.embedding`)
- [x] tests


docarray/typing/tensor/video/abstract_video_tensor.py
```diff
@@ -0,0 +1,52 @@
...
+T = TypeVar('T', bound='AbstractVideoTensor')
+
+
+class AbstractVideoTensor(AbstractTensor, ABC):
+    @abstractmethod
+    def to_numpy(self) -> np.ndarray:
+        """
+        Convert video tensor to numpy.ndarray.
+        """
+        ...
```
JohannesMessner, line 17: I'd rather have this in the comp backends, otherwise we might com into conflict with methods that are already provided by torch/tf/...

docarray/typing/tensor/video/abstract_video_tensor.py
```diff
@@ -0,0 +1,99 @@
...
+import numpy as np
+
+from docarray.typing.tensor.abstract_tensor import AbstractTensor
+from docarray.typing.tensor.audio.audio_tensor import AudioTensor
+
+T = TypeVar('T', bound='AbstractVideoTensor')
+
+
+class AbstractVideoTensor(AbstractTensor, ABC):
+    def save_to_mp4_file(
```
JohannesMessner, line 13: I would rather just call it `save()`, and in the future potentially allow the format as an argument. Otherwise we might again end up in the situation where we have `save_to_x_file()` for 7 different x

docarray/typing/tensor/video/video_ndarray.py
```diff
@@ -0,0 +1,40 @@
...
+        value: Union[T, np.ndarray, List[Any], Tuple[Any], Any],
+        field: 'ModelField',
+        config: 'BaseConfig',
+    ) -> T:
+        array = super().validate(value=value, field=field, config=config)
+        if array.ndim not in [3, 4] or array.shape[-1] != 3:
+            raise ValueError(
+                f'Expects tensor with 3 or 4 dimensions and the last dimension equal'
+                f' to 3, but received {array.shape} in {array.dtype}'
+            )
```
JohannesMessner, line 38: Can this logic be put into the abstract video tensor? It seems to be repeated here and in the torch one

docarray/typing/url/video_url.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if only_keyframes:
+            return video
+        else:
+            if len(audio_frames) == 0:
+                audio = parse_obj_as(AudioNdArray, np.array(audio_frames))
+            else:
+                audio = parse_obj_as(AudioNdArray, np.stack(audio_frames))
+            indices = parse_obj_as(NdArray, keyframe_indices)
+            return audio, video, indices
```
JohannesMessner, line 153: I am not a fan of having different return types depending on an argument. What about having a separate method for the `only_keyframes` case? What do you think @anna-charlotte @samsja?

JohannesMessner (changes requested):
Did you manage to resolve the quirk where saved video frames would slightly change? Or did you get more insight into why it happens / why it is okay?

docarray/typing/tensor/video/abstract_video_tensor.py
samsja: agree

docarray/typing/url/video_url.py
samsja: agree as well

docarray/typing/tensor/video/video_ndarray.py
JohannesMessner: `get_comp_backend()` is a static method, so you should be able to get it from the `cls` argument in `validate`.

docarray/computation/torch_backend.py
```diff
@@ -64,6 +65,14 @@ def to_device(
     def n_dim(array: 'torch.Tensor') -> int:
         return array.ndim
 
+    @staticmethod
+    def shape(array: 'torch.Tensor') -> Tuple:
+        return array.size()
```
samsja, line 70: I think torch.size return a different object that np.shape. Probably array.shape is better here

docarray/documents/video.py
```diff
@@ -0,0 +1,85 @@
...
+            video=Video(
+                url='https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/toydata/mov_bbb.mp4?raw=true'
+            ),
+            text=Text(text='hello world, how are you doing?'),
+        )
+        mmdoc.video.video_tensor = mmdoc.video.url.load(only_keyframes=True)
+    """
+
+    url: Optional[VideoUrl]
+    audio_tensor: Optional[AudioTensor]
```
samsja, line 82: why not have a Audio object here instead of just the tensor ?

docarray/typing/tensor/video/video_tensor_mixin.py
```diff
@@ -0,0 +1,118 @@
...
+
+class VideoTensorMixin:
+    @staticmethod
+    def validate_shape(
+        cls: Union[Type['VideoTorchTensor'], Type['VideoNdArray']], value: 'T'
+    ) -> 'T':
+        comp_backend = cls.get_comp_backend()
+
+        if (
+            comp_backend.n_dim(value) not in [3, 4]  # type: ignore
```
samsja, line 22: why do we need to ignore the type ?

docarray/typing/url/video_url.py
```diff
@@ -0,0 +1,180 @@
...
+        return cls(str(url), scheme=None)
+
+    def _load(
+        self: T, skip_type: str, **kwargs
+    ) -> Tuple[AudioNdArray, VideoNdArray, NdArray]:
+        """
+        Load the data from the url into a Tuple of AudioNdArray, VideoNdArray and
+        NdArray.
+
+        :param skip_type: determines what video frames to discard.
```
samsja, line 60: can u list the supported string here ? Could be even better to have an Enum

samsja: okay maybe not needed to have an Enum as you pass it to another function directly

samsja (commented):
I added some comments. PR looks great ! Looking forward to have video support

docarray/computation/torch_backend.py
samsja: okay yeah you are right my bad. As you want then

docarray/typing/tensor/video/video_tensor_mixin.py
JohannesMessner: I think this is this issue that has to be fixed using the ugly overloads.

docarray/typing/tensor/video/video_tensor_mixin.py
```diff
@@ -0,0 +1,118 @@
...
+
+if TYPE_CHECKING:
+    from docarray.typing import VideoNdArray, VideoTorchTensor
+
+
+T = TypeVar('T', bound='VideoTensorMixin')
+
+
+class VideoTensorMixin:
+    @staticmethod
```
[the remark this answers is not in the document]
JohannesMessner: If you wan to use a specific superclass implementation, instead of using `super()` you can do something like:

```python
SuperclassThatYouWant.foo(*inputs)
```
Or if you need to pass the subclass `cls`:
```python
SuperclassThatYouWant.foo.__func__(cls, *inputs)
```
This unbinds `foo` from the class, so you can give it any `cls` argument that you want.

You can see this in the `__class_getitem__` implementation of `NdArray`.

docarray/typing/tensor/video/video_tensor_mixin.py
samsja: @JohannesMessner need ur help there we did not manage to debug this. I don't understand why it is not working and why we have different TypeVar in AbstractCompBackend

docarray/documents/video.py
```diff
@@ -0,0 +1,85 @@
...
+            video=Video(
+                url='https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/toydata/mov_bbb.mp4?raw=true'
+            ),
+            text=Text(text='hello world, how are you doing?'),
+        )
+        mmdoc.video.video_tensor = mmdoc.video.url.load_key_frames()
+    """
+
+    url: Optional[VideoUrl]
+    audio: Optional[Audio] = Audio()
```
samsja, line 82: why not None as a default value ?

samsja: okay makes sense

samsja (approved):

PR #990, opened 2023-01-04
fix: stack output of slice access on stacked da

JohannesMessner:
Before this fix, slicing into a stacked DocumentArray would yield a non-stacked da:

```python
da = DocumentArray[Image](Image(url='hi.png', tensor=np.array([i])) for i in range(10))
da = da.stack()
print(da[0:10:2])
print(da[0:10:2].tensor)
```
```text
<docarray.array.abstract_array.DocumentArray[Image] object at 0x7fa6195a9c40>
[NdArray([0]), NdArray([2]), NdArray([4]), NdArray([6]), NdArray([8])]
```

After this fix, a stacked da is returned:
```python
# same code as above
da = DocumentArray[Image](Image(url='hi.png', tensor=np.array([i])) for i in range(10))
da = da.stack()
print(da[0:10:2])
print(da[0:10:2].tensor)
```
```text
<docarray.array.abstract_array.DocumentArrayStacked[Image] object at 0x7f7f07388d90>
[[0]
 [2]
 [4]
 [6]
 [8]]
```


docarray/array/array_stacked.py
```diff
@@ -162,12 +162,24 @@ def _set_array_attribute(
...
             setattr(doc, field, self._columns[field][item])
         return doc
 
+    def _get_slice(self: T, item: slice) -> T:
+        """Return a slice of the DocumentArrayStacked
+
+        :param item: the slice to apply
+        :return: a DocumentArrayStacked
+        """
+        docs = self._docs[item].stack()
```
samsja, line 179: I think there is a problem here when you call stack you will perform the stacking on the the slice of the Nornal DocumentArray but this is useless because this slice has already been stacked into the mother one.

I believe what should be done instead is to create a constructor on DocumentArrayStacked that allow to construct a da stacked from a normal document array (the skeleton) and with the column prebuilt.

somethling lime

```python
@classmethod
def _from_columns(cls: Type[T], docs: DocumentArray, columns: Dict[..., ...])->T.
    da_stacked = DocumentArray[cls.document_type]([]).stack()
    da_stacked._columns = columns
    da_stacked._docs = docs
```
```python
def _get_slice(self: T, item: slice) -> T:
    columns = {k: col[item] for k, col in self._columns.items()}
    docs = self.__cls__._from_columns(self._docs[item], columns)
```

samsja (changes requested):

docarray/array/array_stacked.py
JohannesMessner: So what is the problem with the current implementation?
I agree that it is "useless" (basically a no-op that just converts to the other type), but that is exactly what I want her

tests/units/array/test_array.py
```diff
@@ -228,3 +229,51 @@ class MyDoc(BaseDocument):
...
+    for tensor in tensors:
+        assert tensor.shape == (3, 224, 224)
+
+    texts = da_sliced.text
+    assert len(texts) == 5
+    for i, text in enumerate(texts):
+        assert text == f'hello{i*2}'
+
+
+def test_get_from_slice_stacked():
```
samsja, line 259: I believe we can move this test to the `test_stack_mode` file (which you be renamed btw)

docarray/array/array_stacked.py
```diff
@@ -101,6 +101,10 @@ def _create_columns(
                 ):
                     columns_fields.append(field_name)
 
+        if not columns_fields:
+            # nothing to stack
+            return {}
```
samsja, line 106: but in your case why will the columns field be empty here ? Because it sill looks at the at the schema of the da (i.e, document_type) which should still contains tensor field

samsja (approved):

PR #993, opened 2023-01-05
feat: handle proprely all possible union and optional of tensor in da stack

samsja:
# Context

* refactor code to use is tensor union and handle more Union of tensor types 
* rename Embedding to AnyEmbedding
* rename `util` to `utils`


docarray/array/array_stacked.py
```diff
@@ -97,16 +94,13 @@ def _from_columns(
     def _create_columns(
         cls: Type[T], docs: DocumentArray, tensor_type: Type['AbstractTensor']
     ) -> Dict[str, Union['TorchTensor', T, NdArray]]:
+        from docarray.util._typing import is_tensor_union
```
JohannesMessner, line 97: Why the lazy import?

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -117,3 +117,8 @@ def __docarray_from_native__(cls: Type[T], value: Any) -> T:
     def get_comp_backend() -> Type[AbstractComputationalBackend]:
         """The computational backend compatible with this tensor type."""
         ...
+
+    @staticmethod
+    def __docarray_normalize_value__(val: Any) -> Any:
+        """Normalize a value to be compatible with this tensor type."""
+        return val
```
JohannesMessner, line 124: Side note: I know we do this in other places too, but I think the Python convention is that double-dunder `__cool_method__` methods are methods that are called "magically" under the hood, but can be overwritten.
So maybe this should be called just `docarray_normalize_value()`?

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -117,3 +117,8 @@ def __docarray_from_native__(cls: Type[T], value: Any) -> T:
     def get_comp_backend() -> Type[AbstractComputationalBackend]:
         """The computational backend compatible with this tensor type."""
         ...
+
+    @staticmethod
+    def __docarray_normalize_value__(val: Any) -> Any:
```
JohannesMessner, line 122: It is not clear to me what this method is intended for, even after reading the implementation for TorchTensor. Maybe we can a bit more info to the docstring here?

JohannesMessner (changes requested):

docarray/array/array_stacked.py
samsja: ah I need to fix. I had circular import so I did this

docarray/typing/tensor/abstract_tensor.py
samsja: sure

docarray/typing/tensor/abstract_tensor.py
samsja: I refactor the code. I don't have this normalize like this anymore

docarray/typing/tensor/abstract_tensor.py
samsja: this is refactor

docarray/util/_typing.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any, get_args
```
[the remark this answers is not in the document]
samsja: well I want to package to be private. Nobody should use this function outside of us. You prefer to have everything private ? I was assuming if the package is private then we are good

docarray/computation/abstract_comp_backend.py
```diff
@@ -31,6 +31,13 @@ def stack(
     def n_dim(array: 'TTensor') -> int:
         ...
 
+    @staticmethod
+    @abstractmethod
+    def none_value() -> typing.Any:
+        """Provide a compatible value that represents None in the Tensor Backend."""
+        ...
+        # return torch.tensor(float('nan'))
```
JohannesMessner, line 39: comment

docarray/computation/numpy_backend.py
```diff
@@ -44,6 +44,11 @@ def stack(
     def n_dim(array: 'np.ndarray') -> int:
         return array.ndim
 
+    @staticmethod
+    def none_value() -> Any:
+        """Provide a compatible value that represents None in torch."""
+        return None
```
JohannesMessner, line 50: Why not `np.array(float('nan'))`?

JohannesMessner (changes requested):

docarray/computation/numpy_backend.py
samsja: Numpy do handle None properly. Though we can decide to use Nan as well to do the same as in torch not sure though that it makes sense

JohannesMessner (approved):

PR #996, opened 2023-01-06
feat: to device

JohannesMessner:
Goals:
A feature for stacked da to put all of its tensors on a given device:
```python
class MyDoc(BaseDocument):
    tensor: TorchTensor
    docs: Image

da = DocumentArray[MyDoc](
    [MyDoc(tensor=torch.zeros(3, 5), docs=Image(tensor=torch.zeros(3, 5)))],
    tensor_type=TorchTensor,
)
da = da.stack()
da.to('meta')
assert da.tensor.device == torch.device('meta')
assert da.docs.tensor.device == torch.device('meta')
```


samsja (approved):

docarray/array/array_stacked.py
```diff
@@ -91,6 +91,20 @@ def _from_columns(
...
+    def to(self, device: str):
+        """Move the data to the given device
+
+        :param device: the device to move the data to
+        """
+        for field in self._columns.keys():
+            col = self._columns[field]
+            if isinstance(col, TorchTensor):
+                self._columns[field] = col.get_comp_backend().to_device(col, device)
+            elif isinstance(col, NdArray):
```
JohannesMessner, line 103: are thes isntance checks needed or can we do one check for instance of `AbstractTensor`? the would be cleaner and ensure that we don't forget to update once we add support for more frameworks.

docarray/computation/numpy_backend.py
```diff
@@ -40,6 +41,23 @@ def stack(
     ) -> 'np.ndarray':
         return np.stack(tensors, axis=dim)
 
+    @overload
+    @staticmethod
+    def to_device(tensor: 'NdArray', device: str) -> 'NdArray':
```
JohannesMessner, line 46: Good job figuring this out!

docarray/array/array_stacked.py
Jackmin801: This is a simpler example of this issue
```python
from typing import overload, Callable, Union

@overload
def meow(a: int) -> Callable[[int], int]:
    ...

@overload
def meow(a: str) -> Callable[[str], str]:
    ...

def meow(a):
    if isinstance(a, int):
        return lambda x: 1
    if isinstance(a, str):
        return lambda x: 'str'

# OK
a: str = 's'
meow(a)(a)

# Not ok though it should be
s: Union[int, str] = 's'
meow(s)(s)
```
```python
woof.py:23: error: Argument 1 has incompatible type "Union[int, str]"; expected "int"  [arg-type]
woof.py:23: error: Argument 1 has incompatible type "Union[int, str]"; expected "str"  [arg-type]
```

Jackmin801: Hrmm I see why it doesnt make this connection though. We cannot be guaranteed that `meow` is a pure function. `meow` could change the type of `a` in its implementation and thus assuming `s` is always string in `meow(s)(s)` is invalid.

Jackmin801: In the same way, for
```python
self._columns[field] = col.get_comp_backend().to_device(col, device)
```

We cannot be guaranteed that `col.get_comp_backend()` wont change the type of `col`?

samsja: can we open an issue on mypy and linked it in the code ?

docarray/computation/torch_backend.py
```diff
@@ -1,8 +1,9 @@
-from typing import Any, List, Optional, Tuple, Union
+from typing import Any, List, Optional, Tuple, Union, overload
 
 import torch
 
 from docarray.computation.abstract_comp_backend import AbstractComputationalBackend
+from docarray.typing import TorchTensor
```
samsja, line 6:
```suggestion
ìf TYPE_CHECKING:
    from docarray.typing import TorchTensor
```

docarray/computation/abstract_comp_backend.py
```diff
@@ -37,6 +44,32 @@ def none_value() -> typing.Any:
         """Provide a compatible value that represents None in the Tensor Backend."""
         ...
 
+    # @staticmethod
```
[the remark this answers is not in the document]
JohannesMessner: it's already gone ;)

docarray/typing/tensor/ndarray.py
```diff
@@ -210,3 +210,7 @@ def get_comp_backend() -> Type['NumpyCompBackend']:
         from docarray.computation.numpy_backend import NumpyCompBackend
 
         return NumpyCompBackend
+
+    def __class_getitem__(cls, item: Any, *args, **kwargs):
+        # see here for mypy bug: https://github.com/python/mypy/issues/14123
+        return AbstractTensor.__class_getitem__.__func__(cls, item)  # type: ignore
```
JohannesMessner, line 216: Encountered another mypy bug here

samsja: why is the __func_ needed here ?

JohannesMessner: otherwise `__class_getitem__` is bound to `AbstractTensor` and use that as the `cls` input, which will result in the parametrized type being a `AbstractTensor[shape]` instad of a `NdArray[shape]`.
`__func__` gets us the unbound function, so we can pass `cls` explicitly.

docarray/array/array_stacked.py
```diff
@@ -91,10 +91,30 @@ def _from_columns(
...
+        for field in self._columns.keys():
+            col = self._columns[field]
+            if isinstance(col, AbstractTensor):
+                # the casting below is arbitrary, in reality `col` could be of any
+                # subclass of AbstractTensor. But to make mypy happy we have to cast
+                # it to a concrete subclass thereof
+                # see mypy issue: https://github.com/python/mypy/issues/14421
+                col_ = cast('TorchTensor', col)
+                self._columns[field] = col_.get_comp_backend().to_device(col_, device)
+            elif isinstance(col, NdArray):
```
JohannesMessner, line 108: I think this is better than the many `if` statements because it does not require any changes once we add more ML frameworks (and abstract tensors).
Also, imo, if mypy cannot grasp something for whatever reason, and that requires us to do something a bit whacky, then I would rather keep the whakiness in the typing machinery, since it does not affect the actual code that runs.

samsja (approved):
LGTM maybe we can check the answers from the mypy people before merging : https://github.com/python/mypy/issues/14421

JohannesMessner:
> LGTM maybe we can check the answers from the mypy people before merging : [python/mypy#14421](https://github.com/python/mypy/issues/14421)

I think our inheritance structure doesn't allow it. Basically, every native tensor (`torch.Tensor, np.ndarray` etc.) is mirrored by on our "our" tensor types, but this is impossible to express in the type system in the way that we would need.

docarray/array/array_stacked.py
Jackmin801: Ah in that case we dont even need the elif. The elif never gets used anyways

PR #999, opened 2023-01-08
docs: add more explanation for docarray and n_dim

dongxiang123:
After some confusion in the community, I have added some explanations to the DocumentArray document to make it more clear.
1. Avoid confusion that DocumentArray will load data into memory first when having an external database. 
2. Explain that ElasticSearch will require the same n_dim for its storage.

### DIFF
--- a/docs/advanced/document-store/elasticsearch.md
+++ b/docs/advanced/document-store/elasticsearch.md
@@ -51,6 +51,7 @@ da = DocumentArray(storage='elasticsearch', config={'n_dim': 128})
 
 The usage would be the same as the ordinary DocumentArray, but the dimension of an embedding for a Document must be provided at creation time.
 
+Note for the documents stored in the DocumentArray, they should have the same dimension, otherwise for ElasticSearch.
 ### Secure connection
 
 By default, Elasticsearch server runs with security layer that disables the plain HTTP connection. You can pass the `host` with `api_id` or `ca_certs` inside `es_config` to the constructor. For example,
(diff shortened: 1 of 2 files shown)


hanxiao:
Apart from English, I dont think this is PR necessary. (1) apparently that complained user didnt read docs carefully. DocumentStore is apparently  (2) assuming if they read the docs carefully, this PR with information hiding between nowhere helps nothing in the future.
to prove point 1, the first sentence of the Document Store page, before your PR, is: 
image

hokaso:
> Apart from English, I dont think this is PR necessary. (1) apparently that complained user didnt read docs carefully. DocumentStore is apparently (2) assuming if they read the docs carefully, this PR with information hiding between nowhere helps nothing in the future. to prove point 1, the first sentence of the Document Store page, before your PR, is: image

but the fact is that docs of DocumentStore did not clarify the load-on-read mechanism, when the developer like me use debug to figure it out, which will confuse us, so I think it is necessary to add statement of load-on-read to DocumentStore

hokaso:
as for n_dim, the docs also didn't clarify why embeddings should be consistent when developer use ES in document-store

hokaso:
还有啊，columns、metric这些参数文档里都没说清楚具体干嘛用，什么时候用……表示最后都是一个个靠看源码猜出意思的……

numb3r3:
I agreed that the documentation https://docarray.jina.ai/advanced/document-store/elasticsearch/#config for ES backend configurations can be improved.  We need to double-check the parameter explanations.

PR #1007, opened 2023-01-11
feat: allow variable tensor dimension

Jackmin801:
Goals:

- Implements https://github.com/docarray/docarray/issues/871
- [x] Implement for TorchTensor
- [x] Implement for NdArray

### DIFF
--- a/docarray/computation/torch_backend.py
+++ b/docarray/computation/torch_backend.py
@@ -70,7 +70,7 @@ class TorchCompBackend(AbstractComputationalBackend[torch.Tensor, 'TorchTensor']
         return torch.tensor(float('nan'))
 
     @staticmethod
-    def shape(tensor: Union['torch.Tensor', 'TorchTensor']) -> Tuple[int, ...]:
+    def shape(tensor: 'torch.Tensor') -> Tuple[int, ...]:
         return tuple(tensor.shape)
 
     @overload
--- a/docarray/computation/torch_backend.py
+++ b/docarray/computation/torch_backend.py
@@ -31,7 +31,7 @@ def _usqueeze_if_scalar(t: torch.Tensor):
     return t
 
 
-class TorchCompBackend(AbstractComputationalBackend[torch.Tensor]):
+class TorchCompBackend(AbstractComputationalBackend[torch.Tensor, 'TorchTensor']):
     """
     Computational backend for PyTorch.
     """
@@ -68,6 +68,50 @@ class TorchCompBackend(AbstractComputationalBackend[torch.Tensor]):
     def none_value() -> Any:
         """Provide a compatible value that represents None in torch."""
         return torch.tensor(float('nan'))
+
+    @staticmethod
+    def shape(tensor: Union['torch.Tensor', 'TorchTensor']) -> Tuple[int, ...]:
+        return tuple(tensor.shape)
+
+    @overload
+    @staticmethod
+    def reshape(tensor: 'TorchTensor', shape: Tuple[int, ...]) -> 'TorchTensor':
+        """
+        Gives a new shape to tensor without changing its data.
+
+        :param tensor: tensor to be reshaped
+        :param shape: the new shape
+        :return: a tensor with the same data and number of elements as tensor
+            but with the specified shape.
+        """
+        ...
+
+    @overload
+    @staticmethod
+    def reshape(tensor: 'torch.Tensor', shape: Tuple[int, ...]) -> 'torch.Tensor':
+        """
+        Gives a new shape to tensor without changing its data.
+
+        :param tensor: tensor to be reshaped
+        :param shape: the new shape
+        :return: a tensor with the same data and number of elements as tensor
+            but with the specified shape.
+        """
+        ...
+
+    @staticmethod
+    def reshape(
+        tensor: Union['torch.Tensor', 'TorchTensor'], shape: Tuple[int, ...]
+    ) -> Union['torch.Tensor', 'TorchTensor']:
+        """
+        Gives a new shape to tensor without changing its data.
+
+        :param tensor: tensor to be reshaped
+        :param shape: the new shape
+        :return: a tensor with the same data and number of elements as tensor
+            but with the specified shape.
+        """
+        return tensor.reshape(shape)
 
     class Retrieval(AbstractComputationalBackend.Retrieval[torch.Tensor]):
         """
--- a/docarray/typing/tensor/abstract_tensor.py
+++ b/docarray/typing/tensor/abstract_tensor.py
@@ -119,7 +119,7 @@ class AbstractTensor(Generic[TTensor, T], AbstractType, ABC):
                 f'of shape {tshape} to shape {shape}'
             )
             try:
-                value = cls.__docarray_from_native__(comp_be.reshape(t, shape))
+                value = cls._docarray_from_native(comp_be.reshape(t, shape))
                 return cast(T, value)
             except RuntimeError:
                 raise ValueError(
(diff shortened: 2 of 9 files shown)


Jackmin801:
Do we want to add an example usage in the docstrings?

JohannesMessner (changes requested):
Can this logic be move into the `AbstractTensor` class so that it "just works" when adding a new tensor type? Also I'm guessing variables (`'x'`) are not shared between tensors (or the same Document), right?

JohannesMessner:
> Do we want to add an example usage in the docstrings?

Yes please!

docarray/typing/tensor/ndarray.py
```diff
@@ -83,9 +87,29 @@ def __get_validators__(cls):
     def __docarray_validate_shape__(cls, t: T, shape: Tuple[int]) -> T:  # type: ignore
         if t.shape == shape:
             return t
+        elif any(isinstance(dim, str) for dim in shape):
+            if len(t.shape) != len(shape):
+                raise ValueError(
+                    f"Array shape mismatch. Expected {shape}, got {t.shape}"
```
samsja, line 93: single quote

docarray/typing/tensor/ndarray.py
```diff
@@ -83,9 +87,29 @@ def __get_validators__(cls):
     def __docarray_validate_shape__(cls, t: T, shape: Tuple[int]) -> T:  # type: ignore
```
samsja, line 87: shape can be Tuple[Union[int, str]]

docarray/typing/tensor/ndarray.py
```diff
@@ -83,9 +87,29 @@ def __get_validators__(cls):
...
+        elif any(isinstance(dim, str) for dim in shape):
+            if len(t.shape) != len(shape):
+                raise ValueError(
+                    f"Array shape mismatch. Expected {shape}, got {t.shape}"
+                )
+            known_dims: Dict[str, int] = {}
+            for tdim, dim in zip(t.shape, shape):
+                if isinstance(dim, int) and tdim != dim:
+                    raise ValueError(
+                        f"Array shape mismatch. Expected {shape}, got {t.shape}"
```
samsja, line 99: single quote

docarray/typing/tensor/ndarray.py
```diff
@@ -83,9 +87,29 @@ def __get_validators__(cls):
...
+                    )
+                elif isinstance(dim, str):
+                    if dim in known_dims and known_dims[dim] != tdim:
+                        raise ValueError(
+                            f"Array shape mismatch. Expected {shape}, got {t.shape}"
```
samsja, line 104: single quote

samsja (commented):
would be nice to find a way to factorize the code between torch and numpy

samsja:
> First question, unfortunately not. Numpy and torch have slightly different api, particularly in this case, in the view vs reshape. And yep, the 'x' isn't shared.

cannot we abstract the reshaping part, somewhere ?

JohannesMessner:
> > First question, unfortunately not. Numpy and torch have slightly different api, particularly in this case, in the view vs reshape. And yep, the 'x' isn't shared.
> 
> cannot we abstract the reshaping part, somewhere in the computational backend ?

I suggest the same

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -60,26 +72,43 @@ class AbstractTensor(Generic[ShapeT], AbstractType, ABC):
...
-        - If the shape of `t` is not equal to `shape`,
-            but can be reshaped to `shape`, return `t` reshaped to `shape`.
-        - If the shape of `t` is not equal to `shape`
-            and cannot be reshaped to `shape`, raise a ValueError.
-
-        :param t: The tensor to validate.
-        :param shape: The shape to validate against.
-        :return: The validated tensor.
-        """
-        ...
```
samsja, line 82: lets keep the docstring

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -60,26 +72,43 @@ class AbstractTensor(Generic[ShapeT], AbstractType, ABC):
...
+    def __docarray_validate_shape__(
+        cls, t: T, shape: Tuple[Union[int, str]]
+    ) -> T:  # type: ignore
```
samsja, line 77: why is it type ignore her e?

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -60,26 +72,43 @@ class AbstractTensor(Generic[ShapeT], AbstractType, ABC):
...
+        if t.shape == shape:
```
samsja, line 78: Here can we assume that t has a shape field ? I am not sure maybe it is better to call the `get_comp_backend`

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -60,26 +72,43 @@ class AbstractTensor(Generic[ShapeT], AbstractType, ABC):
...
+                        known_dims[dim] = tdim
+            else:
+                return t
+        else:
+            warnings.warn(
+                f'Tensor shape mismatch. Reshaping tensor '
+                f'of shape {t.shape} to shape {shape}'
+            )
+            try:
+                value = cls.__docarray_from_native__(t.reshape(shape))
```
samsja, line 106: same here I am not sure we can do reshape directly

docarray/typing/tensor/abstract_tensor.py
Jackmin801: Ah its an artifact from copying the signature in TorchTensor. Seems to work fine without it. Removed

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -213,3 +214,18 @@ def _docarray_to_json_compatible(self):
...
+        :return: a tensor with the same data and number of elements as self
+            but with the specified shape.
+        """
+        ...
+
+    @property
+    @abc.abstractmethod
+    def shape(self) -> Tuple[int, ...]:
+        """The shape of this tensor."""
+        ...
```
samsja, line 231: we should move these to the computation backend IMO

docarray/typing/tensor/abstract_tensor.py
Jackmin801: Hrmm ok. Ill do that

docarray/typing/tensor/abstract_tensor.py
Jackmin801: okie

docarray/typing/tensor/abstract_tensor.py
Jackmin801: okei

docarray/computation/torch_backend.py
```diff
@@ -73,6 +73,50 @@ def none_value() -> Any:
...
+        Gives a new shape to tensor without changing its data.
+
+        :param tensor: tensor to be reshaped
+        :param shape: the new shape
+        :return: a tensor with the same data and number of elements as tensor
+            but with the specified shape.
+        """
+        ...
+
+    @overload
```
samsja, line 93: Would be nice to automatically generate the overload. But lets not do it in this PR unless you found a brillant solution haha

Jackmin801: I dont think it works :( https://github.com/python/mypy/issues/11488

samsja: but we could generate the code though

samsja: but yeah lets see later

samsja: nah not code generation via our friend the bot.

But we could have a helper function that generate 3 methods with the correct overload

JohannesMessner: could it be done via decorator somehow?

JohannesMessner (approved):

samsja (approved):

PR #1016, opened 2023-01-13
refactor: optimization on stacking

samsja:
# Context

We recently introduced some refactoring on DocumentArrayStacked : https://github.com/docarray/docarray/pull/1008 which change a little bit the way we handled the stacking (We removed any state where the document does not hold the data but instead some time its hold a view of the data that has been moved to the column)

As noticed by @JohannesMessner this new way of stacking could be optimized because once the data has been stacked we re traverse the DocumentArray to put the view from the column to the document.

_Originally posted by @samsja in https://github.com/docarray/docarray/pull/1008#discussion_r1067114135_ there is room for improvements. 
We could:
* pre-allocate the data for the tensor column (we can predict the shape and the type by looking at the first row of the da)
* when collecting the actual data we could replace the data with a view to the pre alocated column 
* when stacking put the data into the preallocate column in place. Therefore each document has already a view to the good part of the column and we don't need iterate again

# What this PR do:

* Use the optimization explained above
* refactor the column creation code base to make it more concise and more readable

### DIFF
--- a/docarray/array/array.py
+++ b/docarray/array/array.py
@@ -123,7 +123,7 @@ class DocumentArray(AnyDocumentArray):
             # most likely a bug in mypy though
             # bug reported here https://github.com/python/mypy/issues/14111
             return self.__class__.__class_getitem__(field_type)(
-                (getattr(doc, field) for doc in self)
+                (getattr(doc, field) for doc in self), tensor_type=self.tensor_type
             )
         else:
             return [getattr(doc, field) for doc in self]
--- a/docarray/array/array_stacked.py
+++ b/docarray/array/array_stacked.py
@@ -158,7 +158,7 @@ class DocumentArrayStacked(AnyDocumentArray):
                     if val is None:
                         val = tensor_type.get_comp_backend().none_value()
 
-                    columns[field][i] = val
+                    cast(AbstractTensor, columns[field])[i] = val
                     setattr(doc, field, columns[field][i])
                     del val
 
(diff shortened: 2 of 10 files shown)


samsja:
> LGTM, but did we make sure that this is equivalent to `torch.stack()` with no unintended consequnces?

I am not sure to understand the question. What "this" refer to here ?

JohannesMessner:
> > LGTM, but did we make sure that this is equivalent to `torch.stack()` with no unintended consequnces?
> 
> I am not sure to understand the question. What "this" refer to here ?

We talked about this on the other PR; "this" being the technique where we pre-allocate a tensor of desired shape and then set slices of it iteratively. I just want to make sure that that doesn't do anything funky in terms of runtime/memory that `torch.stack()` wouldn't do

samsja:
> > LGTM, but did we make sure that this is equivalent to `torch.stack()` with no unintended consequnces?
>
> I am not sure to understand the question. What "this" refer to here ?

okay now I remember, unfortunately https://discuss.pytorch.org/t/what-is-happening-under-the-hood-when-calling-torch-stack/170105 nobody responded to me on pytorch forums. Though I checked it looks like it is doing the same, and from my understanding it should do the same

JohannesMessner:
> > > LGTM, but did we make sure that this is equivalent to `torch.stack()` with no unintended consequnces?
> > 
> > 
> > I am not sure to understand the question. What "this" refer to here ?
> 
> okay now I remember, unfortunately https://discuss.pytorch.org/t/what-is-happening-under-the-hood-when-calling-torch-stack/170105 nobody responded to me on pytorch forums. Though I checked it looks like it is doing the same, and from my understanding it should do the same

Okay then!

JohannesMessner (approved):

PR #1020, opened 2023-01-16
refactor: renamed docarray private method

samsja:

# Context

rename all `__docarray_meth__` to `_docarray_meth`

### DIFF
--- a/docarray/typing/tensor/abstract_tensor.py
+++ b/docarray/typing/tensor/abstract_tensor.py
@@ -61,7 +61,7 @@ class AbstractTensor(Generic[ShapeT], AbstractType, ABC):
 
     @classmethod
     @abc.abstractmethod
-    def __docarray_validate_shape__(cls, t: T, shape: Tuple[int]) -> T:
+    def _docarray_validate_shape(cls, t: T, shape: Tuple[int]) -> T:
         """Every tensor has to implement this method in order to
         enable syntax of the form AnyTensor[shape].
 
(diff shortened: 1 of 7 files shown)


docarray/typing/tensor/abstract_tensor.py
```diff
@@ -61,7 +61,7 @@ class AbstractTensor(Generic[ShapeT], AbstractType, ABC):
 
     @classmethod
     @abc.abstractmethod
-    def __docarray_validate_shape__(cls, t: T, shape: Tuple[int]) -> T:
+    def _docarray_validate_shape(cls, t: T, shape: Tuple[int]) -> T:
```
JohannesMessner, line 64: I belive this should keep the old name, because it legitimately is a magic method that gets called under the hood, very similar to python built-in magic methods

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -82,7 +82,7 @@ def __docarray_validate_shape__(cls, t: T, shape: Tuple[int]) -> T:
         ...
 
     @classmethod
-    def __docarray_validate_getitem__(cls, item: Any) -> Tuple[int]:
+    def _docarray_validate_getitem(cls, item: Any) -> Tuple[int]:
```
JohannesMessner, line 85: same here

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -115,7 +115,7 @@ class _ParametrizedTensor(
             cls,  # type: ignore
             metaclass=cls.__parametrized_meta__,  # type: ignore
         ):
-            __docarray_target_shape__ = shape
+            _docarray_target_shape = shape
```
JohannesMessner, line 118: and same here

JohannesMessner (changes requested):
i believe methods that are generally not called manually, but are rather part of the under-the-hood machinery can be considered magic methods and should have the dunder at beginning and end. Basically all the methods that one might override when subclassing, but would very rarely call manually.

OTOH, methods that are called in various places and not just in a very specific part of the machinery should not be considered magic methods imo.

Just my POV though ^^

JohannesMessner (approved):

Issue #994, opened 2023-01-05
v2: parametrized tensor shape during operations


JohannesMessner:
When operations on a paremetrized tensor are performed, the supposed shape of that tensor (given in the type hint) does not change, even when the operation does change the actual shape:

```python
class Doc(BaseDocument):
    tensor: TorchTensor[3,1]


d = Doc(tensor=torch.rand(size=(3, 1)))
d_t = d.tensor.transpose(0, 1)
print(d_t)
print(d_t.shape)
```
```text
TorchTensor[3, 1]([[0.7769, 0.8053, 0.0161]])
torch.Size([1, 3])
```

I see two options to tackle this:
1. override `__torch_function__` to correctly assign the current shape to the tensor class after every operation. This would make it that in the example above, the first print would produce `TorchTensor[1, 3]([[0.7769, 0.8053, 0.0161]])`. **Problem with this**: This transformation has to happen at every torch operation, which seems prone for things to go wrong
2. override `__torch_function__` to return a `torch.Tensor` instead of a `TorchTensor`. The challenge with this is that in the current state, this would make our type system useless outside of Document. For example, the following would not type check:
```python
def my_helper(t: TorchTensor[512]):
    ...

class Doc(BaseDocument):
    tensor: TorchTensor[512]

d = Doc(tensor=torch.rand(512))
t = d.tensor + d.tensor  # t is now torch.Tensor
my_helper(t)  # but this wants TorchTensor[512]
```
To make this make sense again, we should take a look at [TorchTyping](https://github.com/patrick-kidger/torchtyping) and see how they achieve proper typing despite the data being `torch.Tensor`.
Big advantage of this: As soon as any operations occur on a `TorchTensor`, it turns into a `torch.Tensor`, meaning there is no opportunity for us to f*ck things up inside of a model.

Preliminary conclusion: Let's do it properly through option 2.

PR #1026, opened 2023-01-17
refactor: instances of parametrized tensors are no longer parametrized

Jackmin801:
This is because parametrizing them can cause them to be wrong when operations are performed.
We only need the parametrization during instantiation for validation.
After that, we make no guarantees about the shape.

Goals:
- Addresses #994 
- Doesnt seem like it is mentioned in the docs. Not sure if its relevant to add a section about this

### DIFF
--- a/docarray/typing/tensor/abstract_tensor.py
+++ b/docarray/typing/tensor/abstract_tensor.py
@@ -67,7 +67,9 @@ class _ParametrizedMeta(type):
         is_tensor = isinstance(instance, AbstractTensor)
         if is_tensor:  # custom handling
             _cls = cast(Type[AbstractTensor], cls)
-            if _cls.__parentcls__:
+            if (
+                _cls.__unparametrizedcls__
+            ):  # This is not None if the tensor is parametrized
                 try:
                     parse_obj_as(_cls, instance)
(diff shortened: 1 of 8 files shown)


docarray/typing/tensor/abstract_tensor.py
```diff
@@ -71,6 +83,7 @@ class AbstractTensor(Generic[TTensor, T], AbstractType, ABC):
 
     __parametrized_meta__: type = _ParametrizedMeta
     _PROTO_FIELD_NAME: str
+    __unparametrizedcls__: Optional[type] = None
```
JohannesMessner, line 86: Couldn't this just be `__docarray_target_shape__` so that we don't need a separate class for paremtrized tensors anymore? And if there is no shape given, this is `None`?

JohannesMessner (changes requested):

docarray/typing/tensor/abstract_tensor.py
Jackmin801: We still need a separate class for parametrized tensors right? Otherwise they cant be treated as types?

samsja (approved):

PR #1029, opened 2023-01-17
Refactor save schema in proto

samsja:
# Context

refactor the proto serialization under the hood


JoanFM (changes requested):
A good test could be to add a simple RPC method receiving DocArrays

samsja:
> A good test could be to add a simple RPC method receiving DocArrays

Isn't it overkill ? As long as the proto can be serialize and deserialize aren't we already sure it will work ?

Or where u reference to the case when we don't have the class in the receiving part ?

docarray/typing/id.py
```diff
@@ -12,6 +14,7 @@
 T = TypeVar('T', bound='ID')
 
 
+@register_proto(proto_type_name='id')
```
JoanFM, line 17: id is a type?

docarray/typing/tensor/audio/audio_torch_tensor.py
```diff
@@ -1,12 +1,14 @@
...
 
+from docarray.typing.proto_register import register_proto
 from docarray.typing.tensor.audio.abstract_audio_tensor import AbstractAudioTensor
 from docarray.typing.tensor.audio.audio_ndarray import MAX_INT_16
 from docarray.typing.tensor.torch_tensor import TorchTensor, metaTorchAndNode
 
 T = TypeVar('T', bound='AudioTorchTensor')
 
 
+@register_proto(proto_type_name='audio_torch_tensor')
```
JoanFM, line 11: the proto type should not be `ndarrayTensor`? and nothing related to `audio`?

JoanFM (changes requested):

docarray/typing/id.py
samsja: yes

docarray/typing/tensor/audio/audio_torch_tensor.py
samsja: but you need to store the information that this is a AudioTensor right ?

JoanFM: okey I understand now

docarray/typing/proto_register.py
```diff
@@ -0,0 +1,38 @@
+from typing import Callable, Type
+
+from docarray.typing.abstract_type import AbstractType
+
+_PROTO_TYPE_NAME_TO_CLASS = {}
+
+
+def register_proto(
```
JohannesMessner, line 8: I agree that we should use the class name and avoid this mechanism

JohannesMessner (changes requested):

JohannesMessner:
@samsja raised a point that renaming classes in the future (as part of a refactoring) would break backwards and forwards compatibility of the proto between docarray versions. So let's go with an explicit key instead of the class name

JohannesMessner (approved):
private decorators look so weird, `@_whatever` 🥲

tests/units/document/proto/test_proto_based_object.py
```diff
@@ -6,7 +6,7 @@
 
 def test_nested_item_proto():
     NodeProto(text='hello')
-    NodeProto(nested=DocumentProto())
+    NodeProto(document=DocumentProto())
```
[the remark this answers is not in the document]
samsja: good catch. The tests are useless I removed them

PR #1030, opened 2023-01-17
refactor: comp backend 

samsja:
# Context

Because of some type problem we need multiples times in the codebase to duplicate function definition in the Computional Backend part.

This PR fix this by doing a refactoring on the type part. The new one is less precise and correct that before but more flexible and easier to maintain.

Main difference are two folds:

* Use Instance instead of class in get_comp_backend (to fix the static method pb)
* The return type of the function in the computational backend is always np.ndarray or torch.Tensor. Therefore mypy cannot now that if you passed a TorchTensor (from us) the method will return a TorchTensor. This is a limitation but it allow us to remove the second TypeVar Generic in the Computational backend. This means in some case we need either to tell mypy we know the return type will be a TorchTensor by calling cast or TorchTensor._docarray_from_native()

### DIFF
--- a/docarray/typing/tensor/abstract_tensor.py
+++ b/docarray/typing/tensor/abstract_tensor.py
@@ -202,7 +202,7 @@ class AbstractTensor(Generic[TTensor, T], AbstractType, ABC):
 
     @staticmethod
     @abc.abstractmethod
-    def get_comp_backend() -> AbstractComputationalBackend[TTensor, T]:
+    def get_comp_backend() -> AbstractComputationalBackend:
         """The computational backend compatible with this tensor type."""
         ...
 
--- a/docarray/typing/tensor/abstract_tensor.py
+++ b/docarray/typing/tensor/abstract_tensor.py
@@ -88,7 +88,7 @@ class AbstractTensor(Generic[TTensor, T], AbstractType, ABC):
         :param shape: The shape to validate against.
         :return: The validated tensor.
         """
-        comp_be = t.get_comp_backend()()  # mypy Generics require instantiation
+        comp_be = t.get_comp_backend()  # mypy Generics require instantiation
         tshape = comp_be.shape(t)
         if tshape == shape:
             return t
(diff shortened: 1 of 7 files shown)


docarray/typing/tensor/abstract_tensor.py
```diff
@@ -88,7 +88,7 @@ def __docarray_validate_shape__(cls, t: T, shape: Tuple[Union[int, str]]) -> T:
         :param shape: The shape to validate against.
         :return: The validated tensor.
         """
-        comp_be = t.get_comp_backend()()  # mypy Generics require instantiation
+        comp_be = t.get_comp_backend()  # mypy Generics require instantiation
```
Jackmin801, line 91:
```suggestion
        comp_be = t.get_comp_backend()
```

The comment isnt true anymore because it is already instantiated by the method

Jackmin801:
Other than that LGTM

docarray/typing/tensor/ndarray.py
```diff
@@ -202,11 +202,11 @@ def to_protobuf(self) -> 'NdArrayProto':
         return nd_proto
 
     @staticmethod
-    def get_comp_backend() -> Type['NumpyCompBackend']:
+    def get_comp_backend() -> 'NumpyCompBackend':
         """Return the computational backend of the tensor"""
         from docarray.computation.numpy_backend import NumpyCompBackend
 
-        return NumpyCompBackend
+        return NumpyCompBackend()
```
JohannesMessner, line 209: So this is the only main difference here, right?

JohannesMessner, line 209: Is it a bit of an anti pattern to crate all of these instances if we don't really need them? would it make sense to maintain a singleton instance?

Jackmin801: Oh that is a good idea actually. I support this. Lets go singleton with this.

samsja: Main difference are two folds:
* Use Instance instead of class in get_comp_backend (to fix the static method pb)
* The return type of the function in the computational backend is always np.ndarray or torch.Tensor. Therefore mypy cannot now that if you passed a TorchTensor (from us) the method will return a TorchTensor. This is a limitation but it allow us to remove the second TypeVar Generic in the Computational backend. This means in some case we need either to tell mypy we know the return type will be a TorchTensor by calling cast or TorchTensor._docarray_from_native()

samsja: actually doing the singleton is not that trivial in python ... I think it adds complexity for few real advantages at the end. Only conceptual advantage.

Issue #1033, opened 2023-01-18
Bug: incompatibility between different torch tensor subtypes


JohannesMessner:
How to reprodoce:

```python
import torch

from docarray.typing import AudioTorchTensor, TorchEmbedding

t1, t2 = AudioTorchTensor(torch.rand(128)), TorchEmbedding(128)
print(f'{type(t1)=}')
print(f'{type(t2)=}')

t1 + t2
```
```bash
type(t1)=<class 'docarray.typing.tensor.audio.audio_torch_tensor.AudioTorchTensor'>
type(t2)=<class 'docarray.typing.tensor.embedding.torch.TorchEmbedding'>
Traceback (most recent call last):
  File "/home/johannes/.config/JetBrains/PyCharmCE2022.3/scratches/scratch_121.py", line 9, in <module>
    t1 + t2
TypeError: unsupported operand type(s) for +: 'AudioTorchTensor' and 'TorchEmbedding'

```

PR #1037, opened 2023-01-18
fix: torch subtypes interoperability

JohannesMessner:
Goals:

Fix the bug described in #1033

### DIFF
--- a/docarray/typing/tensor/torch_tensor.py
+++ b/docarray/typing/tensor/torch_tensor.py
@@ -230,3 +230,14 @@ class TorchTensor(
         from docarray.computation.torch_backend import TorchCompBackend
 
         return TorchCompBackend
+
+    @classmethod
+    def __torch_function__(cls, func, types, args=(), kwargs=None):
+        # this tells torch to treat all of our custom tensors just like
+        # torch.Tensor's. Otherwise, torch will complain that it doesn't
+        # know how to handle our custom tensor type.
+        docarray_torch_tensors = TorchTensor.__subclasses__()
+        types_ = tuple(
+            torch.Tensor if t in docarray_torch_tensors else t for t in types
+        )
+        return super().__torch_function__(func, types_, args, kwargs)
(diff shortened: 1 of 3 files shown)


docarray/typing/tensor/torch_tensor.py
```diff
@@ -221,3 +221,14 @@ def get_comp_backend() -> Type['TorchCompBackend']:
...
+    @classmethod
+    def __torch_function__(cls, func, types, args=(), kwargs=None):
+        # this tells torch to treat all of our custom tensors just like
+        # torch.Tensor's. Otherwise, torch will complain that it doesn't
+        # know how to handle our custom tensor type.
+        docarray_torch_tensors = TorchTensor.__subclasses__()
+        types_ = tuple(
+            torch.Tensor if t in docarray_torch_tensors else t for t in types
+        )
+        return super().__torch_function__(func, types_, args, kwargs)
```
samsja, line 234: did u get more insight on what is happening here ?

samsja (approved):
lgtm

docarray/typing/tensor/torch_tensor.py
JohannesMessner: My current understanding is this:
the `types` argument contains all the types of all the tensors that are involved in a given operation.
They are there just for convenience, they don't actually do anything in terms of funcitonality. Their main reasons seems to be that someone implementing `__torch_function__` can conveniently access them to determine if their function/method implementation can handle objects of that type.

The trouble is that the default PyTorch impelementation determines that in the following way: "If all types in `types` are subtypes of `cls` then we're good. If not I don't know what to do with it, so raise an Exception". And this is what happens when operating on `TorchEmbedding` and `AudioTorchTensor`, since they are not subclasses of each other.

So what we do here by passsing `(torch.Tensor,)` as `types` is to tell pytorch: "you _do_ know what to do with these, just treat them like normal `torch.Tensor`!"

If you remember the old PR with a fix similar to this, I think that one was unsafe. There we essentially said "whatever object you get, treat it like torch.Tensor". In this version we make that claim only for subclasses of `TorchTensor`, which is actually what we want.

JohannesMessner: good point, will do

samsja: then solution sounds perfect

PR #1043, opened 2023-01-19
feat(v2): rich display for doc and da

anna-charlotte:
Goals:

- [x] Add `.summary()` to for Document and DocumentArray.
- [x] Add pretty print for Document as well as DocumentArray.

Pretty prints:
image

For `Document`:
- `doc.summary()` to display the Document instance
  - key, value (including type and shape)
  - maybe display shorter id: `2184b5 ...` instead of `2184b53f977f566d72f72a6e706edb00 `
  - if eg. list of 200 elements is part of the Document, don't show all 200 elements, keep max size of such display manageable. 
Example:
image

- `doc.schema_summary()` to display the Document's schema

For `DocumentArray`:
- schema and length
- redundant fields for v2: 
  - `Homogenous Documents`
  - `Common Attributes`, instead show the schema
  - `Multimodal dataclass`
  
image


samsja:
@anna-charlotte I believe we should tackle pretty printing in this PR as well. 

at the moment pritting a document array will lool like this `<docarray.array.abstract_array.DocumentArray[Document][MyDoc] at 0x7fb93800b5e0>` which is not very informative. Same apply to document

docarray/base_document/document.py
```diff
@@ -34,3 +41,169 @@ def _get_field_type(cls, field: str) -> Type['BaseDocument']:
...
+        tree = Tree(name) if doc_name is None else Tree(f'{doc_name}: {name}')
+
+        for k, v in cls.__annotations__.items():
+
+            field_type = cls._get_field_type(k)
+
+            t = str(v).replace('[', '\[')
+            t = re.sub('[a-zA-Z_]*[.]', '', t)
+
+            if v.__name__ in ['Union', 'Optional']:
```
samsja, line 80: we have a way to check if a field is a union somewhere in the code

samsja:
```python
from typing_inspect import is_union_type
is_union_type(field_type)
```

docarray/math/helper.py
```diff
@@ -0,0 +1,34 @@
+from typing import Optional, Tuple
+
+import numpy as np
+
+
+def minmax_normalize(
```
samsja, line 6: couples of questions:

* Is this intended only to work for NumPy array? Why not torch as well? Otherwise, it should be implemented in the computational backend

* If we decide to stick to only NumPy and therefore we don't put it in the computational backend then we should move this to utils.helper or smth like this

docarray/base_document/mixins/plot.py
```diff
@@ -0,0 +1,199 @@
...
+
+        import torch
+        from rich import box, text
+        from rich.table import Table
+
+        import docarray
+
+        table = Table('Attribute', 'Value', width=80, box=box.ROUNDED, highlight=True)
+
+        for k, v in self.__dict__.items():
```
samsja, line 89: let's use proper variable name and not k,v. I suggest field_name and value

docarray/base_document/mixins/plot.py
```diff
@@ -0,0 +1,199 @@
...
+            ):
+                continue
+            elif isinstance(v, str):
+                col_2 = str(v)[:50]
+                if len(v) > 50:
+                    col_2 += f' ... (length: {len(v)})'
+                table.add_row(col_1, text.Text(col_2))
+            elif isinstance(v, (np.ndarray, torch.Tensor)):
+                if isinstance(v, torch.Tensor):
+                    v = v.detach().cpu().numpy()
```
samsja, line 104: why do we need to do this? Detach can makes sense but to cpu and to numpy I am not sure to understand

JohannesMessner:
@anna-charlotte could you please link this PR/Issue to the dev roadmap and tick it off once merged?

JohannesMessner (approved):
Honestly I am lost in all of this `rich` stuff, but the screenshots look pretty, so that's a thumbs up from me ^^

tests/units/array/test_array_stacked.py
```diff
@@ -22,6 +22,10 @@ class Image(BaseDocument):
     return batch.stack()
 
 
+def test_repr(batch):
+    assert batch.__repr__() == '<DocumentArrayStacked[Image] (length=10)>'
+
```
samsja, line 27: same not needed

docarray/base_document/mixins/plot.py
```diff
@@ -0,0 +1,234 @@
...
+        kls = self.__class__.__name__
+        id_abbrv = getattr(self, 'id')[:7]
+        yield f":page_facing_up: [b]{kls}" f"[/b]: [cyan]{id_abbrv} ...[cyan]"
+
+        from rich import box, text
+        from rich.table import Table
+
+        from docarray import BaseDocument, DocumentArray
+
+        table = Table('Attribute', 'Value', width=80, box=box.ROUNDED, highlight=True)
```
samsja, line 100: can we save this 80 into a class variable ?

samsja (changes requested):
lets not merge until we do the refactoring with class that wrap our concept with the display ability @anna-charlotte

docarray/base_document/document.py
```diff
@@ -34,3 +35,10 @@ def _get_field_type(cls, field: str) -> Type['BaseDocument']:
         :return:
         """
         return cls.__fields__[field].outer_type_
+
+    def __str__(self):
+        console = rich.console.Console()
```
samsja, line 40: do we need to create a console each time or can we define it once ? (just asking)

samsja: yes I link the later better only one console object at the module level

docarray/base_document/document.py
```diff
@@ -2,21 +2,23 @@
...
 
 
-class BaseDocument(BaseModel, ProtoMixin, AbstractDocument, BaseNode):
+class BaseDocument(BaseModel, PlotMixin, ProtoMixin, AbstractDocument, BaseNode):
     """
     The base class for Document
     """
 
     id: ID = Field(default_factory=lambda: parse_obj_as(ID, os.urandom(16).hex()))
+    _console: rich.console.Console = rich.console.Console()
```
samsja, line 21: lets put it at the module level

samsja (changes requested):

samsja (approved):

PR #1044, opened 2023-01-23
fix: fix csv string parsing and memory leak in dataset

Jackmin801:
- `line.split(",")` does not take into account cases where there are commas in the caption
- the copy in getitem isnt deep, causing a memory leak
- The maximum token length of captions in flickr8k is 43. Padding up to 512 wastes a lot of VRAM which could be used to increase the batch size for more stable training.


samsja (approved):

samsja (changes requested):
can u add pip install pandas in the begining of the notebook ?

Jackmin801:
Oh I think jupyter notebooks come with pandas tho

samsja (approved):

JohannesMessner (approved):

Issue #1041, opened 2023-01-19
Generic DocumentArray does not work with isinstance and isubclass


samsja:
# Context

We cannot check if a da belongs to its class with our templating.

```python
from docarray import DocumentArray, BaseDocument


class MyDoc(BaseDocument):
    text: str

docs = DocumentArray[MyDoc]([MyDoc(text='hello')])

assert isinstance(docs, DocumentArray[MyDoc])
>>> AssertError
assert issubclass(docs.__class__, DocumentArray[MyDoc])
>>> AssertError
```

JohannesMessner:
This issue is closely related to [this](https://github.com/docarray/docarray/pull/991/) one and can be tackled in the same way

PR #1046, opened 2023-01-24
fix: typed da classing

Jackmin801:
Goals:

- Addresses #1041 
- This solution is different from the one suggested by @JohannesMessner in https://github.com/docarray/docarray/pull/991/
  - This solution is cleaner imo


JohannesMessner:
So if I understand correctly, this basically caches the parametrized classes so that they will not be recreated when it is not necessaray, right? Are there any scenarios where we could end up with two class "instances" with two different caches, such that the original problem could still occur? Maybe through multi-threading/processing? Not sure myself...

Jackmin801:
I dont think worker processes can return a class that was defined in the worker process anyways. So even if different instances existed, it would be illegal for them to interact.

docarray/array/abstract_array.py
```diff
@@ -16,35 +26,39 @@
 class AnyDocumentArray(Sequence[BaseDocument], Generic[T_doc], AbstractType):
     document_type: Type[BaseDocument]
     tensor_type: Type['AbstractTensor'] = NdArray
+    __typed_da__: Dict[Type[BaseDocument], Type] = {}
```
samsja, line 29:
```suggestion
    __typed_da__: Dict[Type[BaseDocument], Type] = dict()
```

samsja (approved):
neat

samsja (changes requested):
oups I forgot that I ask for a change

docarray/array/abstract_array.py
Jackmin801: Is there an error case with using `{}`? I thought it was faster by a few milliseconds?

samsja: Screenshot from 2023-01-25 11-52-36

it is not safe to use {} or [] inside function argument default value

Jackmin801: But its a static variable... its not a default arg

samsja (approved):

Issue #1042, opened 2023-01-19
Pytorch dataset and dataloader


samsja:
# Context

in the clip training notebook : https://github.com/docarray/docarray/blob/feat-rewrite-v2/docs/tutorials/mutlimodal_training_and_serving.md#create-the-dataset we introduced a Pytorch dataset.

Most of the code for this dataset is agnostic from the data schema.

What it does it calling a preprocessing function for each modality and return a Document. There is the collate function as well that is very generic that just return a batch of DocumentArrayStacked instead of just a tensor.

We could provide tools to build this datasets faster. 

At the end of the day the user just need to define a mapping that define for each modality/fields a processing function.

The interface and the scope of the object we could give to the user is still not decided. But we can envision something like this

```python
from docarray import BaseDocument, DocumentArray
from docarray.documents import Image, Text

class MyDoc(BaseDocument):
   image: Image
   text: Text

preprocessing = {Image: lambda x: x.url.load(), Text: tokenizer}

from docarray.utils.data import Dataset

class Mydataset(Dataset[MyDoc]):
    def __init__(self, csv_file: str, preprocessing : Dict[BaseDocument, Callable]):
       super().__init__(preprocessing)
      
      for line in open(csv_file): # custom logic here
         self.da.append(...)

from torch.utils.data import DataLoader


dataset = Mydataset("captions.csv", preprocessing)
loader = DataLoader(
    dataset, batch_size=64, collate_fn=dataset.collate_fn, shuffle=True
)


for i, batch in enumerate(loader):
   assert isinstance(batch, DocumentArrayStacked[MyDoc])
   ...
```

PR #1049, opened 2023-01-24
feat: TorchDataset

Jackmin801:
Goals:

- Address #1042 
- [x] design appropriate API
- [x] single process works
- [x] multiprocessing works
~- [ ] gpu memory pin works~ # Not a significant slowdown at the moment


tests/units/data/test_torch_dataset.py
```diff
@@ -0,0 +1,59 @@
...
+            with open(csv_file, "r") as f:
+                f.readline()
+                da = DocumentArray[PairTextImage](
+                    PairTextImage(
+                        text=Text(text=i[1]),
+                        image=Image(url=f"tests/toydata/image-data/{i[0]}"),
+                    )
+                    for i in map(lambda x: x.strip().split(","), f.readlines())
+                )
+            super().__init__(da, preprocessing)
```
JohannesMessner, line 40: can we switch this test to directly use `TorchDataset[PairTextImage]` se we reflect this pattern that we want to promote?
(actually this would probably also not pass mypy now that I think about it, since the init args are different)

JohannesMessner (changes requested):

tests/units/data/test_torch_dataset.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+
+    from docarray.array.array_stacked import DocumentArrayStacked
+
+    # TEMPORARY PATCH WHILE WE WAIT FOR ISINSTANCE TO BE FIXED
+    isinstance = lambda x, y: str(x.__class__) == str(y)  # noqa: E731
+
+    for batch in loader:
+        assert isinstance(batch, DocumentArrayStacked[PairTextImage])
+        assert len(batch) == BATCH_SIZE
```
samsja, line 59: won't be always true. If len of da is not a multiple of BATCH_SIZE the last batch can be smaller

tests/units/data/test_torch_dataset.py
```diff
@@ -0,0 +1,59 @@
...
+    loader = DataLoader(
+        dataset, batch_size=BATCH_SIZE, collate_fn=dataset.collate_fn, shuffle=True
+    )
+
+    from docarray.array.array_stacked import DocumentArrayStacked
+
+    # TEMPORARY PATCH WHILE WE WAIT FOR ISINSTANCE TO BE FIXED
+    isinstance = lambda x, y: str(x.__class__) == str(y)  # noqa: E731
+
+    for batch in loader:
```
JoanFM, line 57: also if loader does not contain anything, this test would fail

tests/units/data/test_torch_dataset.py
```diff
@@ -0,0 +1,59 @@
...
+            with open(csv_file, "r") as f:
+                f.readline()
+                da = DocumentArray[PairTextImage](
+                    PairTextImage(
+                        text=Text(text=i[1]),
+                        image=Image(url=f"tests/toydata/image-data/{i[0]}"),
+                    )
+                    for i in map(lambda x: x.strip().split(","), f.readlines())
+                )
+            super().__init__(da, preprocessing)
```
JoanFM, line 40: I am not sure to like this pattern so much? Shouldnt' maybe we ask them to implement a `da` property or something like this?

tests/units/data/test_torch_dataset.py
Jackmin801: ah yup. good catch

tests/units/data/test_torch_dataset.py
Jackmin801: does it make sense to have a loader that has nothing in it? I think the dataloader initialisation fails first.

tests/units/data/test_torch_dataset.py
Jackmin801: Yup. Done.
Does it not pass mypy? I tried it and it seems fine.

tests/units/data/test_torch_dataset.py
Jackmin801: Yea, the rest of the team didn't like it either. We decided to go with directly using the `MultiModalDataset` instead. So they have a `da` and a preprocessing dictionary and we give them a dataset.

```python
da = ...
preprocessing = ...
ds = MultiModalDataset[PairTextImage](da, preprocessing)
dl = DataLoader(ds, ...)
...
```

Jackmin801: If they want something more custom, they can do so. They can still use our collate function.

```python
ds = MyCustomDataset(*args)
loader = DataLoader(
    ds,
    collate_fn=MultiModalDataset[PairTextImage].collate_fn,
    ...
)
```

docarray/data/torch_dataset.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    document_type: Optional[Type[BaseDocument]] = None
+    __typed_ds__: Dict[Type[BaseDocument], Type['MultiModalDataset']] = {}
+
+    def __init__(
+        self, da: 'DocumentArray[T_doc]', preprocessing: Dict[str, Callable]
+    ) -> None:
+        self.da = da
+        self._preprocessing = preprocessing
```
JohannesMessner, line 27: I think you already mentioned in the daily that you need to add documentation, let's make sure that we clearly convey the `k`s and `v`s of `preprocessing`, the inputs and outputs of the functions in its `v`s, etc.

docarray/data/torch_dataset.py
```diff
@@ -0,0 +1,73 @@
...
+from docarray.typing import TorchTensor
+
+T_doc = TypeVar('T_doc', bound=BaseDocument)
+
+
+class MultiModalDataset(Dataset, Generic[T_doc]):
+    """
+    A dataset that can be used inside a PyTorch DataLoader.
+    In other words, it implements the PyTorch Dataset interface.
+
```
JohannesMessner, line 15: Maybe a small usage example here would also be nice

JohannesMessner (changes requested):

tests/units/data/test_torch_dataset.py
JohannesMessner: you ran mypy on this test file? Then I'm wrong haha, I expected it to fail

docarray/data/torch_dataset.py
Jackmin801: Yup. It became kind of a long docstring but I suppose that it ok

samsja:
@Jackmin801 don't forget to add integration test to test that the dataloader is working with dataset

docarray/data/torch_dataset.py
```diff
@@ -138,10 +139,23 @@ def __class_getitem__(cls, item: Type[BaseDocument]) -> Type['MultiModalDataset'
             )
 
         if item not in cls.__typed_ds__:
+            global _TypedDataset
```
JoanFM, line 142: I do not like `globals` at all, what is the need for this?

docarray/utils/_typing.py
```diff
@@ -19,3 +19,16 @@ def is_tensor_union(type_: Any) -> bool:
         return is_union and all(
             (is_type_tensor(t) or issubclass(t, type(None))) for t in get_args(type_)
         )
+
+
+def change_cls_name(cls: type, new_name: str, scope=None) -> None:
```
JoanFM, line 24: type annotation of `scope`?

tests/units/data/test_torch_dataset.py
JoanFM: I mean, just assert something that would not fail if the `loader` is empty

JoanFM (changes requested):

docarray/array/abstract_array.py
```diff
@@ -27,18 +28,24 @@
...
                 f'{cls.__name__}[item] item should be a Document not a {item} '
             )
 
-        if item not in cls.__typed_da__:
+        if cls not in cls.__typed_da__:
+            cls.__typed_da__[cls] = {}
+
+        if item not in cls.__typed_da__[cls]:
+            # Promote to global scope so multiprocessing can pickle it
+            global _DocumentArrayTyped
```
JoanFM, line 48: I do not like this solution at all, I see `globals` as very dangerous `smells`

docs/tutorials/multimodal_training_and_serving.md
```diff
@@ -179,56 +182,35 @@ class TextPreprocess:
...
+    )
+    return da
 ```
 
-In the `PairDataset` class we can already see some of the beauty of DocArray.
-The dataset will return Documents that contain the text and image data, accessible via `doc.text` and `doc.image`.
+In the `get_flickr8k_da` method we process the Flickr8k dataset into a `DocumentArray`.
 
-Now let's instantiate this dataset:
+Now let's instantiate this dataset using the `MultiModalDataset` class. The constructor takes in the da and a dictionary of preprocessing transformations:
```
alexcg1, line 196:
```suggestion
Now let's instantiate this dataset using the `MultiModalDataset` class. The constructor takes in the `da` and a dictionary of preprocessing transformations:
```

alexcg1 (changes requested):

alexcg1 (approved):
LGTM :+1:

samsja (approved):

JoanFM (approved):

JohannesMessner (approved):

PR #1051, opened 2023-01-25
feat: add filter capability to DocumentArray

JoanFM:
Goals:

Add filtering capability with the same QueryLanguage as in v1 docarray.

TODO:

- [x] Implement the `operators` in every Document Type 
- [x] Test


docarray/utils/lookup.py
```diff
@@ -0,0 +1,290 @@
+"""
+
+Originally from https://github.com/naiquevin/lookupy
```
JohannesMessner, line 3: why do we copy/paste this instead of depending on it?

JoanFM: we have some edits to do

docarray/utils/filter.py
```diff
@@ -0,0 +1,36 @@
...
+    EXAMPLE USAGE
+
+    .. code-block:: python
+
+        TODO: fill it
+
+    :param index: the index of Documents to filter in
+    :param query: the query to filter by
+    :return: A DocumentArray containing the Documents
+    inside DocumentArray that fullfil the filter conditions
```
JohannesMessner, line 27:
```suggestion
    :return: A DocumentArray containing the Documents
    in `index` that fulfill the filter conditions in the `query`
```

docarray/utils/filter.py
```diff
@@ -0,0 +1,36 @@
...
+) -> AnyDocumentArray:
+    """
+    Filter the Documents in the index according to the given filter query.
+
+
+    EXAMPLE USAGE
+
+    .. code-block:: python
+
+        TODO: fill it
```
JohannesMessner, line 22: Will the QL be explained here?

docarray/utils/lookup.py
```diff
@@ -0,0 +1,325 @@
...
+
+from functools import partial
+
+if TYPE_CHECKING:  # pragma: no cover
+    from docarray import BaseDocument
+
+PLACEHOLDER_PATTERN = re.compile(r'\{\s*([a-zA-Z0-9_]*)\s*}')
+
+
+def dunder_get(_dict: Any, key: str) -> Any:
```
JohannesMessner, line 40: I think we have to unify this, if i am not wrong for nested access we use `a.b` instead of `a__b`. Personally i prefer the dunder bc it is more string-y, but it should be the same everywhere

docarray/utils/lookup.py
```diff
@@ -0,0 +1,325 @@
...
+class LookupyError(Exception):
+    """Base exception class for all exceptions raised by lookupy"""
+
+    pass
+
+
+## utility functions
+
+
+def dunder_partition(key):
```
JohannesMessner, line 281: these utility functions probably come from the other library, but should we add type hints to them? we are pretty strict about that in the rest of the codebase

docarray/utils/query_parser.py
```diff
@@ -0,0 +1,124 @@
...
+    **COMPARISON_OPERATORS,
+    **ARRAY_OPERATORS,
+    **REGEX_OPERATORS,
+    **MEMBERSHIP_OPERATORS,
+}
+
+
+def _parse_lookups(
+    data: Union[Dict, List] = {}, root_node: Optional[LookupTreeElem] = None
+):
```
JohannesMessner, line 40: return type?

JohannesMessner (changes requested):

docarray/utils/lookup.py
JoanFM: I have checked, and MongoDB does `.` access, and since the QueryLanguage is given by MongoDB, I would change

docarray/base_document/document.py
```diff
@@ -43,3 +43,6 @@ def __str__(self):
             _console.print(self)
 
         return capture.get().strip()
+
+    def get_string_for_regex_filter(self):
```
samsja, line 47:
```suggestion
    def _get_string_for_regex_filter(self):
```

docarray/utils/lookup.py
JoanFM: I will try to do so

docarray/utils/filter.py
JoanFM: Will apply in another commit

docarray/documents/text.py
```diff
@@ -65,6 +65,22 @@ class MultiModalDoc(BaseDocument):
...
+
+    .. code-block:: python
+
+        from docarray.documents Text
+
+        doc = Text(text='This is the main text', url='exampleurl.com')
+        doc2 = Text(text='This is the main text', url='exampleurl.com')
+
+        doc == 'This is the main text' # True
+        doc == doc2 # False, their ids are not equivalent
```
samsja, line 83: isn't this a problem btw ? I would expect it to be true even thouhg id are sot similar

docarray/documents/text.py
```diff
@@ -79,3 +95,19 @@ def validate(
...
         return super().validate(value)
+
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, str):
+            return self.text == other
+        else:
+            # BaseModel has a default equality
+            return super().__eq__(other)
+
+    def __contains__(self, item: str):
```
samsja, line 106: can u add to the docstring how this look lile ?

docarray/utils/lookup.py
```diff
@@ -0,0 +1,326 @@
+"""
```
samsja, line 1: can we create a folder filter inside utils where we put the 3 files filter lookup and query parser ?

docarray/base_document/document.py
JoanFM: It will be accessed by an external method and `mypy` or `ruff` may complain?

docarray/documents/text.py
JoanFM: I think this can be discussed in another PR, I would separate the concern, for now I just added tests to show explicitly the behavior

docarray/documents/text.py
JoanFM: what do u mean?

docarray/utils/lookup.py
JoanFM: I will

docarray/base_document/document.py
JoanFM: Okey, will be applied in new commit

JohannesMessner (approved):

PR #1052, opened 2023-01-25
feat: display columns information for stacked array

anna-charlotte:

Goals:

For a `DocumentArrayStacked` instance add some information of the stacked columns to its summary, such as the following:

image

Handle the following cases:
- [x] stacked array is None
- [x] stacked array is NdArray
- [x] stacked array is TorchTensor: additionally show device

### DIFF
--- a/docarray/computation/abstract_comp_backend.py
+++ b/docarray/computation/abstract_comp_backend.py
@@ -112,7 +112,7 @@ class AbstractComputationalBackend(ABC, typing.Generic[TTensor]):
 
     @staticmethod
     @abstractmethod
-    def dtype(tensor: 'TTensor'):
+    def dtype(tensor: 'TTensor') -> Any:
         """Get the data type of the tensor."""
         ...
 


docarray/computation/abstract_comp_backend.py
```diff
@@ -104,14 +104,26 @@ def detach(tensor: 'TTensor') -> 'TTensor':
         """
         ...
 
+    @staticmethod
+    @abstractmethod
+    def dtype(tensor: 'TTensor'):
```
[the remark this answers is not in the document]
JohannesMessner: I think it is fine to have it be `Any`, the different frameworks will have their own classes for this. Since the dtype of own framework should never be used by another framework I don't think we have to unify it.
I think in the subclasses (concrete comp backends) you can then specify the actual type

docarray/computation/abstract_comp_backend.py
```diff
@@ -74,6 +74,12 @@ def to_device(tensor: 'TTensor', device: str) -> 'TTensor':
         """Move the tensor to the specified device."""
         ...
 
+    @staticmethod
+    @abstractmethod
+    def device(tensor: 'TTensor') -> Optional[str]:
```
samsja, line 79: is device always a string ?

samsja: ohh I did not see that you call string on the device, looks good

samsja (approved):

PR #1058, opened 2023-01-26
feat: add validate function to documents to have less verbose api

samsja:
Signed-off-by: samsja <sami.jaghouar@hotmail.fr>
# Comtext

Follow up on this : https://github.com/orgs/docarray/projects/4?pane=issue&itemId=18966242

```python
    class MyDoc(BaseDocument):
        text1: Text
        text2: Text

    MyDoc(text1='hello', text2=Text(text='world'))
```

allow to pass directly a Text

### DIFF
--- a/docarray/documents/audio.py
+++ b/docarray/documents/audio.py
@@ -1,8 +1,18 @@
-from typing import Optional, TypeVar
+from typing import Any, Optional, Type, TypeVar, Union
+
+import numpy as np
 
 from docarray.base_document import BaseDocument
 from docarray.typing import AnyEmbedding, AudioUrl
+from docarray.typing.tensor.abstract_tensor import AbstractTensor
 from docarray.typing.tensor.audio.audio_tensor import AudioTensor
+
+try:
+    import torch
+
+    torch_available = True
+except ImportError:
+    torch_available = False
 
 T = TypeVar('T', bound='Audio')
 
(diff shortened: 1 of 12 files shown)


JohannesMessner (changes requested):
Like it, but I think we should do the same for all of our default docs: identify the main content field and allow that to be passed directly. I think for most of the other ones that would be a tensor instead of `str`

samsja:
yes I will add the resti n this pr as well 

for the other document what do we do 

if it is a string we pass to url ?
if it is a tensor we pass to tensor field ?

@anna-charlotte what do u think ?

JohannesMessner:
> yes I will add the resti n this pr as well
> 
> for the other document what do we do
> 
> if it is a string we pass to url ? if it is a tensor we pass to tensor field ?
> 
> @anna-charlotte what do u think ?

So instead of identifying one main content field per predefined doc we decide based on the input data's type? Could also make sense, but lead to confusion if multiple fields have the same type. E.g. some Document might have two fields of type `str`, then we need to choose one of those

samsja:
> So instead of identifying one main content field per predefined doc we decide based on the input data's type? Could also make sense, but lead to confusion if multiple fields have the same type. E.g. some Document might have two fields of type `str`, then we need to choose one of those

but we do this only for the one that make sense. But nothing automatic

JoanFM (approved):

JoanFM (approved):

samsja:
> For the `Audio` doc we could do the same, right? @samsja

I forgot this one

samsja:
partially fix : https://github.com/docarray/docarray/issues/1056

PR #1064, opened 2023-01-30
feat(v2): add tensorflow support

anna-charlotte:

Goals:

Add Tensorflow support to v2:

- [x] Add TF tensor type
  - As of right now, I haven't found a solution on how to subclass a `tf.Tensor` ([see here](https://github.com/tensorflow/tensorflow/issues/59472)), so for now we weill store the `tf.Tensor` instance as an attribute of a `TensorFlowTensor` instance.
- [x] Add TF computational backend


docarray/typing/tensor/tensorflow_tensor.py
```diff
@@ -0,0 +1,141 @@
...
+
+    __parametrized_meta__ = metaTensorFlow
+
+    def __init__(self, tensor: tf.Tensor):
+        super().__init__()
+        self._tensor = tensor
+
+    @property
+    def tensor(self):
+        return self._tensor
```
JohannesMessner, line 43: Since when do we have a `tensor` property and an `unwrap()` method that do the same thing?

docarray/typing/tensor/tensorflow_tensor.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@_register_proto(proto_type_name='tensorflow_tensor')
+class TensorFlowTensor(AbstractTensor, Generic[ShapeT], metaclass=metaTensorFlow):
+
+    __parametrized_meta__ = metaTensorFlow
+
+    def __init__(self, tensor: tf.Tensor):
+        super().__init__()
+        self._tensor = tensor
```
JohannesMessner, line 39: Should we override `__getattr__()` to pass through all calls to this `self._tensor()`?

docarray/typing/tensor/tensorflow_tensor.py
JohannesMessner: I would leave it public and just keep the `unwrap()` so that the interface is uniform between tf and the other ones (and the other ones don't have a `tensor` that they store).
`unwrap()` seems to be in the abstract tensor already, which is all we need i think :)

docarray/typing/tensor/tensorflow_tensor.py
JohannesMessner: what i mean is that right now our tensorflow tensor is not a subclass of the native tensorflow tensor. But we could fake the same feeling by passing through all operations to `.tensor`. Something like this:

```
def __getattr__(self, name):
    return getattr(self.tensor, name)
```

docarray/computation/tensorflow_backend.py
```diff
@@ -0,0 +1,280 @@
...
+
+def norm_left(t: tf.Tensor) -> TensorFlowTensor:
+    return TensorFlowTensor(tensor=t)
+
+
+def norm_right(t: TensorFlowTensor) -> tf.Tensor:
+    return t.tensor
+
+
+class TensorFlowCompBackend(AbstractNumpyBasedBackend[TensorFlowTensor]):
```
samsja, line 51: general comment about the class, to me it looks good I like that we rely on tnp as much as we can.

Maybe norm_left, norm_right should be named differenlty

norm_left could be : cast_output
norm_right could be: get_tensor

this will make the code clearer

maybe @JohannesMessner as a better idea for the name.

docarray/typing/tensor/tensorflow_tensor.py
```diff
@@ -0,0 +1,224 @@
...
+class metaTensorFlow(
+    AbstractTensor.__parametrized_meta__,  # type: ignore
+    node_base,  # type: ignore
+    tf_base,  # type: ignore
+):  # type: ignore
+    pass
+
+
+@_register_proto(proto_type_name='tensorflow_tensor')
+class TensorFlowTensor(AbstractTensor, Generic[ShapeT], metaclass=metaTensorFlow):
```
samsja, line 35: I think you need to explain that contratry to torch and ndarray we cannot use the TensorFLowTensor as if it was a tf.tensor because we cannot subclass and therefore user have to use unwrap() or .tensor directly

docarray/typing/tensor/tensorflow_tensor.py
JohannesMessner: ok then let's not do it, it's too dangerous

.github/workflows/ci.yml
```diff
@@ -181,12 +183,40 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install poetry
           poetry install --all-extras
-          pip install protobuf==3.19.0 # we check that we support 3.19 
+          poetry run pip install protobuf==3.19.0 # we check that we support 3.19
+
+      - name: Test
+        id: test
+        run: |
```
samsja, line 190: can be removed

docarray/array/array_stacked.py
```diff
@@ -141,7 +149,22 @@ def _create_columns(
         columns: Dict[str, Union[DocumentArrayStacked, AbstractTensor]] = dict()
 
         for field, type_ in column_schema.items():
-            if issubclass(type_, AbstractTensor):
+            if tf_available and isinstance(getattr(docs[0], field), TensorFlowTensor):
+                tf_stack = []
```
samsja, line 153: can you add a commennt on why we do this differently for TF ?

docarray/array/array_stacked.py
```diff
@@ -204,7 +227,12 @@ def __getitem__(self, item):  # note this should handle slices
         doc = self._docs[item]
         # NOTE: this could be speed up by using a cache
```
samsja, line 228: can u remove this comment ? I think it is not true anymore

docarray/array/array_stacked.py
```diff
@@ -204,7 +227,12 @@ def __getitem__(self, item):  # note this should handle slices
...
         for field in self._columns.keys():
-            setattr(doc, field, self._columns[field][item])
+            value = self._columns[field]
```
samsja, line 230: I am not even sure we need this piece of code anymore

tests/integrations/typing/test_typing_proto.py
```diff
@@ -45,3 +47,42 @@ class Mymmdoc(BaseDocument):
             assert isinstance(value, np.ndarray) or isinstance(value, torch.Tensor)
         else:
             assert isinstance(value, doc._get_field_type(field))
+
+
+@pytest.mark.tensorflow
```
samsja, line 52:
```suggestion
@pytest.mark.tensorflow
@pytest.mark.proto

```

samsja (changes requested):
Looks good ! One thing, we need to add in the readme how to install with tensorflow.

docarray/array/array_stacked.py
samsja: yes exactly. Can u remove to try ?

.github/workflows/ci.yml
samsja: well do we need this Test things in the CI ?

docarray/array/array_stacked.py
```diff
@@ -26,15 +26,23 @@
...
 
 try:
     from docarray.typing import TorchTensor
 except ImportError:
     TorchTensor = None  # type: ignore
 
+try:
+    import tensorflow as tf  # type: ignore
+
+    from docarray.typing import TensorFlowTensor
```
JohannesMessner, line 39: for torch i moved this thing to a helper in `utils`, so this check only has to be done once globally. Can we do the same for tf?

JohannesMessner (changes requested):

README.md
```diff
@@ -277,6 +277,94 @@ You instantly win in code readability and maintainability. And for the same pric
...
+        return da
+
+    def call(self, inputs: DocumentArray[PairPodcast]) -> DocumentArray[PairPodcast]:
+        inputs.left = self.forward_podcast(inputs.left)
+        inputs.right = self.forward_podcast(inputs.right)
+
+        return inputs
+```
+
+Much nicer, don't you think? One main difference to using DocArray with PyTorch, is that when using TensorFlowTensor's, you have to access it's `.tensor` attribute directly, as it can be seen in `.forward_podcast()` above. This is due to the fact that while `TorchTensor` is a subclass of `torch.Tensor`, `TensorFlowTensor` is not a subclass of `tf.Tensor` but instead stores a `tf.Tensor` in its `.tensor` attribute.
```
JoanFM, line 366: It may be good to explain why this is the case?

JoanFM: Why is this different?

samsja: > Why is this different?
@anna-charlotte you can probably add due to a technical limitation on tf.Tensor

samsja (approved):

JohannesMessner (approved):
LGTM

PR #1065, opened 2023-01-30
feat: add byte to documents

samsja:
# Context

we need to be able to send documents as bytes inside of tensor (as it is wayyy smaller in bytes).

## What this pr do

- [x] add a byte field to image
- [x] refactor image loading 
- [x] add a byte field to audio
- [x] add a byte field to video
- [x] add a byte field to mesh

What need to be done:

figure it out if we should not do base64 encoding under the hood or let it to the user:

EDIT: conclusion we let the user deal with base64


docarray/typing/tensor/image/abstract_image_tensor.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC, abstractmethod
+
+from docarray.typing.tensor.abstract_tensor import AbstractTensor
+
+
+class AbstractImageTensor(AbstractTensor, ABC):
+    @abstractmethod
+    def bytes(self):
```
JoanFM: why don't we add this `abstract` method directly to `AbstractTensor`? I do not see why this is related to `image` at all

JoanFM (dismissed):

docarray/typing/tensor/image/abstract_image_tensor.py
samsja: this one is specific to Images as it converts the tensor to a PNG image and then saves it. This is different from passing the tensor to bytes. 

I don't think that all of the Tensor should have a bytes method

JohannesMessner: > as it converts the tensor to a PNG image and then saves it
The method name does not really convey that imo

docarray/typing/tensor/image/abstract_image_tensor.py
```diff
@@ -0,0 +1,27 @@
...
+
+from docarray.typing.tensor.abstract_tensor import AbstractTensor
+
+
+class AbstractImageTensor(AbstractTensor, ABC):
+    @abstractmethod
+    def to_bytes(self, format: str = 'PNG') -> bytes:
+        """
+        Convert image tensor to bytes.
+        """
```
JohannesMessner, line 12: `format` and `return` are missing in the docstring

docarray/typing/tensor/image/image_ndarray.py
```diff
@@ -0,0 +1,44 @@
...
+from docarray.typing.tensor.ndarray import NdArray
+
+MAX_INT_16 = 2**15
+
+
+@_register_proto(proto_type_name='image_ndarray')
+class ImageNdArray(AbstractImageTensor, NdArray):
+    """
+    Subclass of NdArray, to represent an image tensor.
+    Adds image-specific features to the tensor.
```
JohannesMessner, line 12: what are such features?

docarray/typing/tensor/image/image_torch_tensor.py
```diff
@@ -0,0 +1,45 @@
...
+from docarray.typing.tensor.torch_tensor import TorchTensor, metaTorchAndNode
+
+T = TypeVar('T', bound='ImageTorchTensor')
+
+
+@_register_proto(proto_type_name='image_torch_tensor')
+class ImageTorchTensor(AbstractImageTensor, TorchTensor, metaclass=metaTorchAndNode):
+    """
+    Subclass of TorchTensor, to represent an image tensor.
+    Adds image-specific features to the tensor.
```
JohannesMessner, line 14: same comment about the docstring as above

docarray/typing/url/any_url.py
```diff
@@ -103,3 +107,21 @@ def from_protobuf(cls: Type[T], pb_msg: 'str') -> T:
         :return: url
         """
         return parse_obj_as(cls, pb_msg)
+
+    def load_bytes(self, timeout: Optional[float] = None) -> bytes:
+        """Convert url to bytes. This will either load or download the file and save
+        it into a bytes object.
+        :param uri: the uri of Document
```
JohannesMessner, line 114:
```suggestion
        :param uri: the URI of Document. Can be a local file path or a (remote) URL
```

docarray/typing/url/any_url.py
```diff
@@ -103,3 +107,21 @@ def from_protobuf(cls: Type[T], pb_msg: 'str') -> T:
...
+        :param timeout: timeout for urlopen. Only relevant if uri is not local
```
JohannesMessner, line 115:
```suggestion
        :param timeout: timeout for urlopen. Only relevant if URI is not local
```

JohannesMessner (changes requested):

docarray/typing/bytes/image_bytes.py
```diff
@@ -0,0 +1,111 @@
...
+    @classmethod
+    def validate(
+        cls: Type[T],
+        value: Any,
+        field: 'ModelField',
+        config: 'BaseConfig',
+    ) -> T:
+
+        value = bytes_validator(value)
+        return cls(value)
```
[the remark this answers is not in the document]
samsja: I think if we ommit this validation then the from protobuf read will fail as it does not define the validate

docarray/typing/tensor/audio/abstract_audio_tensor.py
```diff
@@ -1,19 +1,22 @@
...
 class AbstractAudioTensor(AbstractTensor, ABC):
-    @abstractmethod
-    def to_audio_bytes(self):
+    def to_bytes(self):
         """
         Convert audio tensor to bytes.
         """
-        ...
+        tensor = self.get_comp_backend().to_numpy(self)
+        tensor = (tensor * MAX_INT_16).astype('<h')
```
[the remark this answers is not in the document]
samsja: good point I need to add the test

docarray/typing/tensor/image/abstract_image_tensor.py
samsja: What name would you like ? Now it is `to_bytes`

docarray/typing/bytes/image_bytes.py
```diff
@@ -0,0 +1,111 @@
...
+        :param width: width of the image tensor.
+        :param height: height of the image tensor.
+        :param axis_layout: ordering of the different image axes.
+            'H' = height, 'W' = width, 'C' = color channel
+        :return: np.ndarray representing the image as RGB values
+        """
+
+        from PIL import Image as PILImage
+
+        raw_img = PILImage.open(BytesIO(self))
```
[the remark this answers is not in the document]
samsja: good point I need to add the test

docarray/typing/tensor/image/abstract_image_tensor.py
JohannesMessner: After checking the code i think the name is good, but it doesn't 'save' (to disk) anywhere, right?

JohannesMessner (approved):

docarray/typing/tensor/image/abstract_image_tensor.py
samsja: no no disk involved

docarray/typing/tensor/audio/abstract_audio_tensor.py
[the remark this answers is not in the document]
samsja: lets keep it like this for now I believe

docarray/documents/audio.py
```diff
@@ -86,6 +86,7 @@ class MultiModalDoc(Document):
     url: Optional[AudioUrl]
     tensor: Optional[AudioTensor]
     embedding: Optional[AnyEmbedding]
+    bytes: Optional[bytes]
```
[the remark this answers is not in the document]
samsja: ah good point, I actually forgot to change it

PR #1066, opened 2023-01-31
refactor: depend on jaxtyping

JohannesMessner:
The goal is to depend on jaxtyping for parametrized tensor types. This should cover more edge cases, give us additional features, and reduce our maintenance burden. See jaxtyping [here](https://github.com/google/jaxtyping/blob/main/API.md).
I'll update this description with more details once the PR is further along.

**TODO:**
- [ ] make tests pass
- [x] enable all the dtypes from jaxtyping
  - [x] add a default dtype
- [x] better syntax to specify dtype
- [ ] should classes like `TorchTensor` also be a jaxtyping type?
- [ ] change the name of generated classes
- [ ] handle validation
- [x] maybe generate dtype classes automatically?


docarray/typing/tensor/abstract_tensor.py
```diff
@@ -54,26 +116,23 @@ def _to_node_protobuf(self: T) -> 'NodeProto':
 
     @staticmethod
     def _parse_item(item: Any) -> Tuple[Optional[str], Optional[str]]:
+        def _is_jaxtyping_class(_cls: type) -> bool:
+            return inspect.isclass(_cls) and issubclass(_cls, JaxTypingDType)
```
JoanFM, line 120: why do u need the `isclass` inspection? isn't this already assumed from the `issubclass`?

JoanFM (changes requested):

docarray/typing/tensor/abstract_tensor.py
JohannesMessner: no, subclass will fail if the first input is not  a class, annoying

samsja:
@JohannesMessner should we close for now ?

PR #1074, opened 2023-02-02
feat: advanced indexing

### LINKED ISSUE #1073: toch/np like indexig for `DocumentArrayStacked`
PyTorch (and numpy, too) supports a rich and flexible indexing API: https://pytorch.org/cppdocs/notes/tensor_indexing.html
DocumentArrayStacked should support the same.

Concretely, we should be able to do the following:

```python
da = DocumentArray([... for _ in range(5)]
da[0]  # index by offset; returns doc at this position
da[0:5:2]  # index by slice; returns every other doc
da[0, 1, 2]  # index by sequence of positions; returns DocumentArray with docs at these positions
da[[0, 1, 2]]  # same as above
da[[True, False, True, True, False]]  # index by boolean mask; returns DocumentArray with "True" docs
da[None]  # returns `da`, just to conform with PyTorch convention
```

JohannesMessner:
Goals:

PyTorch like indexing into `DocumentArray` and `DocumentArrayStacked`

```python
da = DocumentArray([... for _ in range(5)]
da[0]  # index by offset; returns doc at this position
da[0:5:2]  # index by slice; returns every other doc
da[0, 1, 2]  # index by sequence of positions; returns DocumentArray with docs at these positions
da[[0, 1, 2]]  # same as above
da[[True, False, True, True, False]]  # index by boolean mask; returns DocumentArray with "True" docs
da[None]  # returns `da`, just to conform with PyTorch convention
```

This also includes a refactor where in `DocumentArrayStacked` `._columns` is now separated into `_doc_columns` and `_tensor_columns`, instead of having both types in the same dict. It makes everything a bit clearer.

closes #1073


docarray/array/array.py
```diff
@@ -99,6 +100,15 @@ def __len__(self):
     def __getitem__(self, item):
         if type(item) == slice:
             return self.__class__(self._data[item])
+        elif isinstance(item, Iterable):
+            offset_to_doc = self._get_offset_to_doc()
```
JoanFM, line 104: this will be super slow right?

JoanFM: We should it compute all the time or `cache` it

JoanFM (changes requested):

docarray/array/array.py
JohannesMessner: the `_get_offset_to_doc()` you mean? we have to iterate over all the data at least once anyways; by doing this we iterate only once (during this call), if we were to not do this we would iterate once for every index given in the `Iterable`, so this is faster than the naive approach.

But yes, it could be made even faster through some caching

docarray/array/array.py
```diff
@@ -87,24 +91,118 @@ class Image(BaseDocument):
...
+        if item is None or isinstance(item, (int, slice, tuple, list)):
+            return item
+
+        index_has_getitem = hasattr(item, '__getitem__')
+        is_valid_bulk_index = index_has_getitem and isinstance(item, Iterable)
+        if not is_valid_bulk_index:
+            raise ValueError(f'Invalid index type {type(item)}')
+
+        if isinstance(item, np.ndarray) and (
+            item.dtype == np.bool or np.issubdtype(item.dtype, np.integer)
```
samsja, line 155: what abour torch tensor as integer here ?

samsja: maybe we don't want to support these though

docarray/array/array.py
```diff
@@ -87,24 +91,118 @@ class Image(BaseDocument):
...
+        ):
+            return item.tolist()
+
+        try:
+            import torch  # noqa: F401
```
samsja, line 160: lets do this import at the tope of the file. Torch import can be slow.

we do this a couple of other time in the codebase it looks like this

```python
try:
    import torch
    torch_imported = True
except ImportError:
    torch_imported = False
```

docarray/array/array.py
```diff
@@ -87,24 +91,118 @@ class Image(BaseDocument):
...
+            torch.bool,
+            torch.int64,
+        ]
+        if isinstance(item, torch.Tensor) and (item.dtype in allowed_torch_dtypes):
+            return item.tolist()
+
+        return item
+
+    def _get_from_indices(self: T, item: Iterable[int]) -> T:
+        offset_to_doc = self._get_offset_to_doc()
```
samsja, line 173: I am not sure to understand why we need this list ?

Cannot we just une the da._data as a list ?

docarray/array/array.py
JohannesMessner: then I guess we should do this once globally and import it everywhere else

docarray/array/array.py
JohannesMessner: after reading into python lists, you are right. This was meant as an optimization, but i incorrectly assumed that indexing a python list is an O(n) operation, but actually it is O(1). So this is not needed

docarray/array/array.py
samsja: but where globaly ?

JoanFM (approved):

JoanFM:
Would u mind adding a description of what it enables?

PR #1076, opened 2023-02-02
feat: reduce and update methods for DocumentArray and BaseDocument

JoanFM:
Goals:

Implement `reduce` methods in `utils`


docarray/utils/reduce.py
```diff
@@ -0,0 +1,164 @@
...
+
+def _types_analysis(doc: 'BaseDocument') -> List[List[str]]:
+    simple_non_empty_fields: List[str] = []
+    list_fields: List[str] = []
+    set_fields: List[str] = []
+    nested_docs_fields: List[str] = []
+    nested_docarray_fields: List[str] = []
+
+    for field_name, field in doc.__fields__.items():
+        field_type = field.outer_type_
```
samsja, line 14:
```suggestion
        field_type = doc._get_field_type(field_name)
```

docarray/utils/reduce.py
```diff
@@ -0,0 +1,164 @@
...
+                nested_docs_fields.append(field_name)
+            else:
+                simple_non_empty_fields.append(field_name)
+    return [
+        simple_non_empty_fields,
+        list_fields,
+        set_fields,
+        nested_docarray_fields,
+        nested_docs_fields,
+    ]
```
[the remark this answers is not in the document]
JoanFM: good point

docarray/utils/reduce.py
JoanFM: will mypy not complain about accessing a private method?

samsja: nop I don't think it is an antipattern in python to access private method of other object when you are inside the same library. Python do it all the time itself with all of the `__dunder__` method and attribute

JohannesMessner: no as long as the types match up it will no complain

JoanFM: ok

docarray/utils/reduce.py
```diff
@@ -0,0 +1,164 @@
...
+
+def reduce_docs(doc1: 'BaseDocument', doc2: 'BaseDocument') -> 'BaseDocument':
+    """
+    Reduces doc1 and doc2 into one Document in-place. Changes are applied to doc1.
+    Reducing 2 Documents consists in setting data properties of the second Document
+    to the first Document if they are empty (priority to the left-most Document)
+    and reducing recursively its nested Documents and DocumentArrays
+    :param doc1: first Document to be reduced. Change is applied in-place
+    :param doc2: second Document to be reduced
+    :return The reduced Document
```
JohannesMessner, line 46:
```suggestion
    :return: The reduced Document
```

docarray/utils/reduce.py
```diff
@@ -0,0 +1,164 @@
...
+    .. note::
+        - Nested DocumentArrays order does not follow any specific rule.
+        You might want to re-sort them in a later step.
+        - The final result depends on the order of DocumentArrays
+        when applying reduction.
+
+    :param docarrays: List of DocumentArrays to be reduced
+    :return: the resulting DocumentArray
+    """
+    assert (
```
samsja, line 156: Better to raise an actual expection as assert can be disable at runtime

docarray/utils/reduce.py
```diff
@@ -0,0 +1,164 @@
...
+    return [
+        simple_non_empty_fields,
+        list_fields,
+        set_fields,
+        nested_docarray_fields,
+        nested_docs_fields,
+    ]
+
+
+def reduce_docs(doc1: 'BaseDocument', doc2: 'BaseDocument') -> 'BaseDocument':
```
JohannesMessner, line 38: Would it make sense to generalize this to take in an number or `*docs` and do things in a loop? Or is it overkill?

samsja, line 38: maybe we could add one very small example ? Reducing with two very basic document

JoanFM: Good

docarray/utils/reduce.py
```diff
@@ -0,0 +1,164 @@
...
+        setattr(doc1, field, reduce_docs(getattr(doc1, field), getattr(doc2, field)))
+
+    for field in set(doc1_list_fields + doc2_list_fields):
+        array1 = getattr(doc1, field)
+        array2 = getattr(doc2, field)
+        if array1 is None and array2 is not None:
+            setattr(doc1, field, array2)
+        elif array1 is not None and array2 is not None:
+            array1.extend(array2)
+            setattr(doc1, field, array1)
```
JohannesMessner, line 80: I think this behaviour (extending the list) is not really expected from the docstring; i would have expected it to be overridden treated like a simple field. It makes sense to do it this way, but it may need an additional sentence to explain

docarray/utils/reduce.py
```diff
@@ -0,0 +1,164 @@
...
+
+    for field in set(doc1_set_fields + doc2_set_fields):
+        array1 = getattr(doc1, field)
+        array2 = getattr(doc2, field)
+        if array1 is None and array2 is not None:
+            setattr(doc1, field, array2)
+        elif array1 is not None and array2 is not None:
+            array1.update(array2)
+            setattr(doc1, field, array1)
```
JohannesMessner, line 89: same as for the lists above

JohannesMessner (changes requested):

docarray/utils/reduce.py
```diff
@@ -0,0 +1,196 @@
...
+     if they are empty (priority to the left-most Document)
+     - Concatenating lists and updating sets
+     - Reducing recursively Documents and DocumentArrays
+     - Updating Dictionaries of the left with the right
+    :param doc1: first Document to be reduced. Change is applied in-place
+    :param doc2: second Document to be reduced
+    :return: The reduced Document
+    """
+    doc1_fields = _group_fields(doc1)
+    doc1_simple_non_empty_fields = doc1_fields.simple_non_empty_fields
```
samsja, line 71: do u need all of this item assignment ? I think you can just use `doc1_fields.simple_non_empty_fields` everywhere where instead of `doc1_simple_non_empty_fields`

docarray/base_document/document.py
```diff
@@ -46,3 +47,39 @@ def __str__(self):
...
+                doc1.update(doc2)
+                assert doc1.content == 'Core content updated'
+                assert doc1.title == 'Title'
+                assert doc1.tags_ == ['python', 'AI', 'docarray']
+
+        :param other: The Document with which to update the contents of this
+        """
+        from docarray.utils.reduce import reduce_docs
+
+        reduce_docs(self, other)
```
samsja, line 85: I would move the code from the function here tbh

docarray/utils/reduce.py
JoanFM: will do

docarray/base_document/document.py
JoanFM: ok

docarray/utils/reduce.py
JoanFM: I think is overkill, it has been moved to `update`

docarray/base_document/document.py
```diff
@@ -79,6 +84,128 @@ class MyDocument(BaseDocument):
...
+                if field_name not in FORBIDDEN_FIELDS_TO_UPDATE:
+                    field_type = doc._get_field_type(field_name)
+                    if not isinstance(field_type, _GenericAlias) and issubclass(
+                        field_type, DocumentArray
+                    ):
+                        nested_docarray_fields.append(field_name)
+                    elif (
+                        isinstance(field_type, _GenericAlias)
+                        and field_type.__origin__ is list
+                    ):
```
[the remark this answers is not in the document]
JoanFM: what Python version are u referring to?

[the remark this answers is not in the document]
samsja: in python 3.9 you can do

```python
class A:
   b: list[inŧ]
```

instead of 

```python
class A:
   b: List[int]
```

I am almost sure they have maintain compatibility with older version but the fact that we are using a private object from the typing library tell me that we are not 100% sure we will support this new way.

Maybe there is a way not to use the `_GenericAlias` since it is private ?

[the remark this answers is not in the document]
samsja: Oh actually we can use 

```python
from typing_inspect import is_generic_type
```

 we already rely on this library for some type hint things and they make sure that their function are compatible with every python version


https://github.com/ilevkivskyi/typing_inspect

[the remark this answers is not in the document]
JoanFM: can u propose a PR on top of mine to do this? I am not sure what u are intending to tell. I did trial and error until I found the working formula

docarray/base_document/document.py
```diff
@@ -46,3 +47,163 @@ def __str__(self):
...
+            ],
+        )
+
+        FORBIDDEN_FIELDS_TO_UPDATE = ['ID']
+
+        def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
+            simple_non_empty_fields: List[str] = []
+            list_fields: List[str] = []
+            set_fields: List[str] = []
+            dict_fields: List[str] = []
```
samsja, line 123: what about tuples btw ? They should be handle as well right ?

samsja (changes requested):

docarray/base_document/document.py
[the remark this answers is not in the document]
samsja: I pushed a commit

docarray/base_document/document.py
```diff
@@ -46,3 +47,165 @@ def __str__(self):
...
+                f'{type(self)} with Document of type '
+                f'{type(other)}'
+            )
+        from collections import namedtuple
+
+        from docarray import DocumentArray
+        from docarray.utils.reduce import reduce
+
+        # Declaring namedtuple()
+        _FieldGroups = namedtuple(
```
samsja, line 107: We should declare this class outside of the the function IMO. So that we can use it as a type hint for the return type

I personnaly prefer the dataclass way of declaring the named tuple as you can declare the type hint

```python
class _FieldGroupds(NamedTuple):
    list_fields: List[str]
```

JoanFM: this is a private method, so I do not see why it has to come outside

docarray/base_document/document.py
JoanFM: Adde docs

samsja (approved):

PR #1078, opened 2023-02-03
feat: support proto 3

samsja:
# Context

the goal of this pr is to support both proto 3 and proto 4. Atm we only support proto 3.

## details.

* First, we relax the proto requirement. `protobuf>= 3.19.0` instead of `protobuf>=4.21`

protoc 3 and 4 does not generate the same python code. Therefore if we use proto 3 to generate the `docarray_pb2.py` it will break when using proto 4 and vice versa.

The solution to this problem is to generate the `docarray_pb2.py` from the proto two times. One with protoc3 and one with protoc4. Then the trick is to check which version of protobuf is installed during import of the proto file. And we serve the correct version


docarray/proto/__init__.py
```diff
@@ -1,11 +1,23 @@
...
+    )
+else:
+    from docarray.proto.pb2.docarray_pb2 import (
+        DocumentArrayProto,
+        DocumentArrayStackedProto,
+        DocumentProto,
+        NdArrayProto,
+        NodeProto,
+        UnionArrayProto,
+    )
```
JohannesMessner, line 20: aren't these two exactly the same? or am I blind? 😅

JohannesMessner: Oh i see the difference now, nevermind

samsja: this PR should be named "spot the difference"

JohannesMessner (approved):

docarray/proto/__init__.py
JohannesMessner: nerd version of "where is waldo?"

PR #1090, opened 2023-02-07
feat: add v1 equivalent Document

samsja:
# Context

add a legacy document following the schema of DocArray v1

### DIFF
--- a/docarray/documents/legacy/legacy_document.py
+++ b/docarray/documents/legacy/legacy_document.py
@@ -36,7 +36,7 @@ class LegacyDocument(BaseDocument):
     tensor: Optional[AnyTensor]
     chunks: Optional[DocumentArray[LegacyDocument]]
     matches: Optional[DocumentArray[LegacyDocument]]
-    blob: Optional[Any]
+    blob: Optional[bytes]
     text: Optional[str]
     url: Optional[str]
     embedding: Optional[AnyEmbedding]
--- a/docarray/documents/legacy/legacy_document.py
+++ b/docarray/documents/legacy/legacy_document.py
@@ -7,6 +7,31 @@ from docarray.typing import AnyEmbedding, AnyTensor
 
 
 class Document(BaseDocument):
+    """
+    This Document is the LegacyDocument. It follows the same schema as in DocArray v1.
+    It can be useful to start migrating a codebase from v1 to v2.
+
+    Nevertheless, the API is not totally compatible with DocAray v1 `Document`.
+    Indeed, none of the method associated with `Document` are present. Only the schema
+    of the data is similar.
+    .. code-block:: python
+
+        from docarray import DocumentArray
+        from docarray.documents.legacy import Document
+        import numpy as np
+
+        doc = Document(text='hello')
+        doc.url = 'http://myimg.png'
+        doc.tensor = np.zeros((3, 224, 224))
+        doc.embedding = np.zeros((100, 1))
+
+        doc.tags['price'] = 10
+
+        doc.chunks = DocumentArray[Document]([Document() for _ in range(10)])
+
+        doc.chunks = DocumentArray[Document]([Document() for _ in range(10)])
+
+    """
 
     tensor: Optional[AnyTensor]
     chunks: Optional[DocumentArray[Document]]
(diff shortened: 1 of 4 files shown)


docarray/documents/legacy/legacy_document.py
```diff
@@ -0,0 +1,19 @@
...
+class Document(BaseDocument):
+
+    tensor: Optional[AnyTensor]
+    chunks: Optional[DocumentArray[Document]]
+    matches: Optional[DocumentArray[Document]]
+    blob: Optional[Any]
+    text: Optional[str]
+    url: Optional[str]
+    embedding: Optional[AnyEmbedding]
+    # tags: Dict[str, Any] = dict()
```
JoanFM, line 18: tags should be there

JoanFM: Is there a way we can make the `oneof` behavior?

JoanFM: should we implement `content`?

JoanFM: as a property setter that checks where to put the value?

JoanFM (changes requested):

docarray/documents/legacy/legacy_document.py
```diff
@@ -0,0 +1,44 @@
...
+        doc.chunks = DocumentArray[Document]([Document() for _ in range(10)])
+
+        doc.chunks = DocumentArray[Document]([Document() for _ in range(10)])
+
+    """
+
+    tensor: Optional[AnyTensor]
+    chunks: Optional[DocumentArray[Document]]
+    matches: Optional[DocumentArray[Document]]
+    blob: Optional[Any]
```
JoanFM, line 39: blob is Any?

JoanFM (changes requested):

docarray/documents/legacy/legacy_document.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
+
+from typing import Any, Dict, Optional
+
+from docarray import BaseDocument, DocumentArray
+from docarray.typing import AnyEmbedding, AnyTensor
+
+
+class Document(BaseDocument):
```
JohannesMessner, line 9: I think we have to discuss the name. `Document` gives it too much importance, `LegaceDocument` is maybe too much of a "we won't support this anymore" vibe. Maybe `ClassicDocument`?

JohannesMessner (dismissed):

docarray/documents/legacy/legacy_document.py
samsja: >  Is there a way we can make the oneof behavior?

Oneof between what and what ?

samsja: > should we implement content

I don't think it is nescerray tbh

docarray/documents/legacy/legacy_document.py
samsja: I think as of today we can put any in blob right ? Like I can put a pillow image in blob for instance.

Or maybe I misremenber and in this case we should put blob

docarray/documents/legacy/legacy_document.py
samsja: but it is Document only under the `docarray.documents.legacy` namespace. But yeah maybe if someone search in the codebase it will find this Document and think it is important.

I think `LegacyDocument` is not too much. Nothing promise we will support this forever. Lets in the future we have to make a choice on a API that will make this Document obsolete then I am not sure we want to say. "No we cannot do it bc we will break `ClassicDocument`"

docarray/documents/legacy/legacy_document.py
JoanFM: blob is `bytes` no?

docarray/documents/legacy/legacy_document.py
JoanFM: I think `LegacyDocument` is good enough

docarray/documents/legacy/legacy_document.py
samsja: done

docarray/documents/legacy/legacy_document.py
samsja: done

JoanFM (approved):

PR #1095, opened 2023-02-07
feat: delitem for DocumentArray

JoanFM:
Signed-off-by: Joan Fontanals Martinez <joan.martinez@jina.ai>

Goals:
Implement `__del_item__` for DocumentArray and DocumentArrayStacked

### DIFF
--- a/docarray/array/array.py
+++ b/docarray/array/array.py
@@ -112,9 +112,7 @@ class DocumentArray(AnyDocumentArray, Generic[T_doc]):
         # [NdArray([0.11299577, 0.47206767, 0.481723  , 0.34754724, 0.15016037,
         #          0.88861321, 0.88317666, 0.93845579, 0.60486676, ... ]), ...]
 
-
     You can index into a DocumentArray like a numpy array or torch tensor:
-
 
     .. code-block:: python
         da[0]  # index by position
@@ -122,6 +120,11 @@ class DocumentArray(AnyDocumentArray, Generic[T_doc]):
         da[[0, 2, 3]]  # index by list of indices
         da[True, False, True, True, ...]  # index by boolean mask
 
+    You can delete items from a DocumentArray like a Python List
+
+    .. code-block:: python
+        del da[0]  # remove first element from DocumentArray
+        del da[0:5]  # remove elements fro 0 to 5 from DocumentArray
 
     """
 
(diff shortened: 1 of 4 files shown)


samsja (changes requested):
regarding DocumentArrayStackedI think we should not even expose delitem at the stack level. The philosophy so far was that if users need to do smth slow: "append, ..." then they need to first unstack.

We maybe can do some optimization later to allow such operation on the DocumentArrayStacked but I would consider this for later

JoanFM:
> regarding DocumentArrayStackedI think we should not even expose delitem at the stack level. The philosophy so far was that if users need to do smth slow: "append, ..." then they need to first unstack.
> 
> We maybe can do some optimization later to allow such operation on the DocumentArrayStacked but I would consider this for later

@JohannesMessner what do you think? should I just remove it for `DocumentArrayStacked`?

docarray/array/array.py
```diff
@@ -381,3 +384,22 @@ def traverse_flat(
...
+
+    @overload
+    def __delitem__(self: T, key: IndexIterType) -> None:
+        ...
+
+    def __delitem__(self, key) -> None:
+        key = self._normalize_index_item(key)
+
+        if key is None:
+            return
```
samsja, line 400: what is happening when you do `return` alone ?

docarray/array/array.py
```diff
@@ -381,3 +384,22 @@ def traverse_flat(
...
+        if type(key) == slice:
+            del self._data[key]
+
+        if isinstance(key, int):
+            del self._data[key]
```
samsja, line 405: not sure but cannot we just do 
```python
del self._data[key]
```
I this the if statement do nothing

docarray/array/array.py
JoanFM: returns None

docarray/array/array.py
JoanFM: makes sense

docarray/array/array_stacked.py
```diff
@@ -49,10 +48,14 @@ class DocumentArrayStacked(AnyDocumentArray):
...
     you can be precise a Document schema by using the `DocumentArray[MyDocument]`
-    syntax where MyDocument is a Document class  (i.e. schema).
+    syntax where MyDocument is a Document class (i.e. schema).
     This creates a DocumentArray that can only contains Documents of
     the type 'MyDocument'.
 
+    .. note::
+        Deleting elements of a DocumentArrayStacked will rebuild the tensors from a
+        reference to the updated unstacked DocumentArray. This operation may be slow
+         as this class is not designed for so regular update.
```
JohannesMessner, line 58: This should be gone now, right?

JohannesMessner (dismissed):

docarray/array/array_stacked.py
JoanFM: sure

JoanFM: recheck

samsja (approved):

PR #1113, opened 2023-02-09
feat(v2): display mesh and pointcloud

anna-charlotte:
Goals:

Add display methods to `PointCloud 3D` and `Mesh3D` urls and the corresponding tensors. For the tensors, first add classes to store those in: `VerticesAndFaces` for mesh (to store vertices: AnyTensor, faces: AnyTensor), `PointAndColors` for point cloud (to store points: AnyTensor, colors: AnyTensor).

For `PointCloud3D` you can call `.display()` on its url or tensors:

```python
doc = PointCloud3D(url='tests/toydata/tetrahedron.obj')
doc.tensors = doc.url.load(samples=10000)
doc.tensors.display()
# or via url
doc.url.display()
```

image

For `Mesh3D` you can call `.display()` on its url or tensors:

```python
doc = Mesh3D(url='tests/toydata/tetrahedron.obj')
doc.tensors = doc.url.load()
doc.tensors.display()
# or via url
doc.url.display()
```

image


JohannesMessner:
> After talking to @samsja we think that it would be nice to not have the `.display()` method on Document level, but rather type level, so that you call `pc_doc.url.display()` or `pc_doc.tensor.display()` instead of `pc_doc.display(display_from='url')`. For Image, Audio, Video this should work, since we have `ImageTensor`, `ImageUrl` and so on. For `Mesh3D` however we don't have such a tensor type, and also the display depends on both vertices and face. Same goes for the pointcloud display, which depends on a vertices tensor and optional color tensor. We could introduce documents for those instead of type, something like:
> 
> ```python
> class VerticesAndFaces(BaseDocument):
>     vertices: AnyTensor
>     faces: AnyTensor
> 
>     def display():
>         """display mesh from tensors"""
> ```
> 
> @JohannesMessner do you have any thought on this?

fully agree with putting it at the type level.

I am not 100% convinced with having "magic" documents such as `VerticesAndFaces` that add `display`, but probably it's the only way, right?

docarray/typing/url/url_3d/mesh_url.py
```diff
@@ -43,16 +41,26 @@ class MyDoc(BaseDocument):
...
 
         mesh = self._load_trimesh_instance(force='mesh')
 
         vertices = parse_obj_as(NdArray, mesh.vertices.view(np.ndarray))
         faces = parse_obj_as(NdArray, mesh.faces.view(np.ndarray))
 
-        return Mesh3DLoadResult(vertices=vertices, faces=faces)
+        return VerticesAndFaces(vertices=vertices, faces=faces)
+
+    def display(self) -> None:
```
JohannesMessner, line 59: `url.display` feels a bit weird to me, because in this case it doesn't really display the url, it displays the thing the url points to.
And to do that, it has to load from that url under the hood.

So is it necessary to expose this? Why not `url.load().display()`?

docarray/documents/mesh/vertices_and_faces.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Document for handling 3D mesh tensor data.
+
+    A VerticesAndFaces Document can contain an AnyTensor containing the vertices
+    information (`VerticesAndFaces.vertices`), and an AnyTensor containing the faces
+    information (`VerticesAndFaces.faces`).
+    """
+
+    vertices: Optional[AnyTensor]
+    faces: Optional[AnyTensor]
```
JohannesMessner, line 19: Should these really be optional? Seems like they would be required

JohannesMessner (changes requested):

docarray/typing/url/url_3d/mesh_url.py
samsja: but this is true that it is weird to display some data (the color) that we cannot load in our tools

docarray/typing/url/url_3d/point_cloud_url.py
```diff
@@ -60,4 +70,38 @@ class MyDoc(BaseDocument):
...
+    def display(self, samples: int = 10000) -> None:
+        """
+        Plot point cloud from url.
+        :param samples: number of points to sample from the mesh.
+        """
+        tensors = self.load(samples=samples)
+        _display_point_cloud(points=tensors.points)
+
+
+def _display_point_cloud(
```
samsja, line 85: lets define `_display_point_cloud` in the PointsAndColor document as a classmethod then you cast tensors.points into a PointsAndColor document and call dispaly

docarray/typing/url/url_3d/mesh_url.py
samsja: let me think about it

docarray/typing/url/url_3d/mesh_url.py
```diff
@@ -43,16 +41,28 @@ class MyDoc(BaseDocument):
...
+        """
+        Plot mesh in notebook from url.
+        This loads the Trimesh instance of the 3D mesh, and then displays it.
+        """
+        from IPython.display import display
```
samsja, line 65: what ahppened if we are not inside IPython ?

poetry.lock
```diff
@@ -1108,11 +1108,11 @@ wcwidth = "*"
 
 [[package]]
 name = "protobuf"
-version = "4.21.9"
-description = ""
+version = "3.19.6"
```
samsja, line 1111: why this one ?

docarray/typing/url/url_3d/mesh_url.py
JohannesMessner: so what's the conclusion here? keep it as it is? I see the point of being able to show colors, so I don't have a strong opinion anymore

samsja (approved):

docarray/typing/url/url_3d/mesh_url.py
samsja: @anna-charlotte what did you decided ?

JohannesMessner (approved):

PR #1120, opened 2023-02-09
feat: smart automatic casting from schema to another

samsja:
# Context

We introduce a simple mechanism to resolve casting between different document schemas. We believe that this casting is important to preserve interoperability in web app based on DocArray (like the one created with the jina framework on with fastAPI)

# how it works

To do this smart cast we follow the simple following algorithm:

Goal: to instantiate an object from this class we want to select from the
input dictionary the field that are valid candidate to respect the Document
schema.

To do so we do:

1 - We start by selecting all the fields from `obj` that match the Document
schema field in name and in type. (The value from obj[field_name] should be
an instance of the field type.

if we collected all the field we are over otherwise:

2 - For each of the remaining field of the schema we iterate over the
remaining value in obj and select the value that match the field type.
This is deterministic since Dict are order by default.

lets take an example

```python
from docarray.typing import AnyUrl, NdArray


class A(BaseDocument):
    url_0: AnyUrl
    url: AnyUrl

    tensor: NdArray
    tensor_2: NdArray


class B(BaseDocument):
    url: AnyUrl
    array: NdArray


a = A(url='file.png', tensor=np.zeros(3))

b = B.smart_parse_obj(a.dict())
print(b.link == a.url)  # True
print((b.array == a.tensor).all())  # True
```

in this example the `url` field of both schema matches in type, so we select
the field `url` over `url_0`. Then we need to find a NdArray object for the
`array`field. Since they are no `array` field in A picked the first field
that is NdArray, in this cas `tensor`

You can as well use the `cast_map`argument to explicit precise the mapping for
certain field. For instance with the Document defines above:

```python
a = A(url='file.png', url_0='hello', tensor=np.zeros(3), tensor_2=np.zeros(4))
b = B.smart_parse_obj(a.dict(), {'url': 'url_0'})
print(b.url == a.url_0)  # True
print((b.array == a.tensor).all())  # True
```

### DIFF
--- a/docarray/base_document/document.py
+++ b/docarray/base_document/document.py
@@ -1,5 +1,7 @@
+from __future__ import annotations
+
 import os
-from typing import List, Type
+from typing import Any, Dict, List, Type, TypeVar
 
 import orjson
 from pydantic import BaseModel, Field, parse_obj_as
@@ -13,6 +15,8 @@ from docarray.base_document.mixins import PlotMixin, ProtoMixin
 from docarray.typing import ID
 
 _console: Console = Console()
+
+T = TypeVar('T', bound='BaseDocument')
 
 
 class BaseDocument(BaseModel, PlotMixin, ProtoMixin, AbstractDocument, BaseNode):
@@ -213,3 +217,43 @@ class BaseDocument(BaseModel, PlotMixin, ProtoMixin, AbstractDocument, BaseNode)
             elif dict1 is not None and dict2 is not None:
                 dict1.update(dict2)
                 setattr(self, field, dict1)
+
+    @classmethod
+    def smart_parse_obj(cls: Type[T], obj: Dict[str, Any]) -> T:
+        """
+        parse the given object, validate the data and return a new initialized objet
+        the difference from `parse_obj` is that it will smartly resolve conflict and
+        undefined key.
+
+        Here we should describe the algo and give an example:
+        :param obj, a dict of field and type
+        :return: a Document initialized from the dict data
+        """
+        fields: Dict[str, Any] = {}
+        for field_name in cls.__fields__.keys() & obj.keys():
+            if isinstance(obj[field_name], cls._get_field_type(field_name)):
+                fields[field_name] = obj[field_name]
+
+        # fields contains for now the field that match in name and in type
+        remaining_field_name = [
+            field_name
+            for field_name in cls.__fields__.keys()
+            if field_name not in fields.keys()
+        ]
+
+        remaining_field_obj = {
+            field_name: field
+            for field_name, field in obj.items()
+            if field_name not in fields.keys()
+        }
+
+        for _, val in remaining_field_obj.items():
+            for field_name in remaining_field_name:
+                if isinstance(val, cls._get_field_type(field_name)):
+                    fields[field_name] = val
+                    break
+
+            if len(fields) == len(cls.__fields__):
+                return cls(**fields)
+
+        raise ValueError('Cannot smartly cast the obj')
--- a/docarray/base_document/document.py
+++ b/docarray/base_document/document.py
@@ -308,7 +308,10 @@ class BaseDocument(BaseModel, PlotMixin, ProtoMixin, AbstractDocument, BaseNode)
         ]
 
         for field_name in remaining_field_name & obj.keys():
-            if isinstance(obj[field_name], cls._get_field_type(field_name)):
+            field_type = cls._get_field_type(field_name)  # TODO handle Union
+            if issubclass(field_type, type) and isinstance(
+                obj[field_name], cls._get_field_type(field_name)
+            ):
                 fields[field_name] = obj[field_name]
 
         if len(fields) == len(cls.__fields__):
--- a/docarray/base_document/document.py
+++ b/docarray/base_document/document.py
@@ -223,30 +223,30 @@ class BaseDocument(BaseModel, PlotMixin, ProtoMixin, AbstractDocument, BaseNode)
         cast_map: Optional[Mapping[str, str]] = None,
     ) -> T:
         """
-        parse the given object, validate the data and return a new initialized objet
-        the difference from `parse_obj` is that it will smartly resolve conflict and
-        undefined key.
+        Parse the given object, validate the data and return a new initialized object of type `cls`.
+        The difference from `parse_obj` is that it will smartly resolve conflicts and
+        undefined keys.
 
         The `cast_map` argument is for you to define an explicit casting from the schema
         of this Document to your object input
 
         To do this smart cast we follow the simple following algorithm:
 
-        Goal: to instantiate an object from this class we want to select from the
-        input dictionary the field that are valid candidate to respect the Document
+        To instantiate an object from this class, `smart_parse_obj()` selects fields from the
+        input dictionary that are valid candidates to respect the Document
         schema.
 
         To do so we do:
 
         1 - We start by selecting all the fields from `obj` that match the Document
         schema field in name and in type. (The value from obj[field_name] should be
-        an instance of the field type.
+        an instance of the field type)
 
-        if we collected all the field we are over otherwise:
+        if we collected all the fields we are over, otherwise:
 
         2 - For each of the remaining field of the schema we iterate over the
         remaining value in obj and select the value that match the field type.
-        This is deterministic since Dict are order by default.
+        This is deterministic since Dict are ordered by default.
 
         lets take an example
 
--- a/docarray/base_document/document.py
+++ b/docarray/base_document/document.py
@@ -227,9 +227,6 @@ class BaseDocument(BaseModel, PlotMixin, ProtoMixin, AbstractDocument, BaseNode)
         The difference from `parse_obj` is that it will smartly resolve conflicts and
         undefined keys.
 
-        The `cast_map` argument is for you to define an explicit casting from the schema
-        of this Document to your object input
-
         To do this smart cast we follow the simple following algorithm:
 
         To instantiate an object from this class, `smart_parse_obj()` selects fields from the
@@ -279,7 +276,6 @@ class BaseDocument(BaseModel, PlotMixin, ProtoMixin, AbstractDocument, BaseNode)
         `array`field. Since they are no `array` field in A picked the first field
         that is NdArray, in this cas `tensor`
 
-
         You can as well use the `cast_map`argument to explicit precise the mapping for
         certain field. For instance with the Document defines above:
 
@@ -290,7 +286,8 @@ class BaseDocument(BaseModel, PlotMixin, ProtoMixin, AbstractDocument, BaseNode)
             print((b.array == a.tensor).all())  # True
 
         :param obj: a dict of field and type that will be parsed into the Document
-        :param cast_map: optional explicit mapping.
+        :param cast_map: define an explicit casting from the schema
+        of this Document to your object input.
         :return: a Document initialized from the dict data
         """
         fields: Dict[str, Any] = {}
--- a/docarray/base_document/document.py
+++ b/docarray/base_document/document.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import os
-from typing import Any, Dict, List, Type, TypeVar
+from typing import Any, Dict, List, Mapping, Optional, Type, TypeVar
 
 import orjson
 from pydantic import BaseModel, Field, parse_obj_as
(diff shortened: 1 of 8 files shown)


JoanFM:
Please, properly separate the branches from other work

samsja:
FYI : this is the bug : https://github.com/docarray/docarray/blob/1e791c06572b809e47f10164ebba8694e11d8252/docarray/array/array_stacked.py#L399

here we need to drop the tensor. otherwise, we save it twices, we can fix it but we should rather drop the optimization for now

samsja:
> Please, properly separate the branches from other work

done

docarray/base_document/document.py
```diff
@@ -213,3 +217,43 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+    def smart_parse_obj(cls: Type[T], obj: Dict[str, Any]) -> T:
+        """
+        parse the given object, validate the data and return a new initialized objet
+        the difference from `parse_obj` is that it will smartly resolve conflict and
+        undefined key.
+
+        Here we should describe the algo and give an example:
+        :param obj, a dict of field and type
+        :return: a Document initialized from the dict data
+        """
```
JohannesMessner, line 231: so obj is being parsed to the schema of cls?

docarray/base_document/document.py
```diff
@@ -213,3 +217,43 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        for _, val in remaining_field_obj.items():
+            for field_name in remaining_field_name:
+                if isinstance(val, cls._get_field_type(field_name)):
+                    fields[field_name] = val
+                    break
+
+            if len(fields) == len(cls.__fields__):
+                return cls(**fields)
+
+        raise ValueError('Cannot smartly cast the obj')
```
JohannesMessner, line 259: Am I understanding correctly that this is the order of resolution?
1. take all fields where key and type match. collect those
2. take the remaining fields. look at these fields in order, and look only at the types. as soon as a field in `obj` is an instance of a type in `cls`, greedily assign those to each other. do do that for all fields in `obj`
3. If 1 and 2 don't succesfully resolve all fields, it fails

docarray/base_document/document.py
```diff
@@ -213,3 +217,43 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        parse the given object, validate the data and return a new initialized objet
+        the difference from `parse_obj` is that it will smartly resolve conflict and
+        undefined key.
+
+        Here we should describe the algo and give an example:
+        :param obj, a dict of field and type
+        :return: a Document initialized from the dict data
+        """
+        fields: Dict[str, Any] = {}
+        for field_name in cls.__fields__.keys() & obj.keys():
```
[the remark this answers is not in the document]
samsja: this is doing intersection

docarray/base_document/document.py
samsja: yes

docarray/base_document/document.py
samsja: exactly

JohannesMessner:
so there is no distinction between exact type match and inheritance match? Not saying that is bad, just for clarification

samsja:
> so there is no distinction between exact type match and inheritance match? Not saying that is bad, just for clarification
yes exactly. I think it makes sense. This is the whole point of polymorphism.

Should I add it to the docstring or is it obvious ?

docarray/base_document/document.py
```diff
@@ -213,3 +215,128 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+    ) -> T:
+        """
+        parse the given object, validate the data and return a new initialized objet
+        the difference from `parse_obj` is that it will smartly resolve conflict and
+        undefined key.
+
+        The `cast_map` argument is for you to define an explicit casting from the schema
+        of this Document to your object input
+
+        To do this smart cast we follow the simple following algorithm:
```
JohannesMessner, line 233:
```suggestion
        Parse the given object, validate the data and return a new initialized object of type `cls`.
        The difference from `parse_obj` is that it will smartly resolve conflicts and
        undefined keys.

        The `cast_map` argument is for you to define an explicit casting from the schema
        of this Document to your object input

        To do this smart cast we follow the simple following algorithm:
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,128 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        obj: Mapping[str, Any],
+        cast_map: Optional[Mapping[str, str]] = None,
+    ) -> T:
+        """
+        parse the given object, validate the data and return a new initialized objet
+        the difference from `parse_obj` is that it will smartly resolve conflict and
+        undefined key.
+
+        The `cast_map` argument is for you to define an explicit casting from the schema
+        of this Document to your object input
```
JohannesMessner, line 231: Maybe this should be further down? At this point of the docstring I am not sure if I understand this.

docarray/base_document/document.py
```diff
@@ -213,3 +215,128 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+
+    @classmethod
+    def smart_parse_obj(
+        cls: Type[T],
+        obj: Mapping[str, Any],
+        cast_map: Optional[Mapping[str, str]] = None,
+    ) -> T:
+        """
+        parse the given object, validate the data and return a new initialized objet
+        the difference from `parse_obj` is that it will smartly resolve conflict and
```
JohannesMessner, line 227: it's better to link to `parse_obj`

docarray/base_document/document.py
```diff
@@ -213,3 +215,128 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        undefined key.
+
+        The `cast_map` argument is for you to define an explicit casting from the schema
+        of this Document to your object input
+
+        To do this smart cast we follow the simple following algorithm:
+
+        Goal: to instantiate an object from this class we want to select from the
+        input dictionary the field that are valid candidate to respect the Document
+        schema.
```
JohannesMessner, line 237:
```suggestion
        To instantiate an object from this class, `smart_parse_obj()` selects fields from the
        input dictionary that are valid candidates to respect the Document
        schema.
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,128 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+
+        To do so we do:
```
JohannesMessner, line 239: Could the following be rewritten avoiding `we`?

docarray/base_document/document.py
```diff
@@ -213,3 +215,128 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+
+        1 - We start by selecting all the fields from `obj` that match the Document
+        schema field in name and in type. (The value from obj[field_name] should be
+        an instance of the field type.
```
JohannesMessner, line 243:
```suggestion
        an instance of the field type)
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,128 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+
+        if we collected all the field we are over otherwise:
```
JohannesMessner, line 245:
```suggestion
        if we collected all the fields we are over, otherwise:
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,128 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+
+        2 - For each of the remaining field of the schema we iterate over the
+        remaining value in obj and select the value that match the field type.
+        This is deterministic since Dict are order by default.
```
JohannesMessner, line 249:
```suggestion
        This is deterministic since Dict are ordered by default.
```

JohannesMessner (changes requested):

docarray/base_document/document.py
samsja: done

docarray/base_document/document.py
samsja: done

docarray/base_document/document.py
samsja: done

docarray/base_document/document.py
```diff
@@ -213,3 +215,126 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        obj: Mapping[str, Any],
+        cast_map: Optional[Mapping[str, str]] = None,
+    ) -> T:
+        """
+        Parse the given object, validate the data and return a new initialized object of
+         type `cls`. The difference from
+         :func:`~docarray.document.BaseDocument.parse_obj` is that it will smartly
+         resolve conflicts and undefined keys.
+
+        To do this smart cast we follow the simple following algorithm:
```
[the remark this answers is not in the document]
JohannesMessner: and needs to be moved above the actual algorithm

docarray/base_document/document.py
```diff
@@ -213,3 +215,126 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        schema field in name and in type. (The value from obj[field_name] should be
+        an instance of the field type)
+
+        if all the fields are collected  we are over, otherwise:
+
+        2 - For each of the remaining field of the schema, this function iterate over
+        the remaining value in obj and select the value that match the field type.
+        This is deterministic since Dict are ordered by default.
+
+        lets take an example
```
JohannesMessner, line 249:
```suggestion
        Let's take an example:
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,126 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+                array: NdArray
+
+
+            a = A(url='file.png', tensor=np.zeros(3))
+
+            b = B.smart_parse_obj(a.dict())
+            print(b.url == a.url)  # True
+            print((b.array == a.tensor).all())  # True
+
+        in this example the `url` field of both schema matches in type, therefore
```
JohannesMessner, line 275:
```suggestion
        In this example the `url` field of both schema matches in type, therefore
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,126 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        the field `url` is selected over `url_0`. So far there is still a missing
+        NdArray object for the `array`field. Since they are no `array` field in A picked
```
JohannesMessner, line 277:
```suggestion
        NdArray object for the `array` field. Since they is no `array` field in A picked
```

JohannesMessner, line 277:
```suggestion
        NdArray object for the `array`field. Since they are no `array` field in A, the method picks
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,126 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        the first field that is NdArray, in this cas `tensor`
```
JohannesMessner, line 278:
```suggestion
        the first field that is NdArray, in this case `tensor`
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,126 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+
+        You can as well use the `cast_map`argument to explicit precise the mapping for
```
JohannesMessner, line 280:
```suggestion
        You can also use the `cast_map` argument to explicitly precise the mapping for
```

docarray/base_document/document.py
```diff
@@ -213,3 +215,126 @@ def _group_fields(doc: 'BaseDocument') -> _FieldGroups:
...
+        certain field. For instance with the Document defines above:
```
JohannesMessner, line 281:
```suggestion
        certain field. For instance, with the Documents defines above:
```

JohannesMessner (changes requested):
some more suggestions, but I think it is clear what is happening, and I like the approach of making this a classmethod on `BaseDocument`.

I would be interested though, @alexcg1 as someone who has not been part of our discussions surrounding this, is it clear to you as well?

samsja:
> > some more suggestions, but I think it is clear what is happening, and I like the approach of making this a classmethod on `BaseDocument`.
> > I would be interested though, @alexcg1 as someone who has not been part of our discussions surrounding this, is it clear to you as well?
> 
> ```python
> class A(BaseDocument):
>     tensor: AnyTensor
> 
> class B(BaseDocument):
>     array: TorchTensor
> 
> a = A(tensor=np.zeros(3))
> b = B.smart_parse_obj(a.dict())
> ```
> 
> What does the `smart_parse_obj()` method do?

Well that is the question that we are asking you. Can you read the docstring and tell us if it understandable?

docarray/base_document/document.py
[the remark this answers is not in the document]
alexcg1: Can we change to "_uses_ the following algo"? Lots of "follow" in that sentence

samsja:
We closed this PR as we cannot ensure consistency as protobuf does not nesscesraly  dict  order : https://github.com/protocolbuffers/protobuf/issues/3848

we will go for an explicit mapping for now

samsja:
We closed this PR as we cannot ensure consistency as protobuf does not nesscesraly  dict  order : https://github.com/protocolbuffers/protobuf/issues/3848

we will go for an explicit mapping for now

PR #1124, opened 2023-02-10
feat: hnswlib document index

JohannesMessner:
Goals:

First implementation of a Document Store.

**Design doc**: https://lightning-scent-57a.notion.site/Document-Stores-v2-design-doc-f11d6fe6ecee43f49ef88e0f1bf80b7f

**Usage Example**:

```python
# create a Document Store
from docarray.storage.backends.HnswDocStore import HnswDocumentIndex
from pydantic import Field
from docarray import DocumentArray, BaseDocument

class MyDocument(BaseDocument):
    tens: NdArray[10]

store = HnswDocumentIndex[MyDocument](work_dir='path/to/work/dir')

# index some data    
data = DocumentArray[MyDocument]([MyDocument(tens=np.random.randn(10)) for i in range(10)])
store.index(data)

# find by np array
result = store.find(np.random.random((10,)), embedding_field='tens', limit=5)
result = store.find_batched(np.random.random((2, 10)), embedding_field='tens', limit=5)

# find by Document
result = store.find(MyDocument(tens=np.random.randn(10)), search_field='tens', limit=5)
result = store.find_batched(DocumentArray[MyDocument]([MyDocument(tens=np.random.randn(10)) for i in range(10)]), search_field='tens', limit=5)

# delete data
del store[data.id]  # delete all

# advanced configs
class MyDocument(BaseDocument):
    tens_one: NdArray = Field(dim=10, space='cosine')
    tens_two: NdArray = Field(dim=10, space='l2')
store = HnswDocumentStore[MyDocument](work_dir='path/to/work/dir')

# select search field
data = DocumentArray[MyDocument]([MyDocument(tens_one=np.random.randn(10), tens_two=np.random.randn(10)) for i in range(10)])
store.index(data)
result = store.find(np.random.random((10,)), search_field='tens_one', limit=5)
result = store.find(np.random.random((10,)), search_field='tens_two', limit=5)

# use query builder
q = store.build_query()
q.find(query=data[0], search_field='tens_one', limit=5)
q.find(query=data[1], search_field='tens_two', limit=5)
q.filter(filter_query={'tens_two': {'$exists': True}})
result = store.execute_query(q.build())

# nested search
class InnerDocument(BaseDocument):
    tens_one: NdArray = Field(dim=10)

class MyDocument(BaseDocument):
    d: InnerDocument
    tens: NdArray = Field(dim=10)

store = HnswDocumentStore[MyDocument](work_dir='path/to/work/dir')
data = DocumentArray[MyDocument]([MyDocument(d=InnerDocument(tens_one=np.random.randn(10)), tens=np.random.randn(10)) for i in range(10)])
store.index(data)

result = store.find(np.random.random((10,)), search_field='tens', limit=5)
result = store.find(np.random.random((10,)), search_field='d__tens_one', limit=5)

```

**ToDo:**
- [x] basic search
- [x] automatic conversions in the base class
- [x] query builder interface
  - [x] autocomplete for method kwargs
- [x] think about other interface methods such as `num_docs()` etc
- [x] think about abstraction layers
- [x] infer tensor dimensionality from type hint
- [x] handle config passing properly
- [x] persistence using SQLite
  - [ ] proper error messages when things go wrong
- [x] handle HNSWLib nuances around saving/loading and changing configurations
- [x] deleting documents
~~- [ ] nested access syntax: change `__` to `.`~~
- [ ] test torch and tf
- [ ] enable arbitrary data types in fields
- [ ] subindices
- [ ] tests

**Note:** unchecked boxes above have been moved to separate PR


JoanFM (commented):
Let's call `DocumentIndexers` and not `stores`

docarray/storage/backends/HnswDocStore.py
```diff
@@ -0,0 +1,352 @@
...
+    class QueryBuilder(BaseDocumentStore.QueryBuilder):
+        def build(self, *args, **kwargs) -> Any:
+            return self._queries
+
+    @dataclass
+    class DBConfig(BaseDocumentStore.DBConfig):
+        work_dir: str = '.'
+
+    @dataclass
+    class RuntimeConfig(BaseDocumentStore.RuntimeConfig):
```
JoanFM, line 83: it is not clear to me how and where does people provide this.

JoanFM (commented):
It is not clear to me how composability of the QueryBuilder work

docarray/storage/backends/HnswDocStore.py
JohannesMessner: what do you mean by "people"?
Backend devs put the dataclass as inner class in their implementation, and users can pass it to the `configure` method

JohannesMessner:
> It is not clear to me how composability of the QueryBuilder work

I think we need to leave that up to the individual backends. Since every backend defines their queries differently, I don't think we can compose queries for them. So the base query builder just collects all the calls and args.

Actually I would love to automate this more across different backends, but couldn't think of an approach. Do you have an idea?

docarray/storage/abstract_doc_store.py
```diff
@@ -81,25 +97,21 @@ class BaseDocumentStore(ABC, Generic[TSchema]):
...
-    def __init__(self, config: Optional[IsDataclass] = None):
+    def __init__(self, db_config=None, **kwargs):
         if self._schema is None:
             raise ValueError(
                 'A DocumentStore must be typed with a Document type.'
                 'To do so, use the syntax: DocumentStore[DocumentType]'
             )
-        self._config = config
-
+        self._db_config = db_config if db_config else self._db_config_cls(**kwargs)
```
JoanFM:
```suggestion
        self._db_config = db_config or self._db_config_cls(**kwargs)
```

JoanFM (changes requested):

JohannesMessner:
I would now call this PR *ready to review & merge*.
**Note**: *Only* the part considering the `AbstractDocumentStore` is considered ~done, there is still more work to do on the HNSWLib implementation. But this will be tackled in a separate PR.
That way we can unblock work on other backends/DBs.

So when reviewing please focus on:
- `AbstractDocumentStore` implementation
- how the HNSWLib implementation interacts with the abstraction layers

docarray/storage/abstract_doc_store.py
```diff
@@ -0,0 +1,646 @@
...
+        def _col_gen(col_name: str):
+            return (self._get_values_by_column([doc], col_name)[0] for doc in docs_seq)
+
+        return {col_name: _col_gen(col_name) for col_name in self._columns}
+
+    ##################################################
+    # Behind-the-scenes magic                        #
+    # Subclasses should not need to implement these  #
+    ##################################################
+    def __class_getitem__(cls, item: Type[TSchema]):
```
samsja, line 547: here you need to do the same magic as in DOcumentArray to handle the case where the item is a string or a typevar

samsja: maybe at some point we might want to abstract aways this templating behavior. But not now I guess

JoanFM (changes requested):
can we please change the folder names to something related to indices?

JoanFM (changes requested):
refactor storage to index in name of folders, files, packages, tests, etc ...

docarray/storage/abstract_doc_store.py
JohannesMessner: you're right

JoanFM (changes requested):
the `folder` storage should be renamed

JoanFM (approved):

docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -0,0 +1,400 @@
...
+            self._get_docs_sqlite_hashed_id(
+                ids_per_query.tolist(),
+            )
+            for ids_per_query in labels
+        ]
+        return FindResultBatched(documents=result_das, scores=distances)
+
+    @composable
+    def _find(
+        self, query: np.ndarray, search_field: str, limit: int, **kwargs
```
samsja, line 215: as I said before, I think this one should accept `kwargs` but actually be specific

docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -0,0 +1,400 @@
...
+
+HNSWLIB_PY_VEC_TYPES = [list, tuple, np.ndarray]
+if torch_imported:
+    import torch
+
+    HNSWLIB_PY_VEC_TYPES.append(torch.Tensor)
+
+
+class HnswDocumentIndex(BaseDocumentIndex, Generic[TSchema]):
+    def __init__(self, db_config=None, **kwargs):
```
samsja, line 55: why `**kwargs` here ? could we do smth so that the docstring are updatrd accordingly ? (could do in another PR but would need a ticket)

docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -0,0 +1,400 @@
...
+        if python_type == docarray.typing.ID:
+            return None  # TODO(johannes): handle this
+
+        raise ValueError(f'Unsupported column type for {type(self)}: {python_type}')
+
+    def _index(self, column_data_dic, **kwargs):
+        # not needed, we implement `index` directly
+        ...
+
+    def index(self, docs: Union[BaseDocument, Sequence[BaseDocument]], **kwargs):
```
samsja, line 141: I am not sure to understand why we need `**kwargs` here. Is it only for compatibility with the mother class ? Is this only a mypy problem ? I would rather silence mypy than offuscated value under a `**kwargs`

samsja (changes requested):
LGTM. I just worry about the utilization of `**kwargs` I think we should never allow them

docarray/doc_index/backends/hnswlib_doc_index.py
JoanFM: I believe is not a bad strategy to make API compatible with other Indexes

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: After thinking about it, I don't think it works. If this doesn't take `**kwargs` but the public `find` does (as it needs to), then how does `find` pass the **kwargs to `_find`?
At runtime the **kwargs will be passed to a specific `_find` implementation that accepts all the needed kwargs, but as far as mypy is concerned, they are passed to the abstract `_find`.

So if that doesn't take `**kwargs` it will not work.

(Nor should it, from a typing perspective, bc we should not rely on the fact that a subclass implementation does something that the base class implementation doesn't do)

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: It is convenience for the user.
Instead of creating a `config` object they can pass all the config params individually, as **kwargs, and it will just work.

Otherwise the user would have to import the config class (differently for each backend), instantiate it with some settings, and pass it, which is cumbersome.

But it can also be useful in certain settings, e.g. if a user want to clearly define all config setting globally. So both options exist.

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: same logic as with `_find`. No value if obfuscated in the **kwargs, the next line checks that **kwargs is empty.

samsja: yes but I think this can be avoided

samsja: > I believe is not a bad strategy to make API compatible with other Indexes

I think we can achieve this without `**kwargs`

JohannesMessner:
I added an .md with instructions for creating a new Doc Index class, I'll address the **kwargs comments tomorrow

docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -0,0 +1,400 @@
...
+if torch_imported:
+    import torch
+
+    HNSWLIB_PY_VEC_TYPES.append(torch.Tensor)
+
+
+class HnswDocumentIndex(BaseDocumentIndex, Generic[TSchema]):
+    def __init__(self, db_config=None, **kwargs):
+        super().__init__(db_config=db_config, **kwargs)
+        self._db_config = cast(HnswDocumentIndex.DBConfig, self._db_config)
```
AnneYang720, line 57: can we move this line to `super().__init__` as well? like
`self._db_config = cast(self.DBConfig, self._db_config)`

docs/tutorials/add_doc_index.md
```diff
@@ -0,0 +1,200 @@
...
+
+You can declare allowed fields and default values for your `_db_config`, but you will see that later.
+
+### The `_runtime_config`
+
+The `_runtime_config` is a dataclass that contains all "dynamic" configurations of your Document Index.
+Users can pass these configurations to the `.configure()` method of your Document Index, and `self._runtime_config` will be populated
+for you, so that you can use it in your implementation.
+
+You can declare allowed fields and default values for your `_db_config`, but you will see that later.
```
AnneYang720, line 89:
```suggestion
You can declare allowed fields and default values for your `_runtime_config `, but you will see that later.
```

docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -0,0 +1,400 @@
...
+        return fn
+
+else:
+    # static type checkers do not like callable objects as decorators
+    from docarray.doc_index.abstract_doc_index import composable
+
+TSchema = TypeVar('TSchema', bound=BaseDocument)
+T = TypeVar('T', bound='HnswDocumentIndex')
+
+HNSWLIB_PY_VEC_TYPES = [list, tuple, np.ndarray]
```
AnneYang720, line 47: what about basic python type like `str`, `int`...

docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -0,0 +1,400 @@
...
+
+    ###############################################
+    # Implementation of abstract methods          #
+    ###############################################
+
+    def python_type_to_db_type(self, python_type: Type) -> Any:
+        """Map python type to database type."""
+        for allowed_type in HNSWLIB_PY_VEC_TYPES:
+            if issubclass(python_type, allowed_type):
+                return np.ndarray
```
AnneYang720, line 130: return `np.ndarray` or `allowed_type `

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: No I don't think so. The issue is that mypy wants `self._db_config` to be of type `HnswDocumentIndex.DBConfig`, and not of type `AbstractDocumentIndex.DBConfig`, and the abstract class (`super()`) does not know about `HnswDocumentIndex.DBConfig`.
But luckily this is only about mypy, the code will run perfectly fine also without this line

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: I still need to add support for those, but will do that in a separate PR.

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: I don't think that implements the same logic

docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -0,0 +1,400 @@
...
+    @dataclass
+    class DBConfig(BaseDocumentIndex.DBConfig):
+        work_dir: str = '.'
+
+    @dataclass
+    class RuntimeConfig(BaseDocumentIndex.RuntimeConfig):
+        default_column_config: Dict[Type, Dict[str, Any]] = field(
+            default_factory=lambda: {
+                np.ndarray: {
+                    'dim': 128,
```
AnneYang720, line 108: do we need to keep this `dim` here when class `_Column` has `n_dim`?

docs/tutorials/add_doc_index.md
```diff
@@ -0,0 +1,200 @@
+# Add a new Document Index
```
samsja, line 1: can you add it in the documentation ?

JohannesMessner: ok

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: good point, let me double check this

docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -0,0 +1,400 @@
...
+    # general helpers
+    @staticmethod
+    def _to_hashed_id(doc_id: str) -> int:
+        # https://stackoverflow.com/questions/16008670/how-to-hash-a-string-into-8-digits
+        # hashing to 18 digits avoids overflow of sqlite INTEGER
+        return int(hashlib.sha256(doc_id.encode('utf-8')).hexdigest(), 16) % 10**18
+
+    def _load_index(self, col_name: str, col: '_Column') -> hnswlib.Index:
+        """Load an existing HNSW index from disk."""
+        index = self._create_index_class(col)
```
AnneYang720, line 307: can we create a `hnswlib.Index` with no params instead if it will load content from disk on next line

AnneYang720: I just read hnswlib documentation. no problem here

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: Ok to clarify: The `n_dim` in the `_Column` is taken from the type parameter, e.g. `NdArray[512]`, then `n_dim` will be 512. The `dim` here is just a parameter that people can pass to `Field()`. So in the `_Column` it could be that `n_dim` is empty while `dim` isn't, or vice versa, or both are empty, etc.

docarray/doc_index/backends/hnswlib_doc_index.py
JohannesMessner: I'll keep it in mind for the follow-up PR that refines the HNSWlib implementation :)

samsja (approved):

samsja (changes requested):
can we mark all of the test as "slow" and "docstore" related ?

JohannesMessner:
> can we mark all of the test as "slow" and "docstore" related ?

yep makes sense

tests/doc_index/hnswlib/test_index_get_del.py
```diff
@@ -0,0 +1,246 @@
+import numpy as np
+import pytest
+from pydantic import Field
+
+from docarray import BaseDocument, DocumentArray
+from docarray.doc_index.backends.hnswlib_doc_index import HnswDocumentIndex
+from docarray.typing import NdArray
+
+pytestmark = [pytest.mark.slow, pytest.mark.doc_index]
```
samsja, line 9: oh did not know you could do this interesting

samsja (approved):

PR #1133, opened 2023-02-14
feat: add schema map in protoload

samsja:
# Context

This pr allow loading from protobuf with a field mapping.

## what this pr do

```python
from docarray.typing import AnyUrl, TorchTensor

import torch


class A(BaseDocument):
    url: AnyUrl
    tensor: TorchTensor


class B(BaseDocument):
    link: AnyUrl
    array: TorchTensor


a = A(url='hello', tensor=torch.zeros(3))

b = B.from_protobuf_with_schema_map(
    a.to_protobuf(), field_map={'url': 'link', 'tensor': 'array'}
)

assert b.link == a.url
assert (b.array == a.tensor).all()
```

### DIFF
--- a/docarray/array/array.py
+++ b/docarray/array/array.py
@@ -7,6 +7,7 @@ from typing import (
     Generic,
     Iterable,
     List,
+    Mapping,
     Optional,
     Sequence,
     Type,
@@ -316,9 +317,25 @@ class DocumentArray(AnyDocumentArray, Generic[T_doc]):
 
     @classmethod
     def from_protobuf(cls: Type[T], pb_msg: 'DocumentArrayProto') -> T:
-        """create a Document from a protobuf message"""
+        """create a DocumentArray from a protobuf message"""
+        return cls.from_protobuf_field_map(pb_msg, None)
+
+    @classmethod
+    def from_protobuf_field_map(
+        cls: Type[T],
+        pb_msg: 'DocumentArrayProto',
+        field_map: Optional[Mapping[str, str]] = None,
+    ) -> T:
+        """
+        create a DocumentArray from a protobuf message with a field_map. Apply the same
+        behavior of {meth}`~docarray.BaseDocument.from_protobuf_field_map`
+        :param pb_msg:
+        :param field_map:
+        :return: a DocumentArray initialized from a proto message
+        """
         return cls(
-            cls.document_type.from_protobuf(doc_proto) for doc_proto in pb_msg.docs
+            cls.document_type.from_protobuf_field_map(doc_proto, field_map)
+            for doc_proto in pb_msg.docs
         )
 
     def to_protobuf(self) -> 'DocumentArrayProto':
--- a/docarray/base_document/mixins/proto.py
+++ b/docarray/base_document/mixins/proto.py
@@ -84,14 +84,14 @@ class ProtoMixin(AbstractDocument, BaseNode):
 
                 value = pb_msg.data[field_name]
 
-                fields[field_name_mapped] = cls._proto_get_content_from_node_proto(
+                fields[field_name_mapped] = cls._get_content_from_node_proto(
                     value, field_name_mapped
                 )
 
         return cls(**fields)
 
     @classmethod
-    def _proto_get_content_from_node_proto(cls, value: 'NodeProto', field_name):
+    def _get_content_from_node_proto(cls, value: 'NodeProto', field_name):
         """
         load the proto data from a node proto
 
--- a/docarray/base_document/mixins/proto.py
+++ b/docarray/base_document/mixins/proto.py
@@ -70,8 +70,6 @@ class ProtoMixin(AbstractDocument, BaseNode):
             value.type if value.WhichOneof('docarray_type') is not None else None
         )
 
-        return_field = None
-
         if content_type in content_type_dict:
             return_field = content_type_dict[content_type].from_protobuf(
                 getattr(value, content_key)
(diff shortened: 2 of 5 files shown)


JohannesMessner:
Is there a fundamental reason this only works from protobuf? I imagine it would be equally as useful to be able to do this from dict/json, or just from another non-serialized Document.

docarray/base_document/mixins/proto.py
```diff
@@ -15,75 +14,147 @@
...
+
+    @classmethod
+    def from_protobuf_field_map(
+        cls: Type[T],
+        pb_msg: 'DocumentProto',
+        field_map: Optional[Mapping[str, str]] = None,
+    ) -> T:
+        """create a Document from a protobuf message. You can specify a mapping
+        that will be used to convert key name from the proto to the class fields field
+        name by using the `field_map` field.
```
JohannesMessner, line 33:
```suggestion
        """Create a Document from a protobuf message. You can optionally specify a mapping
        that will be used to map keys in the protobuf definition to keys (names) of the Document fields.
```

docarray/base_document/mixins/proto.py
```diff
@@ -15,75 +14,147 @@
...
-
-                    fields[field] = tuple(MessageToDict(value.tuple))
-                elif content_key == 'dict':
-                    from google.protobuf.json_format import MessageToDict
-
-                    fields[field] = MessageToDict(value.dict)
+            assert b.link == a.url
+            assert (b.array == a.tensor).all()
+
+        :param pb_msg: the proto message of the Document
```
JohannesMessner, line 61:
```suggestion
        :param pb_msg: The proto message representing the Document
```

docarray/base_document/mixins/proto.py
```diff
@@ -15,75 +14,147 @@
...
+        :param field_map: map proto key to schema key
```
JohannesMessner, line 62:
```suggestion
        :param field_map: Optional mapping from proto fields to Document fields
```

docarray/base_document/mixins/proto.py
```diff
@@ -15,75 +14,147 @@
...
+        :return: a Document initialize with the proto data
```
JohannesMessner, line 63:
```suggestion
        :return: A new Document initialized with the data from `pb_msg`
```

docarray/base_document/mixins/proto.py
```diff
@@ -15,75 +14,147 @@
...
+
+            do_pass_loop = False
+            if field_name not in cls.__fields__.keys():
+                if field_name in field_map.keys():
+                    field_name_mapped = field_map[field_name]
                 else:
-                    raise ValueError(
-                        f'key {content_key} is not supported for' f' deserialization'
-                    )
+                    do_pass_loop = True  # optimization we don't even load the data if the key does not
```
JohannesMessner, line 77: I think we can remove the entire `do_pass_loop` variable and logic and instead `continue` here

docarray/base_document/mixins/proto.py
```diff
@@ -15,75 +14,147 @@
...
+                value = pb_msg.data[field_name]
+
+                fields[field_name_mapped] = cls._proto_get_content_from_node_proto(
+                    value, field_name_mapped
+                )
+
+        return cls(**fields)
+
+    @classmethod
+    def _proto_get_content_from_node_proto(cls, value: 'NodeProto', field_name):
```
JohannesMessner, line 95:
```suggestion
    def _get_content_from_node_proto(cls, value: 'NodeProto', field_name):
```

JohannesMessner (changes requested):

docarray/base_document/mixins/proto.py
samsja: shit I know I was doing something wrong I was using pass instead of continue

samsja:
> Is there a fundamental reason this only works from protobuf? I imagine it would be equally as useful to be able to do this from dict/json, or just from another non-serialized Document.

Good question. I hesitated to do it at the dict level. The things is it is more efficient to do it at each serialization level (you can drop data that you don't want to load). Though I have to say I am not sure it is a common practice to do this in the json/rest wold :thinking: 

I would say we can merge this one and see for the other serialization later.

JohannesMessner:
> Though I have to say I am not sure it is a common practice to do this in the json/rest wold thinking

Maybe not, but I think we should not favour the protobuf/grpc world over the json/http world if we don't have to. We aspire to be a library for _sending_ multi-modal data, so I would say features that help in that should not be restricted to one tech stack.

samsja:
Then I will do the same for JSON in this pr @JohannesMessner

docarray/array/array.py
```diff
@@ -316,9 +317,25 @@ def _set_array_attribute(
...
+    def from_protobuf_field_map(
+        cls: Type[T],
+        pb_msg: 'DocumentArrayProto',
+        field_map: Optional[Mapping[str, str]] = None,
+    ) -> T:
+        """
+        create a DocumentArray from a protobuf message with a field_map. Apply the same
+        behavior of {meth}`~docarray.BaseDocument.from_protobuf_field_map`
+        :param pb_msg:
+        :param field_map:
```
JohannesMessner, line 333: missing param description

JohannesMessner (changes requested):
just one small thing, the rest LGTM

docarray/array/array.py
```diff
@@ -316,9 +317,25 @@ def _set_array_attribute(
 
     @classmethod
     def from_protobuf(cls: Type[T], pb_msg: 'DocumentArrayProto') -> T:
```
JoanFM, line 319: why don't we extend `from_protobuf` with a new Optional argument?

JoanFM (changes requested):

samsja:
on hold for now

Issue #1137, opened 2023-02-15
Handle da inside da in stack mode


AnneYang720:
```python

from docarray import BaseDocument, DocumentArray
from docarray.typing import AnyTensor

class Image(BaseDocument):
    tensor : AnyTensor

class Video(BaseDocument):
    images: DocumentArray[Image]



da = DocumentArray[Video]( Video(images = DocumentArray[Image]([ Image()])])

da.stack()
da[0].images # is not stack but should be

```

in this case `da.stack()` will not stack the tensor in Image because of the nested DocumentArray

this will probably be solved by introducing the concept of dimension in document array

PR #1138, opened 2023-02-15
feat(v2): stack da inside da

AnneYang720:
Goals:
Make function `stack` work on nested DocumentArray inside DocumentArray, issue #1137 
- [x] change function `stack` and `unstack`
- [x] add tests
- [x] check and update documentation, if required (not updated)


docarray/array/array.py
```diff
@@ -408,8 +408,10 @@ def validate(
         value: Union[T, Iterable[BaseDocument]],
         field: 'ModelField',
         config: 'BaseConfig',
-    ) -> T:
-        if isinstance(value, cls):
+    ):
```
samsja, line 411:
```suggestion
    )-> Union[T, DocumentArrayStacked]:
```

docarray/array/array_stacked.py
```diff
@@ -90,11 +90,23 @@ def from_iterable_document(
                 docs, tensor_type=self.tensor_type
             )
         )
+        self._nested_stack(self._docs)
         self.tensor_type = self._docs.tensor_type
         self._doc_columns, self._tensor_columns = self._create_columns(
             self._docs, tensor_type=self.tensor_type
         )
 
+    def _nested_stack(self: T, docs: DocumentArray):
```
samsja, line 99: I believe you created this function because the DA field will not be collect in `_get_columns_schema` right ? If yes I believe we might just drop this `_get_columns_schema` and merge your for loop with the one inside `_create_columns`.

smth like this

```python
for field, type_ in cls.__fields__.items():
    if issubclas(type_, AbstractTensor):
        ...
    elif issublcas(type_, BaseDocument):
        ...
    elif issubclass(type_, DocumentArray):
        ... # here this is the content of this PR
    else:
        pass
```

docarray/array/array.py
AnneYang720: > 

mypy isn't happy with this
```
 error: Return type "Union[DocumentArray[T_doc], DocumentArrayStacked]" of "validate" incompatible with return type "DocumentArray[T_doc]" in supertype "AbstractType"
```

samsja: hmm I see then lets do you way

docarray/array/array_stacked.py
```diff
@@ -165,67 +139,81 @@ def _create_columns(
...
 
-        column_schema = cls._get_columns_schema(tensor_type)
-
         doc_columns: Dict[str, DocumentArrayStacked] = dict()
         tensor_columns: Dict[str, AbstractTensor] = dict()
 
-        for field, type_ in column_schema.items():
-            if tf_available and isinstance(getattr(docs[0], field), TensorFlowTensor):
+        for field_name, field in cls.document_type.__fields__.items():
+            type_ = field.outer_type_
```
samsja, line 146:
```suggestion
            type_ = cls.document_type._get_field_type(field)

```

samsja: hmm I see

Issue #1131, opened 2023-02-13
`TorchEmbedding` cannot be deep copied


Jackmin801:
**Describe the bug**
`TorchEmbedding` cannot be deep copied.

**To Reproduce**
script.py
```python
import torch
from docarray import BaseDocument
from docarray.typing.tensor.embedding.torch import TorchEmbedding

class Meow(BaseDocument):
    embedding : TorchEmbedding

a = Meow(embedding=torch.randn(42))
a.copy(deep=True)
```

output
```console
Traceback (most recent call last):
  File "/Users/jackmin/Jina/docarray/oink.py", line 16, in <module>
    a.copy(deep=True)
  File "pydantic/main.py", line 657, in pydantic.main.BaseModel.copy
  File "pydantic/main.py", line 612, in pydantic.main.BaseModel._copy_and_set_values
  File "/Users/jackmin/miniconda3/envs/docarray/lib/python3.10/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/Users/jackmin/miniconda3/envs/docarray/lib/python3.10/copy.py", line 231, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Users/jackmin/miniconda3/envs/docarray/lib/python3.10/copy.py", line 153, in deepcopy
    y = copier(memo)
  File "/Users/jackmin/miniconda3/envs/docarray/lib/python3.10/site-packages/torch/_tensor.py", line 100, in __deepcopy__
    return handle_torch_function(Tensor.__deepcopy__, (self,), self, memo)
  File "/Users/jackmin/miniconda3/envs/docarray/lib/python3.10/site-packages/torch/overrides.py", line 1534, in handle_torch_function
    result = torch_func_method(public_api, types, args, kwargs)
  File "/Users/jackmin/Jina/docarray/docarray/typing/tensor/torch_tensor.py", line 225, in __torch_function__
    return super().__torch_function__(func, types_, args, kwargs)
  File "/Users/jackmin/miniconda3/envs/docarray/lib/python3.10/site-packages/torch/_tensor.py", line 1279, in __torch_function__
    ret = func(*args, **kwargs)
  File "/Users/jackmin/miniconda3/envs/docarray/lib/python3.10/site-packages/torch/_tensor.py", line 185, in __deepcopy__
    raise RuntimeError(
[...]

PR #1143, opened 2023-02-16
fix: enable torchembedding deepcopy

AnneYang720:
Goals:

This PR is related to #1131

### DIFF
--- a/tests/units/typing/tensor/test_torch_tensor.py
+++ b/tests/units/typing/tensor/test_torch_tensor.py
@@ -152,3 +152,17 @@ def test_parametrized_operations():
     assert isinstance(t_result, torch.Tensor)
     assert isinstance(t_result, TorchTensor)
     assert isinstance(t_result, TorchTensor[128])
+
+
+def test_deepcopy():
+    from docarray import BaseDocument
+
+    class MMdoc(BaseDocument):
+        embedding: TorchEmbedding
+
+    doc = MMdoc(embedding=torch.randn(32))
+    doc_copy = doc.copy(deep=True)
+
+    assert id(doc) != id(doc_copy)
+    assert id(doc.embedding) != id(doc_copy.embedding)
+    assert (doc.embedding == doc_copy.embedding).all()
--- a/tests/units/typing/tensor/test_torch_tensor.py
+++ b/tests/units/typing/tensor/test_torch_tensor.py
@@ -163,6 +163,8 @@ def test_deepcopy():
     doc = MMdoc(embedding=torch.randn(32))
     doc_copy = doc.copy(deep=True)
 
-    assert id(doc) != id(doc_copy)
-    assert id(doc.embedding) != id(doc_copy.embedding)
+    assert doc.embedding.data_ptr() != doc_copy.embedding.data_ptr()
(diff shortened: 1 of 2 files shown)


AnneYang720:
After the modification
```python
import torch
from docarray.typing.tensor.embedding.torch import TorchEmbedding

class Meow(BaseDocument):
    embedding : TorchEmbedding

a = Meow(embedding=torch.randn(42))
print(id(a.embedding), id(a.copy(deep=True).embedding))
```
output has no error
```
140248435632272 140248435636032
```

samsja:
Can u add to test to check that it is actual deep copied and not shallow copy ?

tests/units/typing/tensor/test_torch_tensor.py
```diff
@@ -152,3 +152,17 @@ def test_parametrized_operations():
...
+    from docarray import BaseDocument
+
+    class MMdoc(BaseDocument):
+        embedding: TorchEmbedding
+
+    doc = MMdoc(embedding=torch.randn(32))
+    doc_copy = doc.copy(deep=True)
+
+    assert id(doc) != id(doc_copy)
+    assert id(doc.embedding) != id(doc_copy.embedding)
```
samsja, line 167: I think that this test is not enough.

Two pytorch variable can have different ID, which mean they are different variable, but still point to the same underlying data storage. 

You can use `tensor.data_ptr()` to get the pytorch pointer to the actual data (we use it in a couple of otehr test).


Another thing that you can do is to actually check that if you changed the value of `doc_copy` it does not affect `doc`

JohannesMessner:
Can you explain this fix briefly? It is not obvious to me why this is fixing it

AnneYang720:
> Can you explain this fix briefly? It is not obvious to me why this is fixing it

The deepcopy error is thrown from here https://github.com/pytorch/pytorch/blob/master/torch/_tensor.py#L170, and line above is 
```python
new_tensor = self.new_empty([])
if type(new_tensor) is not type(self):
```
where `type(new_tensor)` is  `<class 'torch.Tensor'>` and 
`type(self)` is `<class 'docarray.typing.tensor.embedding.torch.TorchEmbedding'>`.

So according to the error msg, I should implement `new_empty()` for the subclass, which means the `new_tensor` should also be a `TorchEmbedding`.

samsja:
> > Can you explain this fix briefly? It is not obvious to me why this is fixing it
> 
> The deepcopy error is thrown from here https://github.com/pytorch/pytorch/blob/master/torch/_tensor.py#L170, and lines above are
> 
> ```python
> new_tensor = self.new_empty([])
> if type(new_tensor) is not type(self):
> ```
> 
> where `type(new_tensor)` is `<class 'torch.Tensor'>` and `type(self)` is `<class 'docarray.typing.tensor.embedding.torch.TorchEmbedding'>`.
> 
> So according to the error msg, I should implement `new_empty()` for the subclass, which means the `new_tensor` should also be a `TorchEmbedding`.

@AnneYang720  can u add this to the docstring of `new_empty` in your PR so that ppl who will read the code will understand the same ?

samsja (approved):

PR #1144, opened 2023-02-17
feat(v2): load da from csv and save to csv

anna-charlotte:
Goals:

- [x] load DocumentArray from csv, with a given Document class to provide a schema
  - `document_type` needs to be set to determine schema.
  - column names have to match the Document types fields
  - for nested fields, use access paths as column name, such as `'image__url'`
  - if value is None, leave csvfield empty, or enter `None`
- [x] save a DocumentArray to csv
  - following the same guidelines as above.

### DIFF
--- a/docarray/display/document_array_summary.py
+++ b/docarray/display/document_array_summary.py
@@ -38,7 +38,7 @@ class DocumentArraySummary:
 
                 if isinstance(val, AbstractTensor):
                     comp_be = val.get_comp_backend()
-                    if comp_be.isnan(val).all():
+                    if comp_be.to_numpy(comp_be.isnan(val)).all():
                         col_2 = f'None ({val.__class__.__name__})'
                     else:
                         col_2 = (


JohannesMessner:
Will nested schemas work somehow?

tests/units/array/test_array_from_to_csv.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class MyDoc(BaseDocument):
+    count: Optional[int]
+    text: str
+
+
+class MyDocNested(MyDoc):
+    image: Image
+    image2: Image
```
samsja, line 23: either create fixture or copy paste the doc in the tests.

otherwise one test could modify one class and it could impact another test

tests/toydata/docs_nested.csv
```diff
@@ -0,0 +1,4 @@
+count,text,image,image2.url
```
samsja, line 1:
```suggestion
count,text,image,image2__url
```

Not really sure here but I think we cannot/should not use `.` inside column name

tests/units/array/test_array_from_to_csv.py
```diff
@@ -0,0 +1,153 @@
...
+        img: Optional[Image]
+        middle: Optional[Middle]
+
+    doc = Outer(img=Image(), middle=Middle(img=Image(), inner=Inner(img=Image())))
+    return doc
+
+
+def test_from_csv_without_schema_raise_exception():
+    with pytest.raises(TypeError, match='no document schema defined'):
+        DocumentArray.from_csv(file_path=str(TOYDATA_DIR / 'docs_nested.csv'))
```
samsja, line 95: Makes sense that this raise an Expection for now. But what about implementing at this level the auto schema detection ?

tests/toydata/docs_nested.csv
samsja: Yeah kind of because, though it is more gut feeling that actual knowledge about CSV convention. But generally I think we should keep with `a__b` sintax rather than `a.b` one. Same for vector database I believe @JohannesMessner

tests/units/array/test_array_from_to_csv.py
samsja: yes makes sense to do it in another PR

docarray/array/array/io.py
```diff
@@ -291,6 +296,69 @@ def to_json(self) -> str:
...
+        from docarray import DocumentArray
+
+        doc_type: Type[BaseDocument] = cls.document_type
+        if doc_type == AnyDocument:
+            raise TypeError(
+                'There is no document schema defined. '
+                'To load from csv, please specify the DocumentArray\'s document type.'
+            )
+
+        da = DocumentArray[doc_type]()  # type: ignore
```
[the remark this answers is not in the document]
samsja: https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases

> The main difference is that the target of an alias is precisely known statically, and this means that they can be used in type annotations and other type contexts

I think this just mean than you cannot do type things with variable but only with TypeAlias. So here I guess that you should try to do a TypeAlias

docarray/array/array/io.py
```diff
@@ -291,6 +296,69 @@ def to_json(self) -> str:
         """
         return json.dumps([doc.json() for doc in self])
 
+    @classmethod
+    def from_csv(cls, file_path: str, encoding: str = 'utf-8') -> 'DocumentArray':
+        """
+        Load a DocumentArray from a csv file following the schema defined in the
+        :attr:`~docarray.DocumentArray.document_type` attribute.
+        The column names have to match the fields of the document type.
```
JohannesMessner, line 304:
```suggestion
        The column names have to match the field names of the Document type.
```

docarray/array/array/io.py
```diff
@@ -291,6 +296,69 @@ def to_json(self) -> str:
...
+        For nested fields use dot-separated access paths, such as 'image__url'.
```
JohannesMessner, line 305: Can we handle DocumentArray (list-like) fields? I'm guessing not, does it even make sense in csv? either way, I'd mention it here

docarray/array/array/io.py
```diff
@@ -291,6 +296,69 @@ def to_json(self) -> str:
...
+        :param encoding: encoding used to read the csv file. Defaults to 'utf-8'.
+        :return: DocumentArray
+        """
+        from docarray import DocumentArray
+
+        doc_type: Type[BaseDocument] = cls.document_type
+        if doc_type == AnyDocument:
+            raise TypeError(
+                'There is no document schema defined. '
+                'To load from csv, please specify the DocumentArray\'s document type.'
```
JohannesMessner, line 317:
```suggestion
                'To load from csv, please specify the DocumentArray\'s Document type using `DocumentArray[MyDoc]`.'
```

docarray/array/array/io.py
```diff
@@ -291,6 +296,69 @@ def to_json(self) -> str:
...
         return json.dumps([doc.json() for doc in self])
 
+    @classmethod
+    def from_csv(cls, file_path: str, encoding: str = 'utf-8') -> 'DocumentArray':
+        """
+        Load a DocumentArray from a csv file following the schema defined in the
+        :attr:`~docarray.DocumentArray.document_type` attribute.
+        The column names have to match the fields of the document type.
+        For nested fields use dot-separated access paths, such as 'image__url'.
+
```
JohannesMessner, line 306: Also I'm guessing every row in the csv will be mapped to one doc in the array? maybe worth mentioning?

docarray/array/array/io.py
```diff
@@ -291,6 +296,69 @@ def to_json(self) -> str:
...
+        doc_type: Type[BaseDocument] = cls.document_type
+        if doc_type == AnyDocument:
+            raise TypeError(
+                'There is no document schema defined. '
+                'To load from csv, please specify the DocumentArray\'s document type.'
+            )
+
+        da = DocumentArray[doc_type]()  # type: ignore
+        with open(file_path, 'r', encoding=encoding) as fp:
+            lines = csv.DictReader(fp, dialect='excel')
```
JohannesMessner, line 322: I think the `dialect` should be passed as an (optional) kwarg to the method

docarray/array/array/io.py
```diff
@@ -291,6 +296,69 @@ def to_json(self) -> str:
...
+            fields = lines.fieldnames
+            if fields is None:
+                raise TypeError("No field names are given.")
+
+            valid = [is_access_path_valid(doc_type, field) for field in fields]
+            if not all(valid):
+                raise ValueError(
+                    f'Fields provided in the csv file do not match the schema of the DocumentArray\'s '
+                    f'document type ({doc_type.__name__}): {list(compress(fields, [not v for v in valid]))}'
```
JohannesMessner, line 331: TIL about `compress`,  nice :)

docarray/array/array/io.py
```diff
@@ -291,6 +296,69 @@ def to_json(self) -> str:
...
+        if doc_type == AnyDocument:
+            raise TypeError(
+                'There is no document schema defined. '
+                'To load from csv, please specify the DocumentArray\'s document type.'
+            )
+
+        da = DocumentArray[doc_type]()  # type: ignore
+        with open(file_path, 'r', encoding=encoding) as fp:
+            lines = csv.DictReader(fp, dialect='excel')
+            fields = lines.fieldnames
```
JohannesMessner, line 323: What's the type of `lines` and `fields`? I think a type annotation would be nice here, would make it easier to understand the code further down below

docarray/array/array/io.py
```diff
@@ -530,3 +598,85 @@ def save_binary(
             file_ctx=file_ctx,
             show_progress=show_progress,
         )
+
+
+def is_access_path_valid(doc: Type['BaseDocument'], access_path: str) -> bool:
+    """
+    Check if a given access path is a valid path for a given Document class.
+    """
+    field, _, remaining = access_path.partition('__')
```
JohannesMessner, line 607: Didn't we use "dot-notation" in other places, e.g. `nested.field` instead of `nested__field`? We need to make sure that this is uniform everywhere.

docarray/array/array/io.py
```diff
@@ -530,3 +598,85 @@ def save_binary(
...
+    result = {}
+    for k, v in d.items():
+        if isinstance(v, dict):
+            v = _dict_to_access_paths(v)
+            for nested_k, nested_v in v.items():
+                new_key = '__'.join([k, nested_k])
+                result[new_key] = nested_v
+        else:
+            result[k] = v
+    return result
```
JohannesMessner, line 655: We should consider making these methods global helpers, they will come in handy in other places of the code base as well, e.g. i already implemented something similar for document stores

JohannesMessner (changes requested):
Some comments, but I think this will be a truly cool feature! We should promote this properly once it is out, I bet the data validation folks that already use pydantic would love it

docarray/array/array/io.py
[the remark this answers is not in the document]
samsja: yes you can add `typing_extension` as a dependency

docarray/array/array/io.py
JohannesMessner: yes exactly, in a `helper.py` or `utils.py` at the top level or something like that

samsja (approved):

JohannesMessner (approved):
Nice feature!

PR #1147, opened 2023-02-17
fix: better generic support in document array stack

samsja:
# Context

Make AbstractArray a generic compatbile with type var and string

fix this : https://github.com/docarray/docarray/issues/1146

```python
from docarray import DocumentArray, BaseDocument
from typing import TypeVar, List

T = TypeVar('T', bound=BaseDocument)


def f(a: DocumentArray[T]) -> DocumentArray[T]:
    return a


def g(a: DocumentArray['BaseDocument']) -> DocumentArray['BaseDocument']:
    return a

```
this work now


docarray/array/abstract_array.py
```diff
@@ -34,7 +35,10 @@ def __repr__(self):
         return f'<{self.__class__.__name__} (length={len(self)})>'
 
     @classmethod
-    def __class_getitem__(cls, item: Type[BaseDocument]):
+    def __class_getitem__(cls, item: Union[Type[BaseDocument], TypeVar, str]):
+        if not isinstance(item, type):
+            return Generic.__class_getitem__.__func__(cls, item)  # type: ignore
```
Jackmin801, line 40: How did you know this is what is required?

samsja: > is this even doing anything?

this is not doing anything else than checking the the `item` is valid Generic. I think it does not harm to keep it

Jackmin801 (approved):
Tried with pydantic generic model. Seems to work

docarray/array/abstract_array.py
954-Ivory: > > is this even doing anything?
> 
> this is not doing anything else than checking the the `item` is valid Generic. I think it does not harm to keep it

Crashed in Python 3.12.0
```
    return Generic.__class_getitem__.__func__(cls, item)  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'builtin_function_or_method' object has no attribute '__func__'. Did you mean: '__doc__'?
```

PR #1148, opened 2023-02-20
fix: update tensor in stack mode

samsja:
# Context

wip, fix : https://github.com/docarray/docarray/issues/1081 I will add more info once I am done with the implementation

## What this pr do ?

each time we modify a field in a Document that is inside document array stacked, the change will apply back to the document array stacked
```python


    class Image(BaseDocument):
        tensor: TorchTensor[3, 224, 224]

    class MMdoc(BaseDocument):
        image: Image

    batch = DocumentArray[MMdoc](
        [MMdoc(image=Image(tensor=torch.zeros(3, 224, 224))) for _ in range(10)]
    )

    batch = batch.stack()

    batch[0].image = Image(tensor=torch.ones(3, 224, 224))

    assert (batch.image[0].tensor == torch.ones(3, 224, 224)).all()

```

same for tensor

```python

    class Image(BaseDocument):
        tensor: TorchTensor[3, 224, 224]

    batch = DocumentArray[Image](
        [Image(tensor=torch.zeros(3, 224, 224)) for _ in range(10)]
    )

    batch = batch.stack()

    batch[0].tensor = torch.ones(3, 224, 224)

    assert (batch.tensor[0] == torch.ones(3, 224, 224)).all()
```


JohannesMessner:
For archival purposes I'm adding the Discord discussion around this issue: https://discord.com/channels/1062677585999429653/1078671075761725560/1078685668617769060

JoanFM (changes requested):
I would like to understand better how the view is implemented in Pytorch. Keeping a circular dependency does not seem to me as appropiate

JoanFM:
What happens if we put the Document into another `DocumentArrayStacked`? Then to whom does it belong?

samsja:
> I would like to understand better how the view is implemented in Pytorch. Keeping a circular dependency does not seem to me as appropiate

In Pytorch they have a circular dependency as well. The view and the original object share the same storage object.

Btw I wanted to use weakref originally to avoid memory problem. Turn out it messed up the multiprocessing part (not sure why). But I found out that python is able to deal with cyclic reference so it is not a problem

samsja:
> > I would like to understand better how the view is implemented in Pytorch. Keeping a circular dependency does not seem to me as appropiate
> 
> In Pytorch they have a circular dependency as well. The view and the original object share the same storage object.
> 
> Btw I wanted to use weakref originally to avoid memory problem. Turn out it messed up the multiprocessing part (not sure why). But I found out that python is able to deal with cyclic reference so it is not a problem

But maybe @JoanFM you are right I should be more careful. Should I add a test to see if the GC handles it properly?

samsja:
@JoanFM what would be an alternative according to you (I am open to any alternative tbh). The thing is that conceptually if a document update triggers a da update then they need to somehow be linked right ?

samsja:
> What happens if we put the Document into another `DocumentArrayStacked`? Then to whom does it belong?

This is not possible. The new DocumentArray will check that there are no `_da_ref` in the document before allowing him.

Basically, oncee a Document is appended to a DocumentArrayStack it is the da that has full ownership of the doc

docarray/array/stacked/array_stacked.py
```diff
@@ -98,6 +98,20 @@ def from_iterable_document(
...
+        self._add_ref_to_docs()
+
+    def _add_ref_to_docs(self) -> None:
+        """Add a ref to self to the document"""
+        from docarray.base_document.document import _DaRef
+
+        for i, doc in enumerate(self._docs):
+            if doc._da_ref is not None and doc._da_ref.ref is not self:
+                raise ValueError(
+                    f'{doc} already belong to {doc._da_ref} and cannot be stack again'
```
JohannesMessner, line 110:
```suggestion
                    f'{doc} already belongs to {doc._da_ref} and cannot be stacked again'
```
Maybe I would say something like `doc is a view of [...] and cannot be stacked again`

docarray/array/stacked/array_stacked.py
```diff
@@ -513,3 +535,41 @@ def traverse_flat(
             return flattened[0]
         else:
             return flattened
+
+    def stacked_field(self) -> Iterable[str]:
```
JohannesMessner, line 539:
```suggestion
    def stacked_fields(self) -> Iterable[str]:
```

docarray/array/stacked/array_stacked.py
```diff
@@ -513,3 +535,41 @@ def traverse_flat(
...
+        """
+        :return: return the field name of this document array stacked which are stack
+        into a column
+        """
+        return chain(self._doc_columns.keys(), self._tensor_columns.keys())
+
+    def __update_columns__(self, field_name: str, da_index: int, value: Any) -> Any:
```
JohannesMessner, line 546: why is this a dunder method?

docarray/array/stacked/array_stacked.py
```diff
@@ -513,3 +535,41 @@ def traverse_flat(
...
+        """This function allow to update a column value in place
```
JohannesMessner, line 547:
```suggestion
        """This function allows to update a column value in-place
```

docarray/array/stacked/array_stacked.py
```diff
@@ -513,3 +535,41 @@ def traverse_flat(
...
+        :param field_name: the name of the column to update
+        :param da_index: the index of the column to update
+        :param value: the value to update the column with
+        :return: return the updated view of the column with the value data
```
JohannesMessner, line 551:
```suggestion
        :return: return the updated view of the column, at position `da_index`, with the `value` data
```

docarray/array/stacked/array_stacked.py
JohannesMessner, line 546: can value really be Any?

docarray/array/stacked/array_stacked.py
```diff
@@ -513,3 +535,41 @@ def traverse_flat(
...
+                issubclass(field_type, TensorFlowTensor)
+                or isinstance(value, TensorFlowTensor)
+                or isinstance(value, tf.Tensor)
+            ):
+                raise RuntimeError(
+                    'Tensorflow tensor are immutable, Since this '
+                    'Document is stack you cannot change its '
+                    'tensor value. You should either unstack your '
+                    'DocumentArrayStacked or work at the '
+                    'DocumentArrayStack column level directly'
```
JohannesMessner, line 568:
```suggestion
                    'Tensorflow tensors are immutable, therefore '
                    'you cannot change the tensor value of a stacked '
                    'DocumentArray with Tensorflow. You should either unstack your '
                    'DocumentArrayStacked or work at the '
                    'DocumentArrayStacked column level directly.'
```

JohannesMessner, line 568: not sure what this "work at column level directly" means

docarray/base_document/document.py
```diff
@@ -59,3 +74,32 @@ def schema_summary(cls) -> None:
     def _ipython_display_(self):
         """Displays the object in IPython as a summary"""
         self.summary()
+
+    def _is_inside_da_stack(self) -> bool:
+        """
+        :return: return true if the Document is inside a DocumentArrayStack
+        """
+        return self._da_ref is not None
```
JohannesMessner, line 82: I would make this public and call it `is_view` or something like that

docarray/base_document/document.py
```diff
@@ -1,24 +1,39 @@
...
+
+
 _console: Console = Console()
 
+object_setattr = object.__setattr__
+
+
+@dataclass
+class _DaRef:
+    ref: 'DocumentArrayStacked'
```
JohannesMessner, line 25: I would call this `parent_da` or something like that instead of `ref`, which is very non-descript

JohannesMessner (changes requested):
I have three questions:
1. when stacking a da, the small tensor that get stacked up get "destroyed", right?
2. for this PR, would it make sense to define a dedicated class for Documents that are in a stacked DA, something like `DocumentView`? I think it would make code separation a bit clearer, and also clarify some concepts.
3. If we do 2., then would it make sense to store all data on the array level, and nothing on the document level (for `DocumentArrayStacked` only)? Even if it isn't truly stacked, then it would be just a list. And accessing a Document in the stacked da returns a `DocumentView`, which is rally just a view into the array? I think it would complete our `view` narrative and make it as close to torch, np etc. as possible

samsja:
new PR : https://github.com/docarray/docarray/pull/1183

Issue #998, opened 2023-01-07
Support m4a audio format in v2


samsja:
# Context

We should be able to support other audio format like m4a.

To do this we need to refactor our `load()` method in `AudioUrl` to use PyDub (which is already installed):

and we need to add test with other audio file extension

samsja:
@anna-charlotte maybe we should use https://github.com/jiaaro/pydub#installation for the audio loading

PR #1155, opened 2023-02-21
feat: support other audio format

AnneYang720:
Goals:

This PR is related to issue #998 
- [x] refactor `load()` method in AudioUrl to use PyDub
- [x] remove useless code
- [x] pass current tests
- [x] add tests for other format


docarray/typing/url/audio_url.py
```diff
@@ -66,9 +66,16 @@ class MyDoc(Document):
...
 
-        bytes_ = AudioBytes(self.load_bytes())
-        return bytes_.load()
+        bytes_ = self.load_bytes()
+        segment = AudioSegment.from_file(BytesIO(bytes_))
+
+        samples = segment.get_array_of_samples()
+        samples = np.array(samples)
+
+        samples_norm = samples / 2 ** (segment.sample_width * 8 - 1)
```
JohannesMessner, line 77: I have no idea about audio data so maybe this is a stupid question, but what are these numbers here? Are they guaranteed to be always these same numbers, no matter the data? And what does this transformation do?

JohannesMessner (changes requested):

docarray/typing/url/audio_url.py
AnneYang720: Line 74 and 75 are how to get ndarray from audio file with pydub AudioSegment.

.github/workflows/ci.yml
```diff
@@ -210,6 +214,7 @@ jobs:
           poetry install --all-extras
           poetry run pip install protobuf==3.19.0
           poetry run pip install tensorflow==2.11.0
+          sudo apt-get install --no-install-recommends ffmpeg
```
samsja, line 217: `--no-install-recommends` what is this doing ?

AnneYang720: Don't let apt to install some "recommended" (not needed) packages, to save a little bit of CI time.

samsja (approved):

docarray/typing/url/audio_url.py
```diff
@@ -66,9 +44,14 @@ class MyDoc(Document):
...
 
         """
+        from pydub import AudioSegment  # type: ignore
+
+        bytes_ = self.load_bytes()
+        segment = AudioSegment.from_file(BytesIO(bytes_))
+        samples = np.array(segment.get_array_of_samples())
 
-        bytes_ = AudioBytes(self.load_bytes())
-        return bytes_.load()
```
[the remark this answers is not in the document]
samsja: good point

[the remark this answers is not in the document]
AnneYang720: The problem is `AudioBytes` and its `load()` method are not currently used anywhere.

AnneYang720:
I added the function `save_to_audio_file`. During the development, I found the `frame_rate` of an audio file is essential, so I added an variable `frame_rate` for `class Audio`. So now it is
```python
audio = Audio(
    url='https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/toydata/hello.wav?raw=true'
)
audio.tensor, audio.frame_rate = audio.url.load()
```
just like in video we have
```python
vid.audio.tensor, vid.tensor, vid.key_frame_indices = vid.url.load()
```

It may not be very compact. Maybe I can wrap the return to a class like `VideoLoadResult`.

docarray/typing/tensor/audio/abstract_audio_tensor.py
```diff
@@ -43,6 +43,28 @@ def save_to_wav_file(
             f.setframerate(sample_rate)
             f.writeframes(self.to_bytes())
 
+    def save_to_audio_file(
```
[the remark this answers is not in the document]
samsja: yes

samsja (changes requested):
add sample rate to audio predefined doc

docarray/typing/url/audio_url.py
JohannesMessner: okk I see. Maybe we should have a comment like `# normalize amplitude to [-1, 1]`?

samsja:
> 
    * simply have a huge-ass list of all supported file extensions, move it to a different file if needed
    * if a path has an extension and it is not in that list, then fail
    * if a path does not have an extension, pass
  
I like this tradeoff. 

The goal of validation is mostly to catch if someone put an `file.png` instead of a `file.wav`.

docarray/typing/tensor/audio/abstract_audio_tensor.py
```diff
@@ -20,28 +19,36 @@ def to_bytes(self):
...
 
-    def save_to_wav_file(
+    def save(
         self: 'T',
         file_path: Union[str, BinaryIO],
-        sample_rate: int = 44100,
+        format: str = 'wav',
+        frame_rate: int = 44100,
         sample_width: int = 2,
+        *args,
```
samsja, line 28: I know we had this pattern a lot in v1, but I am not sure i lile it tbh.

What about having a `pydub_args: Dict[str, Any]` parameters

and do segment.export(..., **pydub_arg)

samsja (changes requested):
looks great, one comment about the save function

docarray/typing/tensor/audio/abstract_audio_tensor.py
samsja: @JohannesMessner what do u think ?

samsja (approved):

docarray/typing/tensor/audio/abstract_audio_tensor.py
```diff
@@ -20,28 +19,35 @@ def to_bytes(self):
...
 
-    def save_to_wav_file(
+    def save(
         self: 'T',
         file_path: Union[str, BinaryIO],
-        sample_rate: int = 44100,
+        format: str = 'wav',
+        frame_rate: int = 44100,
         sample_width: int = 2,
+        pydub_args: Dict[str, Any] = {},
```
samsja, line 28: can u add the corresponding docstring as well ? Just say that this arg are forwarded to the export function

samsja (approved):

JohannesMessner (approved):

PR #1159, opened 2023-02-21
fix: remove setetiem in getiem

samsja:
# Context

we don't need the setitem when calling `da[0]` anymore. This is because we store the tensor in the document as a view of the column


Jackmin801 (approved):
LGTM. But after stacking, does the test cover checking that indexing into the batch still behaves as expected?

samsja:
> does the test cover checking that indexing into

yes the test check this I believe

samsja:
I will check this in my other PR https://github.com/docarray/docarray/pull/1148

PR #1161, opened 2023-02-22
feat(v2): add from and to pandas df for documentarray

anna-charlotte:
Add from and to pandas to DocumentArray IOMixin
- [x] add `DocumentArray[Doc].from_pandas(df)`
- [x] add `da.to_pandas()`

### DIFF
--- a/docarray/array/array/io.py
+++ b/docarray/array/array/io.py
@@ -24,8 +24,6 @@ from typing import (
     Union,
 )
 
-import pandas as pd
-
 from docarray.base_document import AnyDocument, BaseDocument
 from docarray.helper import (
     _access_path_dict_to_nested_dict,
@@ -35,6 +33,7 @@ from docarray.helper import (
 from docarray.utils.compress import _decompress_bytes, _get_compress_ctx
 
 if TYPE_CHECKING:
+    import pandas as pd
 
     from docarray import DocumentArray
     from docarray.proto import DocumentArrayProto
@@ -394,7 +393,7 @@ class IOMixinArray(Iterable[BaseDocument]):
                 writer.writerow(doc_dict)
 
     @classmethod
-    def from_pandas(cls, df: pd.DataFrame) -> 'DocumentArray':
+    def from_pandas(cls, df: 'pd.DataFrame') -> 'DocumentArray':
         """
         Load a DocumentArray from a `pandas.DataFrame` following the schema
         defined in the :attr:`~docarray.DocumentArray.document_type` attribute.
@@ -405,6 +404,30 @@ class IOMixinArray(Iterable[BaseDocument]):
         such as 'image__url'.
 
         List-like fields (including field of type DocumentArray) are not supported.
+
+        EXAMPLE USAGE:
+
+        .. code-block:: python
+
+            import pandas as pd
+
+            from docarray import BaseDocument, DocumentArray
+
+
+            class Person(BaseDocument):
+                name: str
+                follower: int
+
+
+            df = pd.DataFrame(
+                data=[['Maria', 12345], ['Jake', 54321]], columns=['name', 'follower']
+            )
+
+            da = DocumentArray[Person].from_pandas(df)
+
+            assert da.name == ['Maria', 'Jake']
+            assert da.follower == [12345, 54321]
+
 
         :param df: pandas.DataFrame to extract Document's information from
         :return: DocumentArray where each Document contains the information of one
@@ -443,7 +466,7 @@ class IOMixinArray(Iterable[BaseDocument]):
 
         return da
 
-    def to_pandas(self) -> pd.DataFrame:
+    def to_pandas(self) -> 'pd.DataFrame':
         """
         Save a DocumentArray to a `pandas.DataFrame`.
         The field names will be stored as column names. Each row of the dataframe corresponds
@@ -453,6 +476,8 @@ class IOMixinArray(Iterable[BaseDocument]):
 
         :return: pandas.DataFrame
         """
+        import pandas as pd
+
         fields = self.document_type._get_access_paths()
         df = pd.DataFrame(columns=fields)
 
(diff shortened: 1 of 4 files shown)


docarray/array/array/io.py
```diff
@@ -302,6 +304,56 @@ def to_json(self) -> str:
...
+            is_access_path_valid(cls.document_type, field) for field in field_names
+        ]
+        if not all(valid):
+            raise ValueError(
+                f'Column names do not match the schema of the DocumentArray\'s document type '
+                f'({cls.document_type.__name__}): {list(compress(field_names, [not v for v in valid]))}'
+            )
+
+    @staticmethod
+    def access_path_dict_to_nested_dict(
```
samsja, line 330: should we make is private ?

pyproject.toml
```diff
@@ -22,6 +22,7 @@ fastapi = {version = ">=0.87.0", optional = true }
 rich = ">=13.1.0"
 lz4 = {version= ">=1.0.0", optional = true}
 pydub = {version = "^0.25.1", optional = true }
+pandas = ">=1.1.0"
```
samsja, line 25: it should be an optional dependecy

samsja (commented):
look goods, I added some comments

docarray/array/array/io.py
JohannesMessner, line 330: didn't you already implement something like this in your previous PR? or am i mixing things up?

docarray/array/array/io.py
```diff
@@ -392,6 +430,59 @@ def to_csv(
                 doc_dict = _dict_to_access_paths(doc.dict())
                 writer.writerow(doc_dict)
 
+    @classmethod
+    def from_pandas(cls, df: pd.DataFrame) -> 'DocumentArray':
+        """
+        Load a DocumentArray from a `pandas.DataFrame` following the schema
+        defined in the :attr:`~docarray.DocumentArray.document_type` attribute.
+        Every row of the dataframe will be mapped to one document in the array.
```
JohannesMessner, line 438:
```suggestion
        Every row of the dataframe will be mapped to one Document in the array.
```

docarray/array/array/io.py
```diff
@@ -392,6 +430,59 @@ def to_csv(
...
+        The column names of the dataframe have to match the field namesof the
```
JohannesMessner, line 439:
```suggestion
        The column names of the dataframe have to match the field names of the
```

docarray/array/array/io.py
```diff
@@ -392,6 +430,59 @@ def to_csv(
...
+        Document type.
+        For nested fields use "__"-separated access paths as column names,
+        such as 'image__url'.
+
+        List-like fields (including field of type DocumentArray) are not supported.
+
+        :param df: pandas.DataFrame to extract Document's information from
+        :return: DocumentArray
```
JohannesMessner, line 447: maybe elaborate this a bit

docarray/array/array/io.py
```diff
@@ -302,6 +304,56 @@ def to_json(self) -> str:
         """
         return json.dumps([doc.json() for doc in self])
 
+    @classmethod
+    def _check_for_valid_document_type(cls) -> None:
```
JohannesMessner, line 308: I would prefer returning a bool and rasing at the call site. That way it is clearer where the error actually occurs, and this method can be re-used in contexts where only the information is needed, without wanting to raise an exception

docarray/array/array/io.py
```diff
@@ -302,6 +304,56 @@ def to_json(self) -> str:
...
+        if cls.document_type == AnyDocument:
+            raise TypeError(
+                'There is no document schema defined. '
+                'Please specify the DocumentArray\'s Document type using `DocumentArray[MyDoc]`.'
+            )
+
+    @classmethod
+    def _check_for_valid_access_paths(cls, field_names: Optional[List[str]]) -> None:
```
JohannesMessner, line 316: same here

docarray/array/array/io.py
```diff
@@ -302,6 +304,56 @@ def to_json(self) -> str:
         """
         return json.dumps([doc.json() for doc in self])
 
+    @classmethod
+    def _check_for_valid_document_type(cls) -> None:
+        if cls.document_type == AnyDocument:
```
JohannesMessner, line 309: is there are plan to make this check more sophisticated as part of this PR? otherwise this doesn't need to be a method imo

JohannesMessner (changes requested):

docarray/array/array/io.py
```diff
@@ -16,20 +16,21 @@
...
+    List,
     Optional,
-    Sequence,
     Tuple,
     Type,
     TypeVar,
     Union,
 )
 
+import pandas as pd
```
samsja, line 27: this should be an optional dependency, therefore we cannot import from it directly

docarray/array/array/io.py
```diff
@@ -392,6 +393,75 @@ def to_csv(
...
+        Load a DocumentArray from a `pandas.DataFrame` following the schema
+        defined in the :attr:`~docarray.DocumentArray.document_type` attribute.
+        Every row of the dataframe will be mapped to one Document in the array.
+        The column names of the dataframe have to match the field names of the
+        Document type.
+        For nested fields use "__"-separated access paths as column names,
+        such as 'image__url'.
+
+        List-like fields (including field of type DocumentArray) are not supported.
+
```
samsja, line 408: can u put a code example ? just a couple of  lines

pyproject.toml
samsja: separate. we will reorginaze at some point anyway

docarray/array/array/io.py
JohannesMessner: ok sounds good, yes keeping them in the same helper file would be ideal

samsja (changes requested):
we should make pandas install optional. Otherwise looks good

samsja (approved):

JohannesMessner (approved):

PR #1163, opened 2023-02-23
fix: fix setitiem

samsja:
# Context

This pr fix the implementation of setitiem on da stacked

### DIFF
--- a/tests/units/array/test_indexing.py
+++ b/tests/units/array/test_indexing.py
@@ -233,3 +233,18 @@ def test_boolmask_setitem(stack_left, stack_right, da, da_to_set, index):
         else:
             assert d.text == f'hello {i}'
             assert torch.all(d.embedding == torch.ones((4,)) * i)
+
+
+def test_setitiem_update_column():
+    texts = [f'hello {i}' for i in range(10)]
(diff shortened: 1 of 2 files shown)


tests/units/array/test_indexing.py
```diff
@@ -233,3 +233,18 @@ def test_boolmask_setitem(stack_left, stack_right, da, da_to_set, index):
         else:
             assert d.text == f'hello {i}'
             assert torch.all(d.embedding == torch.ones((4,)) * i)
+
+
+def test_setitiem_update_column():
```
JohannesMessner, line 238: does this test ever fail? i think we need to explicitly look at the columns in the da, otherwise it might fool us again like it fooled me

JohannesMessner (changes requested):

tests/units/array/test_indexing.py
samsja: let me take a look

samsja: should I add a test to test that the test fail ???? (kidding ofc)

samsja: Okay I updated the test, now I check inside the column as well

JohannesMessner (approved):

PR #1175, opened 2023-02-24
fix: allow relative paths in any url validation

anna-charlotte:
Goals:

Allow relative file paths while pydantic does not accept those in validation, by first checking if path is relative, if so get the absolute version, and later return the relative version of the validated version.

### DIFF
--- a/docarray/typing/url/any_url.py
+++ b/docarray/typing/url/any_url.py
@@ -47,6 +47,7 @@ class AnyUrl(BaseAnyUrl, AbstractType):
     ) -> T:
         import os
 
+        abs_path: Union[T, np.ndarray, Any]
         if (
             isinstance(value, str)
             and not value.startswith('http')


docarray/typing/url/any_url.py
```diff
@@ -35,6 +38,31 @@ def _to_node_protobuf(self) -> 'NodeProto':
...
+    def validate(
+        cls: Type[T],
+        value: Union[T, np.ndarray, Any],
+        field: 'ModelField',
+        config: 'BaseConfig',
+    ) -> T:
+        import os
+
+        try:
+            if not value.startswith('http') and not os.path.isabs(value):  # type: ignore
```
JohannesMessner, line 51: why is mypy complaining here?

docarray/typing/url/any_url.py
```diff
@@ -35,6 +38,31 @@ def _to_node_protobuf(self) -> 'NodeProto':
...
+                is_relative = True
+                value = os.path.abspath(value)  # type: ignore
```
JohannesMessner, line 53: I think for stuff like this we should to the proper thing and assign to a new variable instead of silencing mypy

docarray/typing/url/any_url.py
```diff
@@ -35,6 +38,31 @@ def _to_node_protobuf(self) -> 'NodeProto':
...
+            else:
+                is_relative = False
+        except TypeError:
+            is_relative = False
+
+        url = super().validate(value, field, config)  # basic url validation
+
+        if is_relative:
+            return cls(os.path.relpath(str(url)), scheme=None)
```
JohannesMessner, line 62: can't we just return the original input instead of casting back?

JohannesMessner (changes requested):
a lot of nit-picks for such a small PR, sorry hehe

docarray/typing/url/any_url.py
JohannesMessner: Ok makes sense. So seems like the TypeError is something that we expect during normal/correct execution of the program? That is usually a bit of a code smell, can comparatively slow etc. Can we proactively check the types to avoid this?

JohannesMessner (changes requested):

samsja (approved):

JohannesMessner (approved):
Nice!

Issue #1152, opened 2023-02-20
Dynamic class creation


jupyterjazz:
**Is your feature request related to a problem? Please describe.**
In some scenarios users might want to dynamically create a class.
For example, in the search apps, data might be given through different sources, it might have a folder structure, and we are then responsible for converting it to a suitable docarray format, where this feature will come handy.

**Describe the solution you'd like**
Having a function that takes pairs of field names and their corresponding types and returns a class

**Describe alternatives you've considered**
Tried with [Pydantic's `create_model`](https://docs.pydantic.dev/usage/models/#dynamic-model-creation), and it works ok, but this doesn't return a `BaseDocument` so would be nice to have it handled inside docarray.

PR #1179, opened 2023-02-27
feat: dynamic class creation

AnneYang720:
Goals:

This PR is for issue #1152 

In the issue, [Pydantic's create_model](https://docs.pydantic.dev/usage/models/#dynamic-model-creation) was mentioned. In order for the function to return a subclass of `BaseDocument`, we need to pass the parameter `__base__`. 

- [x] code
- [x] tests

### DIFF
--- a/docarray/documents/helper.py
+++ b/docarray/documents/helper.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Type, Union
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Type, TypeVar
 
 from pydantic import create_model, create_model_from_typeddict
 from pydantic.config import BaseConfig
@@ -9,20 +9,20 @@ from docarray import BaseDocument
 if TYPE_CHECKING:
     from pydantic.typing import AnyClassMethod
 
+    Document = TypeVar('Document', bound=BaseDocument)
+
 
 def create_doc(
     __model_name: str,
     *,
     __config__: Optional[Type[BaseConfig]] = None,
-    __base__: Union[
-        None, Type['BaseDocument'], Tuple[Type['BaseDocument'], ...]
-    ] = BaseDocument,
+    __base__: Type['Document'] = BaseDocument,  # type: ignore
     __module__: str = __name__,
     __validators__: Dict[str, 'AnyClassMethod'] = None,  # type: ignore
     __cls_kwargs__: Dict[str, Any] = None,  # type: ignore
     __slots__: Optional[Tuple[str, ...]] = None,
     **field_definitions: Any,
-):
+) -> Type['Document']:
     """
     Dynamically create a subclass of BaseDocument. This is a wrapper around pydantic's create_model.
     :param __model_name: name of the created model
@@ -56,7 +56,7 @@ def create_doc(
 
     """
 
-    if not issubclass(__base__, BaseDocument):  # type: ignore
+    if not issubclass(__base__, BaseDocument):
         raise ValueError(f'{type(__base__)} is not a BaseDocument or its subclass')
 
     doc = create_model(
--- a/docarray/documents/helper.py
+++ b/docarray/documents/helper.py
@@ -1,18 +1,76 @@
-from typing import Any, Type
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Type, TypeVar, Union
 
-from pydantic import create_model
+from pydantic import create_model, create_model_from_typeddict
+from pydantic.config import BaseConfig
+from typing_extensions import TypedDict
 
 from docarray import BaseDocument
 
+if TYPE_CHECKING:
+    from pydantic.typing import AnyClassMethod
 
-def create_from_dict(model_name: str, **field_definitions: Any) -> Type[BaseDocument]:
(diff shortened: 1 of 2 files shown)


tests/integrations/document/test_document.py
```diff
@@ -32,3 +33,22 @@ class ChunksDocument(BaseDocument):
     )
 
     assert isinstance(doc.images, DocumentArray)
+
+
+def test_create_from_doct():
```
JoanFM, line 38:
```suggestion
def test_create_from_dict():
```

JoanFM (dismissed):

docarray/documents/helper.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any, Type
+
+from pydantic import create_model
+
+from docarray import BaseDocument
+
+
+def create_from_dict(model_name: str, **field_definitions: Any) -> Type[BaseDocument]:
+    """
+    Dynamically create a subclass of BaseDocument from a dict of field definitions.
```
JohannesMessner, line 10: Let's mention that this is very similar to `create_model` in pydantic

JohannesMessner (changes requested):
I think we should split this out into two functions:
- one that mirrors `create_model` as closely as possible, has a similar name, takes all the same arguments, and just makes sure that the `__base__` is set correctly. It should also take the fields in the same way pydantic does
- another, slightly higher level one, that would be called `creat_from_dict`, and take the fields as a dictionary. This would also closely mirror functionality in pydantic: https://docs.pydantic.dev/usage/models/#model-creation-from-namedtuple-or-typeddict

docarray/documents/helper.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any, Type
+
+from pydantic import create_model
+
+from docarray import BaseDocument
+
+
+def create_from_dict(model_name: str, **field_definitions: Any) -> Type[BaseDocument]:
```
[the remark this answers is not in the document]
samsja: I would call it `create_doc` similar to `create_model`

docarray/documents/helper.py
```diff
@@ -0,0 +1,74 @@
...
+        None, Type['BaseDocument'], Tuple[Type['BaseDocument'], ...]
+    ] = BaseDocument,
+    __module__: str = __name__,
+    __validators__: Dict[str, 'AnyClassMethod'] = None,  # type: ignore
+    __cls_kwargs__: Dict[str, Any] = None,  # type: ignore
+    __slots__: Optional[Tuple[str, ...]] = None,
+    **field_definitions: Any,
+):
+    """
+    Dynamically create a subclass of BaseDocument. This is a wrapper around pydantic's create_model.
```
samsja, line 27: can you put an python example in the docstring ?

docarray/documents/helper.py
```diff
@@ -0,0 +1,74 @@
...
+
+    return doc
+
+
+def create_from_typeddict(
+    typeddict_cls: Type['TypedDict'],  # type: ignore
+    **kwargs: Any,
+):
+    """
+    Create a subclass of BaseDocument based on the fields of a `TypedDict`. This is a wrapper around pydantic's create_model_from_typeddict.
```
samsja, line 61: same here can u put a python example ?

docarray/documents/helper.py
```diff
@@ -0,0 +1,120 @@
...
+    __config__: Optional[Type[BaseConfig]] = None,
+    __base__: Union[
+        None, Type['BaseDocument'], Tuple[Type['BaseDocument'], ...]
+    ] = BaseDocument,
+    __module__: str = __name__,
+    __validators__: Dict[str, 'AnyClassMethod'] = None,  # type: ignore
+    __cls_kwargs__: Dict[str, Any] = None,  # type: ignore
+    __slots__: Optional[Tuple[str, ...]] = None,
+    **field_definitions: Any,
+):
```
samsja, line 25: would be nice to add the return type.

I suggested introducing a TypeVar that you can use in `__base__` and use in the return statement

docarray/documents/helper.py
```diff
@@ -0,0 +1,120 @@
...
+            title=(str, ...),
+            tensor=(AudioNdArray, ...),
+        )
+
+        assert issubclass(MyAudio, BaseDocument)
+        assert issubclass(MyAudio, Audio)
+
+    """
+
+    if not issubclass(__base__, BaseDocument):  # type: ignore
```
samsja, line 59: why type ignore here ?

samsja (changes requested):
good job ! I added a small requests

docarray/documents/helper.py
AnneYang720: okay I removed the return type because of mypy error.
I'll try to find a way for both adding the return and pass mypy.

docarray/documents/helper.py
```diff
@@ -9,20 +9,20 @@
 if TYPE_CHECKING:
     from pydantic.typing import AnyClassMethod
 
+    Document = TypeVar('Document', bound=BaseDocument)
```
samsja, line 12:
```suggestion
    T_doc = TypeVar('T_doc', bound=BaseDocument)
```

samsja: I think it is common practice to name them `T_smth`

samsja (approved):

Issue #1134, opened 2023-02-14
Implement push/pull interface that uses protocol to push and pull with streaming from s3 or file.


JoanFM:
**Is your feature request related to a problem? Please describe.**
Implement something like this:

```python
da.push(url='s3 or file or jinaai')
da.pull(url=..., streaming=True)
```

samsja:
implemented in https://github.com/docarray/docarray/pull/1182

PR #1182, opened 2023-02-27
feat: implement push/pull interface from JAC, file and s3

Jackmin801:
## Goals
1. Allow users to push and pull DAs into their programs from local filesystem, S3 or Jina AI Cloud
2. Allow users to stream Documents such that only part of the document is in memory at any one time
3. Do the streaming fast enough that it is worth training models with just pure streaming (no storing dataset to disk)

## Basic Usage
Push and pull from File
```python
N: int = 2 ** 20
DocumentArray[TextDoc].push_stream((TextDoc(text=f'text {i}') for i in range(N)), url=f'file://./data/da-{N}', show_progress=True)
da = DocumentArray[TextDoc].pull(url=f'jinaai://./data/da-{N}', show_progress=True)
print(da)
```

Push and pull from S3
```python
N: int = 2 ** 20
DocumentArray[TextDoc].push_stream((TextDoc(text=f'text {i}') for i in range(N)), url=f's3://bucket/path/da-{N}', show_progress=True)
da = DocumentArray[TextDoc].pull(url=f's3://bucket/path/da-{N}', show_progress=True)
print(da)
```

Push and pull from JAC
```python
N: int = 2 ** 20
DocumentArray[TextDoc].push_stream((TextDoc(text=f'text {i}') for i in range(N)), url=f'jinaai://da-{N}', show_progress=True)
da = DocumentArray[TextDoc].pull(url=f'jinaai://da-{N}', show_progress=True)
print(da)
```

## Streaming

Streaming should allow users to push DAs from a generator. This allows users to use significantly less memory in certain use cases.
```python
import time
import tracemalloc
from docarray.documents import Text

N: int = 2 ** 12

# Warmup so the comparison is fair
da = DocumentArray[Text](Text(text=f'text {i}') for i in range(N))
da.push(url='file://da1', show_progress=False)
DocumentArray[Text].push_stream((Text(text=f'text {i}') for i in range(N)), url='file://da1', show_progress=False)

# Without streaming
start_time = time.time()
tracemalloc.start()

da = DocumentArray[Text](Text(text=f'text {i}') for i in range(N))
da.push(url='file://da1', show_progress=False)

current, peak = tracemalloc.get_traced_memory()
tracemalloc.stop()
print(f'Push (without stream) memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB')
print(f'Push (without stream) took {time.time() - start_time:.4f} seconds')

# With streaming
start_time = time.time()
tracemalloc.start()

DocumentArray[Text].push_stream((Text(text=f'text {i}') for i in range(N)), url='file://da1', show_progress=False)

current, peak = tracemalloc.get_traced_memory()
tracemalloc.stop()
print(f'Push stream memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB')
print(f'Push stream took {time.time() - start_time:.4f} seconds')
```
```console
Push (without stream) memory usage is 2.930839MB; Peak was 3.238053MB
Push (without stream) took 0.2543 seconds
Push stream memory usage is 0.001392MB; Peak was 0.309381MB
Push stream took 0.2494 seconds
```

Users should also be able to stream down DAs. This would make the memory usage of reduce operations to be independent of the size of the DA.
```python
import time
import tracemalloc
from docarray.documents import Text

# Warmup so the comparison is fair
sum(len(d.text) for d in DocumentArray[Text].pull_stream(url='file://da1', show_progress=False))

# Without streaming
start_time = time.time()
tracemalloc.start()

total_chars = sum(len(d.text) for d in DocumentArray[Text].pull_stream(url='file://da-1024', show_progress=False))

current, peak = tracemalloc.get_traced_memory()
tracemalloc.stop()
print(f'Resulting total chars is {total_chars}')
print(f'Pull 1024 memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB')
print(f'Pull 1024 took {time.time() - start_time:.4f} seconds')

# With streaming
start_time = time.time()
tracemalloc.start()

total_chars = sum(len(d.text) for d in DocumentArray[Text].pull_stream(url='file://da-4096', show_progress=False))

current, peak = tracemalloc.get_traced_memory()
tracemalloc.stop()
print(f'Resulting total chars is {total_chars}')
print(f'Pull 4096 memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB')
print(f'Pull 4096 took {time.time() - start_time:.4f} seconds')
```
```console
Resulting total chars is 8106
Pull 1024 memory usage is 0.001209MB; Peak was 0.090019MB
Pull 1024 took 0.0636 seconds
Resulting total chars is 35754
Pull 4096 memory usage is 0.001113MB; Peak was 0.089923MB
Pull 4096 took 0.2435 seconds
```

## Progress:
- Implements #1134
- [x] JAC
~JAC Streaming~
- [x] File
- [x] File Streaming
- [x] File concurrency (eventual consistency, first push wins)
- [x] S3
- [x] S3 Streaming
- [x] S3 Caching
- [x] S3 Concurrency
- [x] Progress bars
- [x] Tables
- [x] Add ticket about S3 multipart upload https://github.com/docarray/docarray/issues/1230


JohannesMessner:
Could you update the PR description with some example usage and API? That would make it easier to review

docarray/array/array/io.py
```diff
@@ -177,37 +230,53 @@ def _write_bytes(
...
+            'Serializing', disable=not show_progress, total=len(self)
+        )
+
+        yield self._stream_header
+
+        with pbar:
+            _total_size = 0
+            pbar.start_task(t)
+            for doc in self:
+                doc_bytes = doc.to_bytes(protocol=protocol, compress=compress)
```
girishc13, line 267: This is the main bottleneck when deserializing s very large documents. Does this support streaming option?

JoanFM: a single Document itself is a bottleneck?

girishc13: If its large enough and the docarray has a lot of documents then most of the doc bytes will stay in memory. Anyway its up to the user to decide how to load a huge docarray. I still think that we need to test the new version with the laion400m dataset available in the jina hub. Nobody investigated this [issue](https://github.com/docarray/docarray/issues/1121).

girishc13:
@Jackmin801 Please think about adding debug and/or trace logging otherwise it become very difficult to debug issues. Having a dedicate module based logging will be very helpful.

docarray/array/array/pushpull/s3.py
```diff
@@ -0,0 +1,117 @@
...
+        return da_names
+
+    @staticmethod
+    def delete(name: str, missing_ok: bool = True) -> bool:
+        bucket, name = name.split('/', 1)
+        s3 = boto3.resource('s3')
+        object = s3.Object(bucket, name + '.da')
+        try:
+            object.load()
+        except botocore.exceptions.ClientError as e:
```
girishc13, line 42: Why is there additional check for 404 here? The boto client already handles this case so why the duplication?

Jackmin801: Because the response from [object.delete()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/object/load.html) does not indicate if the delete was successful. It returns 204 regardless of whether or not the object we wished to delete existed in the first place.

The [object.load()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/object/load.html) here actually calls [head_object()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/head_object.html) which only retrieves metadata. So it shouldn't be that costly. We aren't paying double cost here.

docarray/array/array/pushpull/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+    from docarray import BaseDocument, DocumentArray
+
+
+class ConcurrentPushException(Exception):
+    """Exception raised when a concurrent push is detected."""
+
+    pass
+
+
+class PushPullLike(Protocol):
```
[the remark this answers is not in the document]
samsja: well actually I am not sure if docstring in protocol makes sense :thinkinf_face:

docarray/array/array/pushpull/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+        elif protocol == 'file':
+            from docarray.array.array.pushpull.file import PushPullFile
+
+            cls.__backends__[protocol] = PushPullFile
+            logging.info('Loaded Local Filesystem backend')
+        elif protocol == 's3':
+            from docarray.array.array.pushpull.s3 import PushPullS3
+
+            cls.__backends__[protocol] = PushPullS3
+            logging.info('Loaded S3 backend')
```
samsja, line 103: info or debug for the logging ?

docarray/array/array/pushpull/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+
+    __backends__: Dict[str, PushPullLike] = {}
+    document_type: Type['BaseDocument']
+
+    @abstractmethod
+    def __len__(self) -> int:
+        ...
+
+    @classmethod
+    def get_backend(cls, protocol: str) -> PushPullLike:
```
samsja, line 79: Maybe we should use Literal here ? https://docs.python.org/3/library/typing.html#typing.Literal


```suggestion
    def get_backend(cls, protocol: Literal['jnaai', 's3', 'file']) -> PushPullLike:
```

docarray/array/array/pushpull/__init__.py
```diff
@@ -0,0 +1,244 @@
+import logging
+from abc import abstractmethod
+from pathlib import Path
+from typing import TYPE_CHECKING, Dict, Iterable, Iterator, List, Optional, Type
+
+from typing_extensions import Protocol
+
+__cache_path__ = Path.home() / '.cache' / 'docarray-v2'
```
samsja, line 8: should rather be `docarray` no ?

docarray/array/array/pushpull/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+
+        If no url is provided, the DocumentArrays in the local cache will be listed.
+
+        :param url: should be of the form ``protocol://namespace``. e.g. ``s3://bucket/path/to/namespace``, ``file:///path/to/folder``
+        :param show_table: whether to show the table of artifacts
+        :return: a list of artifact names
+        """
+        logging.info(f'Listing artifacts from {url}')
+        protocol, namespace = url.split('://', 2)
+        return PushPullMixin.get_backend(protocol).list(namespace, show_table)
```
samsja, line 125: I would make it a class method here rather that staticmethod that called itself. This would make it easier if someone want to extend the behavior of the mixin without copy pasting code

docarray/array/array/pushpull/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+        """Push this DocumentArray object to the specified url.
+
+        :param url: url specifying the protocol and save name of the DocumentArray. Should be of the form ``protocol://namespace/name``. e.g. ``s3://bucket/path/to/namespace/name``, ``file:///path/to/folder/name``
+        :param public:  Only used by ``jinaai`` protocol. If true, anyone can pull a DocumentArray if they know its name.
+            Setting this to false will restrict access to only the creator.
+        :param show_progress: If true, a progress bar will be displayed.
+        :param branding: Only used by ``jinaai`` protocol. A dictionary of branding information to be sent to Jina AI Cloud. {"icon": "emoji", "background": "#fff"}
+        """
+        logging.info(f'Pushing {len(self)} docs to {url}')
+        protocol, name = url.split('://', 2)
```
samsja, line 162: Can we put this `url.split('://', 2)` into a static function ? I saw that we are using it a couple of times

docarray/array/array/pushpull/__init__.py
samsja: just an idea, should not we make this `__chache_path__` a function ? WHat happened if I change the my os path during runtime ? should the cache folder change as well ?

docarray/array/array/pushpull/file.py
```diff
@@ -0,0 +1,182 @@
...
+                try:
+                    f.write(next(source))
+                except StopIteration:
+                    break
+        path.rename(path.with_suffix(''))
+        return {}
+
+    @staticmethod
+    def pull(
+        cls: Type['DocumentArray'],
```
samsja, line 137: it is missleading to call this one `cls` I thought it was a classmethod. Maybe `da_cls` ?

docarray/array/array/pushpull/file.py
```diff
@@ -0,0 +1,182 @@
...
+
+        return cls(
+            PushPullFile.pull_stream(
+                cls, name, show_progress=show_progress, local_cache=local_cache
+            )
+        )
+
+    @staticmethod
+    def pull_stream(
+        cls: Type['DocumentArray'],
```
samsja, line 158: same here

docarray/array/array/pushpull/jinaai.py
```diff
@@ -0,0 +1,345 @@
...
+            name='Multimodal dataclass',
+            value=True,
+            description='Whether all documents are multimodal',
+        ),
+    ]
+
+    return items
+
+
+class PushPullJAC(PushPullLike):
```
samsja, line 64: why do u need to inherit from the protocol here?

docarray/array/array/pushpull/__init__.py
```diff
@@ -0,0 +1,244 @@
+import logging
```
samsja, line 1: can we move the implementation from the __init__.py ? I am not a big fan

docarray/array/array/array.py
```diff
@@ -57,7 +58,7 @@ def _delegate_meth(self, *args, **kwargs):
 
 
 class DocumentArray(
-    IndexingSequenceMixin[T_doc], IOMixinArray, AnyDocumentArray[T_doc]
+    IndexingSequenceMixin[T_doc], PushPullMixin, IOMixinArray, AnyDocumentArray[T_doc]
 ):
```
samsja, line 62: Not sure about making Da directly a PushPullMixin. I think this quite polute the namespace of DocumentArray :thinking_face: At the same time I don't have an alternative in mind

tests/integrations/remotes/docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: "3"
+services:
+  minio:
+    container_name: minio
+    image: minio/minio
```
samsja, line 5: maybe we need a tag to make the CI reliable ?

pyproject.toml
```diff
@@ -51,6 +51,14 @@ uvicorn = ">=0.19.0"
 httpx = ">=0.23.0"
 pytest-asyncio = ">=0.20.2"
 
+
+[tool.poetry.group.jina.dependencies]
+jina-hubble-sdk = "^0.34.0"
```
samsja, line 56:
```suggestion
jina-hubble-sdk = ">=0.34.0"
```

samsja: we should make this optional

samsja (changes requested):
Great PR: Looking forward to merge it.

I added several comments

docarray/array/array/pushpull/jinaai.py
Jackmin801: Yea you are right. This isn't needed. I didn't know this until later and accidentally left this here.

docarray/array/array/pushpull/__init__.py
Jackmin801: Ah ok sure

docarray/array/array/pushpull/__init__.py
Jackmin801: Hrmm ok. I figured it was quite short and not too hard to understand. But the literal typing requires this now so its done.

docarray/array/array/pushpull/__init__.py
Jackmin801: Yea. I think with a function we can capture better resolution properties. I added the option to change the cache path with env variables into the function.

Hrmmm, changing the path at runtime is a bit weird to me so I disabled it by making it only evaluate once. I think if users want to change cache path at runtime, they should do it in their own code and not count on our DOCARRAY_CACHE env variable.

We can revert [1efc33a](https://github.com/docarray/docarray/pull/1182/commits/1efc33abf5b8e7d669d98807b228ad6567156e53) if we want it to always look at the env variable.

docarray/array/array/pushpull/__init__.py
Jackmin801: Now that you've pointed it out, yea this makes more sense as debug

tests/integrations/remotes/docker-compose.yml
Jackmin801: Oh yes definitely. Pinning tags is a good idea

pyproject.toml
Jackmin801: oops. doone!

docarray/array/array/pushpull/__init__.py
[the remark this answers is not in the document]
Jackmin801: Yea. In hindsight. Maybe this isnt the best use case of structural subtyping. I changed the implementation to use a an ABC instead.

Oh and yes. With docstring : D

docarray/array/array/pushpull.py
```diff
@@ -0,0 +1,180 @@
...
+            from docarray.store.s3 import S3DocStore
+
+            cls.__backends__[protocol] = S3DocStore
+            logging.debug('Loaded S3 backend')
+        else:
+            raise NotImplementedError(f'protocol {protocol} not supported')
+
+        return cls.__backends__[protocol]
+
+    def push(
```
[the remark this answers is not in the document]
Jackmin801: Having it at the document array level is less verbose though. Users can still do it the long -- safer way if they want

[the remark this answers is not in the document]
samsja: I see

samsja (approved):

PR #1183, opened 2023-02-27
refactor: da stack full column wise

samsja:
# Context

Make `DocumentArrayStack` a full column base data structure and introduce the Document that are just a view.

PR is still in progress

## Early design

* introducing the concept of `Storage` that holds the data of the `DocumentArrayStack`
* introducing the concept of `StorageView`. It looks and feels like a `dict` but is actually a view `Storage`. It has only a reference to `Storage` as well as an index value
* Document can now be a "view"  of a `DocumentArrayStack` by holding a `StorageView`. 
* `DocumentArrayStack` does not hold any reference to any `Document` (no cyclic reference). Each call to `da[0]` will produce a different `Document` that points to the same data. This behavior is the same than for Pytorch


docarray/array/stacked/storage.py
```diff
@@ -0,0 +1,136 @@
...
+tf_available = is_tf_available()
+if tf_available:
+    import tensorflow as tf  # type: ignore
+
+    from docarray.typing import TensorFlowTensor
+else:
+    TensorFlowTensor = None  # type: ignore
+
+
+class Storage:
```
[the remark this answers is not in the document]
samsja: Columns ?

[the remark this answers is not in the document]
samsja: DatStorage?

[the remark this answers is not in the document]
JoanFM: Storage may conflict with concepts worked on by @Jackmin801 on push/pull

docarray/array/stacked/storage.py
```diff
@@ -0,0 +1,136 @@
...
+
+    tensor_storage: Dict[str, AbstractTensor]
+    document_storage: Dict[str, 'DocumentArrayStacked']
+    da_storage: Dict[str, List['DocumentArrayStacked']]
+    any_storage: Dict[str, List]
+
+    def __init__(
+        self,
+        docs: Sequence[BaseDocument],
+        document_type: Type[BaseDocument],
```
JoanFM, line 34: I guess these would later go into the `[][]` annotations right?

docarray/array/stacked/column_storage.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class ColumnStorage:
+
+    document_type: Type[BaseDocument]
+
+    tensor_columns: Dict[str, AbstractTensor]
+    doc_columns: Dict[str, 'DocumentArrayStacked']
+    da_columns: Dict[str, List['DocumentArrayStacked']]
+    any_columns: Dict[str, List]
```
JohannesMessner, line 46: why is `da_columns` separate? It's just a list anyways, right?

samsja: We need special treatment for the da inside da. So gut feeling is that we should make it a different variable

docarray/array/array/array.py
```diff
@@ -136,10 +135,27 @@ def __init__(
...
 
+    @classmethod
+    def construct(
+        cls: Type[T],
+        docs: Sequence[T_doc],
+        tensor_type: Type['AbstractTensor'] = NdArray,
+    ) -> T:
+        """
+        Create a DocumentArray without validation any data. The data must come from a
+        trusted sources
```
JohannesMessner, line 149:
```suggestion
        trusted source
```

docarray/array/stacked/array_stacked.py
```diff
@@ -70,205 +67,67 @@ class DocumentArrayStacked(AnyDocumentArray[T_doc]):
...
+        :param storage: the underlying storage.
+        :return: a DocumentArrayStack
+        """
+        da = cls.__new__(cls)
+        da.tensor_type = storage.tensor_type
+        da._storage = storage
+        return da
+
+    @classmethod
+    def validate(
```
JohannesMessner, line 95: What is this used for? Isn't this usually a pydantic thing?

docarray/array/stacked/array_stacked.py
```diff
@@ -366,108 +205,151 @@ def _set_data_and_columns(
...
+            # TODO complex indexing will not work here
+
+    def _set_array_attribute(
+        self: T,
+        field: str,
+        values: Union[List, T, DocumentArray, AbstractTensor],
+    ) -> None:
+        """Set all Documents in this DocumentArray using the passed values
+
+        :param field: name of the fields to extract
```
JohannesMessner, line 238: the value to `set` instead of `extract`?

tests/units/array/stack/storage/test_storage.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+
+from docarray import BaseDocument
+from docarray.array import DocumentArrayStacked
+from docarray.array.stacked.column_storage import ColumnStorage, ColumnStorageView
+from docarray.typing import AnyTensor, NdArray
+
+
+def test_storage_init():
```
JoanFM, line 9:
```suggestion
def test_column_storage_init():
```

docarray/array/stacked/column_storage.py
```diff
@@ -0,0 +1,196 @@
...
+            tensor_columns,
+            doc_columns,
+            da_columns,
+            any_columns,
+            self.document_type,
+            self.tensor_type,
+        )
+
+
+class ColumnStorageView(dict, MutableMapping[str, Any]):
```
JohannesMessner, line 174: is this double inheritance needed? isn't dict already a mutable mapping?

tests/units/array/stack/storage/test_storage.py
```diff
@@ -0,0 +1,54 @@
...
+
+    assert (storage.tensor_columns['tensor'] == np.zeros((4, 10))).all()
+    assert storage.any_columns['name'] == ['hello' for _ in range(4)]
+    inner_docs = storage.doc_columns['doc']
+    assert isinstance(inner_docs, DocumentArrayStacked[InnerDoc])
+    # for i, doc in enumerate(inner_docs):
+    #     assert doc.price == i
+
+
+def test_storage_view():
```
JoanFM, line 33:
```suggestion
def test_column_storage_view():
```

JoanFM (commented):
I think is going in the right direction

docarray/array/stacked/column_storage.py
JohannesMessner: Also, docstring plz :)

docarray/array/array/array.py
```diff
@@ -136,10 +135,27 @@ def __init__(
...
         self.tensor_type = tensor_type
 
+    @classmethod
+    def construct(
+        cls: Type[T],
+        docs: Sequence[T_doc],
+        tensor_type: Type['AbstractTensor'] = NdArray,
+    ) -> T:
+        """
+        Create a DocumentArray without validation any data. The data must come from a
```
JohannesMessner, line 148: When would validation happen? In the `parse_obj_as`? So are we extendind pydantic-style validation to non-pydantic classes?

JohannesMessner (commented):
Overall structure makes a lot of sense to me, especially the `ColumnStorageView` and setting that as the `__dict__`

docarray/array/stacked/array_stacked.py
samsja: Yes but we need it actually if I want to do 
```python
class MyDoc(BaseDocument):
   images: DocumetArrayStacked[Image]
```

docarray/array/stacked/column_storage.py
samsja: docstring coming everywhere !

samsja: You cannot do 
```python
class MyClass(dict[str, Any]):
   ...
```

and if you do only 
```
class MyClass(MutableMapping[str, Any]):
   ...
```

then you don't have dict features.


So if you want both type hint and dict behavior you need both I think ? Maybe using `Dict[str,Any]` could work.

docarray/array/array/sequence_indexing_mixin.py
```diff
@@ -0,0 +1,216 @@
...
+        try:
+            return ndim == 0 and np.issubdtype(dtype, np.integer)
+        except TypeError:
+            return False
+    return False  # this is unreachable, but mypy wants it
+
+
+class IndexingSequenceMixin(Iterable[T_item]):
+    """
+    This mixin allow to extend a list into an object that can be indexed
```
JohannesMessner, line 38:
```suggestion
    This mixin allow sto extend a list into an object that can be indexed
```

docarray/array/array/sequence_indexing_mixin.py
```diff
@@ -0,0 +1,216 @@
...
+
+    You can index into a IndexingSequenceMixin like a numpy array or torch tensor:
+
+    .. code-block:: python
+        da[0]  # index by position
+        da[0:5:2]  # index by slice
+        da[[0, 2, 3]]  # index by list of indices
+        da[True, False, True, True, ...]  # index by boolean mask
+
+    You can delete items from a DocumentArray like a Python List
```
JohannesMessner, line 49: So no deletion with torch-like indexing? Is there a technical limitation that requires this? I can see that being surprising/confusing

docarray/array/array/sequence_indexing_mixin.py
```diff
@@ -0,0 +1,216 @@
...
+            results.append(self._data[ix])
+        return self.__class__(results)
+
+    def _set_by_indices(self: T, item: Iterable[int], value: Iterable[T_item]):
+        # here we cannot use _get_offset_to_doc() because we need to change the doc
+        # that a given offset points to, not just retrieve it.
+        # Future optimization idea: _data could be List[DocContainer], where
+        # DocContainer points to the doc. Then we could use _get_offset_to_container()
+        # to swap the doc in the container.
+        for ix, doc_to_set in zip(item, value):
```
JohannesMessner, line 118: I think the spirit of this comment still makes sense and is true, but the reference to `_get_offset_to_doc()` is outdated

docarray/array/array/sequence_indexing_mixin.py
```diff
@@ -0,0 +1,216 @@
...
+                self._data[i] = value[i_value]
+                i_value += 1
+
+    def _del_from_mask(self: T, item: Iterable[bool]) -> None:
+        idx_to_delete = [i for i, val in enumerate(item) if val]
+        self._del_from_indices(idx_to_delete)
+
+    def _del_from_indices(self: T, item: Iterable[int]) -> None:
+        for ix in sorted(item, reverse=True):
+            # reversed in needed here otherwise some the indice are not up to date after
```
JohannesMessner, line 142:
```suggestion
            # reversed is needed here otherwise some the indices are not up to date after
```

docarray/array/array/sequence_indexing_mixin.py
```diff
@@ -0,0 +1,216 @@
...
+
+        if item is None:
+            return
+        elif isinstance(item, (int, slice)):
+            del self._data[item]
+        else:
+            head = item[0]  # type: ignore
+            if isinstance(head, bool):
+                item_ = cast(Iterable[bool], item)
+                return self._del_from_mask(item_)
```
JohannesMessner, line 157: Here it looks like can actually do the more advanced stuff for deletion as well. So is the docstring outdated?

docarray/array/array/array.py
samsja: > When would validation happen? In the parse_obj_as? So are we extendind pydantic-style validation to non-pydantic classes?

yes we expand pydantic validation to array

docarray/array/stacked/array_stacked.py
```diff
@@ -56,125 +58,80 @@ class DocumentArrayStacked(AnyDocumentArray[T_doc]):
...
+    creating the tensor column). If the field is a `BasedDocument` as well
+    (nested document) then the column will be another DocumentArrayStack that follow the
+    schema of the nested Document. If the field is a `DocumentArray` or
+    `DocumentArrayStacked` then the column will be a list of `DocumentArrayStacked`.
+    For any other type the column is just a Python List.
+
+    `Document`s inside the `DocumentArrayStack` are just row view. The `Document`s  does
+     not hold any data anymore but only reference to the columns. The behavior of
+     this Document "view" is similar to the behavior of `view = batch[i]` in
+     Numpy/Pytorch.
```
JohannesMessner, line 80:
```suggestion
    A DocumentArrayStacked has a similar interface as
    {class}`~docarray.array.DocumentArray` but with an underlying implementation that is
    column based instead of row based. Each field
    of the schema of the DocumentArrayStack
    (the :attr:`~docarray.array.stacked.DocumentArrayStacked.document_type` which is a
    `BaseDocument`) will be stored in a column. If the field is a tensor, the data from all Documents will be stored as a single, stacked (torch/np/tf) tensor.
    If the tensor field
    is `AnyTensor` or a Union of tensor types, the
    :attr:`~docarray.array.stacked.DocumentArrayStacked.tensor_type` will be used to determine
    the type of the stacked column.
    
    If the field is another `BasedDocument` the column will be another DocumentArrayStacked that follows the
    schema of the nested Document.
    If the field is a `DocumentArray` or
    `DocumentArrayStacked` then the column will be a list of `DocumentArrayStacked`.
    For any other type the column is a Python list.

    Every `Document` inside a `DocumentArrayStacked` is a view into the data columns stored at the `DocumentArrayStacked` level. The `Document`  does
     not hold any data itself. The behavior of
     this Document "view" is similar to the behavior of `view = tensor[i]` in
     numpy/PyTorch.

```

docarray/array/array/sequence_indexing_mixin.py
samsja: Well I copy paste your implementation and your docstring :sweat_smile:

samsja: https://github.com/docarray/docarray/blob/76ecf8c62768b134538e660ce36bb2bb8ec5e28f/docarray/array/array/array.py#L115

docarray/array/stacked/array_stacked.py
```diff
@@ -56,125 +58,80 @@ class DocumentArrayStacked(AnyDocumentArray[T_doc]):
...
 
     :param docs: a DocumentArray
     :param tensor_type: Class used to wrap the stacked tensors
 
     """
 
-    document_type: Type[BaseDocument] = AnyDocument
-    _docs: DocumentArray
+    document_type: Type[T_doc]
+    _storage: ColumnStorage
```
JohannesMessner, line 87: why is the storage a class var?

docarray/array/array/sequence_indexing_mixin.py
samsja: actually I don't understand this comment

docarray/array/stacked/array_stacked.py
```diff
@@ -56,125 +58,80 @@ class DocumentArrayStacked(AnyDocumentArray[T_doc]):
...
         tensor_type: Type['AbstractTensor'] = NdArray,
     ):
-        self._doc_columns: Dict[str, 'DocumentArrayStacked'] = {}
-        self._tensor_columns: Dict[str, AbstractTensor] = {}
         self.tensor_type = tensor_type
 
-        self.from_iterable_document(docs)
+        tensor_columns: Dict[str, AbstractTensor] = dict()
+        doc_columns: Dict[str, 'DocumentArrayStacked'] = dict()
+        da_columns: Dict[str, ListAdvanceIndex['DocumentArrayStacked']] = dict()
```
JohannesMessner, line 98: Not sure where this is defined, but it should be `ListAdvancedIndexing` with a `d` :)

docarray/array/array/sequence_indexing_mixin.py
samsja: we never had such a `_get_offset_to_doc` method

docarray/array/array/sequence_indexing_mixin.py
samsja: @JohannesMessner what should I add for this "torch-like" indexing

docarray/array/stacked/array_stacked.py
```diff
@@ -56,125 +58,80 @@ class DocumentArrayStacked(AnyDocumentArray[T_doc]):
...
-
-        if len(docs) == 0:
-            return {}, {}
-
-        doc_columns: Dict[str, DocumentArrayStacked] = dict()
-        tensor_columns: Dict[str, AbstractTensor] = dict()
-
-        for field_name, field in cls.document_type.__fields__.items():
-            field_type = field.outer_type_
+        for field_name, field in self.document_type.__fields__.items():
```
JohannesMessner, line 109: Maybe a short comment about what we are doing in this loop?

docarray/array/stacked/array_stacked.py
```diff
@@ -196,79 +153,80 @@ def _create_columns(
...
-        :return: Returns a list of the field value for each document
-        in the array like container
+    @classmethod
+    def from_columns_storage(cls: Type[T], storage: ColumnStorage) -> T:
         """
-        if field in self._doc_columns.keys():
-            return self._doc_columns[field]
-        elif field in self._tensor_columns.keys():
-            return self._tensor_columns[field]
+        Create a DocumentArrayStack directly from a storage object
```
JohannesMessner, line 182:
```suggestion
        Create a DocumentArrayStacked directly from a storage object
```

docarray/array/stacked/array_stacked.py
JohannesMessner: Ok. So we copy the pydantic validation interface, but acutally just call this method manually, right?

docarray/array/stacked/array_stacked.py
```diff
@@ -278,84 +236,65 @@ def __getitem__(self: T, item: int) -> T_doc:
...
+        return self.document_type.from_view(ColumnStorageView(item, self._storage))
 
-    @overload
-    def __setitem__(self: T, key: IndexIterType, value: T):
-        ...
+    def _get_array_attribute(
+        self: T,
+        field: str,
+    ) -> Union[List, 'DocumentArrayStacked', AbstractTensor]:
+        """Return all values of the fields from all docs this array contains
```
JohannesMessner, line 252: I'm not sure I undrstand this sentence. Isn't it equivalend to "return one column of the data"? If so, I would also rename the method to that effect

docarray/array/stacked/list_advance_indexing.py
```diff
@@ -0,0 +1,42 @@
+from typing import Iterator, MutableSequence, TypeVar
+
+from docarray.array.array.sequence_indexing_mixin import IndexingSequenceMixin
+
+T_item = TypeVar('T_item')
+
+
+class ListAdvanceIndex(IndexingSequenceMixin[T_item]):
```
JohannesMessner, line 8:
```suggestion
class ListAdvancedIndexing(IndexingSequenceMixin[T_item]):
```
The `d` is needed I think, on the `ing` I could be convinced otherwise ;)

docarray/array/stacked/list_advance_indexing.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    A list wrapper that implement custom indexing
```
JohannesMessner, line 10:
```suggestion
    A list wrapper that implements custom indexing
```

docarray/array/stacked/list_advance_indexing.py
```diff
@@ -0,0 +1,42 @@
...
+
+    You can index into a ListAdvanceIndex like a numpy array or torch tensor:
+
+    .. code-block:: python
+        da[0]  # index by position
+        da[0:5:2]  # index by slice
+        da[[0, 2, 3]]  # index by list of indices
+        da[True, False, True, True, ...]  # index by boolean mask
+
+    You can delete items from a DocumentArray like a Python List
```
JohannesMessner, line 20: Same question here about "advanced deletion"

docarray/array/stacked/list_advance_indexing.py
```diff
@@ -0,0 +1,42 @@
...
+
+    _data: MutableSequence[T_item]
+
+    def __init__(self, data: MutableSequence[T_item]):
+        self._data = data
+
+    def __len__(self) -> int:
+        return len(self._data)
+
+    @property
```
JohannesMessner, line 36: We might as well make `data` public? Or is the point to not have a setter?

JohannesMessner (changes requested):
I made a pass, but there's a lot to take in, so expect more suggestions at the next round, sorry in advance ;)

docarray/array/stacked/array_stacked.py
samsja: fixed

docarray/array/stacked/list_advance_indexing.py
samsja: done

docarray/array/array/sequence_indexing_mixin.py
```diff
@@ -0,0 +1,216 @@
...
+        for ix in item:
+            results.append(self._data[ix])
+        return self.__class__(results)
+
+    def _set_by_indices(self: T, item: Iterable[int], value: Iterable[T_item]):
+        # here we cannot use _get_offset_to_doc() because we need to change the doc
+        # that a given offset points to, not just retrieve it.
+        # Future optimization idea: _data could be List[DocContainer], where
+        # DocContainer points to the doc. Then we could use _get_offset_to_container()
+        # to swap the doc in the container.
```
JohannesMessner, line 117:
```suggestion
```

docarray/array/array/sequence_indexing_mixin.py
JohannesMessner: actually with this refactor the comment maybe doesn't make sense at all, i suggest we remove it

docarray/array/array/sequence_indexing_mixin.py
```diff
@@ -0,0 +1,216 @@
...
+            return False
+    return False  # this is unreachable, but mypy wants it
+
+
+class IndexingSequenceMixin(Iterable[T_item]):
+    """
+    This mixin allow sto extend a list into an object that can be indexed
+    a la numpy/pytorch.
+
+    You can index into a IndexingSequenceMixin like a numpy array or torch tensor:
```
JohannesMessner, line 41:
```suggestion
    You can index into, delete from, and set items in a IndexingSequenceMixin like a numpy array or torch tensor:
```

docarray/array/array/sequence_indexing_mixin.py
```diff
@@ -0,0 +1,216 @@
...
+        da[0]  # index by position
+        da[0:5:2]  # index by slice
+        da[[0, 2, 3]]  # index by list of indices
+        da[True, False, True, True, ...]  # index by boolean mask
+
+    You can delete items from a DocumentArray like a Python List
+
+    .. code-block:: python
+        del da[0]  # remove first element from DocumentArray
+        del da[0:5]  # remove elements for 0 to 5 from DocumentArray
```
JohannesMessner, line 53:
```suggestion
```

docarray/array/array/sequence_indexing_mixin.py
JohannesMessner:
```suggestion
```

docarray/array/stacked/list_advance_indexing.py
```diff
@@ -0,0 +1,42 @@
...
+        da[0]  # index by position
+        da[0:5:2]  # index by slice
+        da[[0, 2, 3]]  # index by list of indices
+        da[True, False, True, True, ...]  # index by boolean mask
+
+    You can delete items from a DocumentArray like a Python List
+
+    .. code-block:: python
+        del da[0]  # remove first element from DocumentArray
+        del da[0:5]  # remove elements for 0 to 5 from DocumentArray
```
JohannesMessner, line 24:
```suggestion
```

JohannesMessner (approved):

PR #1184, opened 2023-02-27
refactor: added "doc" namespace

### LINKED ISSUE #1135: Put all docarray method in document and da into a common namespace
# context

as we allow flexible schema we need a way to prevent user to use a key name that is already used by docarray. The best way to do it is to have a namespace for al of the method of `Document`. 

Example:

`doc.from_protobuf -> doc.doc_from_protobuf`

or 

`doc.da_from_protobuf`

then we just need to tell the user you cannot have a key name that start with `doc` as it is reserved for docarray method.

Pydantic v2 will have a similar approach : https://docs.pydantic.dev/blog/pydantic-v2/#model-namespace-cleanup

AsRaNi1:
Goals:
As docarray allow flexible schema it needs a way to prevent user to use a key name that is already used by docarray. The best way to do it is to have a namespace for all of the method of Document.
fixes #1135


docarray/array/array/array.py
```diff
@@ -313,15 +313,15 @@ def _set_by_mask(self: T, item: Iterable[bool], value: Sequence[BaseDocument]):
                 self._data[i] = value[i_value]
                 i_value += 1
 
-    def append(self, doc: T_doc):
```
samsja, line 316: this one will need to stay `append`

AsRaNi1:
@samsja I have made the changes

jupyterjazz:
The issue was already solved with a different approach so I'll close this PR. 
Thank you for your interest! and feel free to tackle another ticket 💪

PR #1187, opened 2023-02-28
feat: add map function

anna-charlotte:
Goals:

Add `map_docs` function and `map_batch`:

```python
docs = list(map_docs(da, lambda x : x, num_worker=4 ))
```
This will be different from doing 

```python
for doc in da:
    doc = f(doc)

```

- leverage multiprocessing by benchmarking in test, check that using 2 CPUS is faster then using 1

- [x] map_docs()
- [x] map_docs_batch()
- [x] benchmarking tests

### DIFF
--- a/docarray/utils/map.py
+++ b/docarray/utils/map.py
@@ -25,6 +25,28 @@ def map_docs(
     Return an iterator that applies `func` to every Document in `da` in parallel,
     yielding the results.
 
+    EXAMPLE USAGE
+
+    .. code-block:: python
+
+        from docarray import DocumentArray
+        from docarray.documents import Image
+        from docarray.utils.map import map_docs
+
+
+        def load_url_to_tensor(img: Image) -> Image:
+            img.tensor = img.url.load()
+            return img
+
+
+        da = DocumentArray[Image]([Image(url='/path/to/img.png') for _ in range(100)])
+        da = DocumentArray[Image](
+            list(map_docs(da, load_url_to_tensor, backend='thread'))
+        )  # threading is usually a good option for IO-bound tasks such as loading an image from url
+
+        for doc in da:
+            assert doc.tensor is not None
+
     :param da: DocumentArray to apply function to
     :param func: a function that takes a :class:`BaseDocument` as input and outputs
         a :class:`BaseDocument`.
@@ -86,6 +108,35 @@ def map_docs_batch(
     Return an iterator that applies `func` to every **minibatch** of iterable in parallel,
     yielding the results.
     Each element in the returned iterator is an :class:`AnyDocumentArray`.
+
+    EXAMPLE USAGE
+
+    .. code-block:: python
+        from docarray import BaseDocument, DocumentArray
+        from docarray.utils.map import map_docs_batch
+
+
+        class MyDoc(BaseDocument):
+            name: str
+
+
+        def upper_case_name(da: DocumentArray[MyDoc]) -> DocumentArray[MyDoc]:
+            da.name = [n.upper() for n in da.name]
+            return da
+
+
+        batch_size = 16
+        da = DocumentArray[MyDoc]([MyDoc(name='my orange cat') for _ in range(100)])
+        it = map_docs_batch(da, upper_case_name, batch_size=batch_size)
+        for i, d in enumerate(it):
+            da[i * batch_size : (i + 1) * batch_size] = d
+
+        assert len(da) == 100
+        print(da.name[:3])
+
+    .. code-block:: text
+
+        ['MY ORANGE CAT', 'MY ORANGE CAT', 'MY ORANGE CAT']
 
     :param batch_size: Size of each generated batch (except the last one, which might
         be smaller).


JohannesMessner:
Is `apply_batched` also in scope for this issue? And `map`/`map_batched`?

docarray/utils/apply.py
```diff
@@ -0,0 +1,93 @@
...
+
+T = TypeVar('T', bound=AnyDocumentArray)
+
+
+def apply(
+    da: T,
+    func: Callable[[BaseDocument], BaseDocument],
+    num_worker: Optional[int] = None,
+    pool: Optional['Pool'] = None,
+    show_progress: bool = False,
```
JohannesMessner, line 20: Should we have a `backend` options to switch between multi-processing and multi-threading?

samsja: @JoanFM maybe know more. But to me only multi processing makes sense. Can multi threading really improve performance here ?

JohannesMessner: Just the keep the PR discussion up to date with what we discussed on Discord: there will be multi threading since it makes sense for IO bound ops and tf/np/torch stuff

docarray/utils/apply.py
```diff
@@ -0,0 +1,306 @@
...
+        be responsible for closing the pool.
+
+    :yield: anything return from ``func``
+    """
+    if backend == 'process' and _is_lambda_or_partial_or_local_function(func):
+        raise ValueError(
+            f'Multiprocessing does not allow functions that are local, lambda or partial: {func}'
+        )
+
+    from rich.progress import track
```
samsja, line 266: you can move this one to the top. Rich is always install with docarray

docarray/utils/apply.py
```diff
@@ -0,0 +1,299 @@
...
+from docarray.array.abstract_array import AnyDocumentArray
+
+T = TypeVar('T', bound=AnyDocumentArray)
+T_doc = TypeVar('T_doc', bound=BaseDocument)
+
+
+def apply(
+    da: T,
+    func: Callable[[T_doc], T_doc],
+    backend: str = 'thread',
```
JohannesMessner, line 19: Not sure about the default option. The GIL means that threading is not a good option for many tasks; so mp seems like the safer default choice? Even though there is more overhead?

docarray/utils/apply.py
```diff
@@ -0,0 +1,299 @@
...
+        for doc in da:
+            assert doc.tensor is not None
+
+    :param da: DocumentArray to apply function to
+    :param func: a function that takes a :class:`BaseDocument` as input and outputs
+        a :class:`BaseDocument`.
+    :param backend: `thread` for multithreading and `process` for multiprocessing.
+        Defaults to `thread`.
+        In general, if `func` is IO-bound then `thread` is a good choice.
+        On the other hand, if `func` is CPU-bound, then you may use `process`.
```
JohannesMessner, line 56: I think we should mention that computation that is offloaded to non-python code (e.g. through np/torch/tf) falls under the "IO-bound" category for the purposes of this. Otherwise users would think that they have to use processing when they actually don't

docarray/utils/apply.py
```diff
@@ -0,0 +1,299 @@
...
+        In practice, you should try yourselves to figure out the best value.
+        However, if you wish to modify the elements in-place, regardless of IO/CPU-bound,
+        you should always use `thread` backend.
+
+        .. warning::
+            When using `process` backend, your `func` should not modify elements in-place.
+            This is because the multiprocessing backend passes the variable via pickle
+            and works in another process.
+            The passed object and the original object do **not** share the same memory.
```
JohannesMessner, line 65: Ok this is a good argument to keep the default as `thread`

docarray/utils/apply.py
```diff
@@ -0,0 +1,299 @@
...
+
+    :yield: Documents returned from `func`
+    """
+
+    if backend == 'process' and _is_lambda_or_partial_or_local_function(func):
+        raise ValueError(
+            f'Multiprocessing does not allow functions that are local, lambda or partial: {func}'
+        )
+
+    ctx_p: Union[nullcontext, Union[Pool, ThreadPool]]
```
JohannesMessner, line 124: Let's not be stingy on chars when it comes to variable naming ;) what does this mean? context pool?

docarray/utils/apply.py
```diff
@@ -0,0 +1,299 @@
...
+        return ThreadPool(processes=num_worker)
+    elif backend == 'process':
+        return Pool(processes=num_worker)
+    else:
+        raise ValueError(
+            f'`backend` must be either `process` or `thread`, receiving {backend}'
+        )
+
+
+def _is_lambda_or_partial_or_local_function(func: Callable[[Any], Any]) -> bool:
```
JohannesMessner, line 291: maybe make this a global helper? could be helpful elsewhere?

JohannesMessner (changes requested):
Looks good! We will make the map functions public, right?

.github/workflows/ci.yml
```diff
@@ -222,6 +222,31 @@ jobs:
           poetry run pytest -m 'tensorflow' tests
         timeout-minutes: 30
 
+  docarray-test-benchmarks:
```
samsja, line 225: you need to add this to `sucess-all-test` needs

JohannesMessner:
> > Looks good! We will make the map functions public, right?
> 
> I dont really have an opinion on this. @samsja suggested to keep only one of the two exposed, `map` or `apply`. Why would you prefer this @samsja

Why should we keep only one @samsja? I think both make sense, the user might already have an in-place or pure function that they want to use without rewriting

docarray/utils/apply.py
```diff
@@ -0,0 +1,296 @@
...
+            da, func, batch_size, backend, num_worker, shuffle, pool, show_progress
+        )
+    ):
+        indices = [i for i in range(i * batch_size, (i + 1) * batch_size)]
+        da[indices] = batch
+
+
+def _map_batch(
+    da: T,
+    func: Callable[[T], T],
```
JohannesMessner, line 219: Shouldn't this func also be from docarrray to docarray?

docarray/utils/apply.py
```diff
@@ -0,0 +1,296 @@
...
+        p = pool
+        context_pool = nullcontext()
+    else:
+        p = _get_pool(backend, num_worker)
+        context_pool = p
+
+    with context_pool:
+        imap = p.imap(func, da._batch(batch_size=batch_size, shuffle=shuffle))
+        for x in track(
+            imap, total=ceil(len(da) / batch_size), disable=not show_progress
```
samsja, line 280: what happened if the batch size is not a multiple of the len of da ? Can u add a test to cover this ?

docarray/utils/apply.py
```diff
@@ -0,0 +1,296 @@
...
+
+    with context_pool:
+        imap = p.imap(func, da)
+        for x in track(imap, total=len(da), disable=not show_progress):
+            yield x
+
+
+def apply_batch(
+    da: T,
+    func: Callable[[T], T],
```
JohannesMessner, line 144: Should we also allow funcs that go from a docarray to a single doc?

samsja (changes requested):
looks greats, I added a couple of suggestion

JohannesMessner (changes requested):
for the batched ones, do we have tests that use a `func` that changes the length of the docarray? Because that should be allowed, right?

tests/units/util/test_apply.py
```diff
@@ -0,0 +1,97 @@
...
+        doc.tensor = doc.url.load()
+    return da
+
+
+class MyImage(BaseDocument):
+    tensor: Optional[NdArray]
+    url: ImageUrl
+
+
+@pytest.mark.parametrize('n_docs,batch_size', [(10, 5), (10, 8)])
```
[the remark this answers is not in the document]
samsja: oh yes, good then

docarray/utils/apply.py
JohannesMessner: oops

docarray/utils/map.py
```diff
@@ -0,0 +1,151 @@
...
+def map_docs(
+    da: T,
+    func: Callable[[T_doc], T_doc],
+    backend: str = 'thread',
+    num_worker: Optional[int] = None,
+    pool: Optional[Union[Pool, ThreadPool]] = None,
+    show_progress: bool = False,
+) -> Generator[T_doc, None, None]:
+    """
+    Return an iterator that applies `func` to every Document in `da` in parallel,
```
samsja, line 25: can u put a python code example ?

docarray/utils/map.py
```diff
@@ -0,0 +1,151 @@
...
+        p = _get_pool(backend, num_worker)
+        context_pool = p
+
+    with context_pool:
+        imap = p.imap(func, da)
+        for x in track(imap, total=len(da), disable=not show_progress):
+            yield x
+
+
+def map_batch(
```
samsja, line 75:
```suggestion
def map_docs_batch(
```
maybe to be consistent ? No strong opinion though

samsja (approved):

JohannesMessner (approved):

PR #1196, opened 2023-03-01
feat: elasticsearch document index

AnneYang720:
Goals:

This PR will implement ElasticSearch Document Store based on elastic v7.  #1209 

There will be another ticket&PR for elastic v8.

**ToDo**
- [x] `__init__` and build index
- [x] getsetdel
- [x] find
- [x] filter
- [x] text search
- [x] query builder
- [x] tests
- [x] ci
- [ ] debug &optimization


docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,271 @@
...
+                'blob': doc.to_base64(),
+            }
+            for col_name, col in self._columns.items():
+                if not col.config:
+                    continue
+                request[col_name] = doc[col_name].tolist()
+            requests.append(request)
+
+        self._send_requests(request)
+        self._refresh(self._db_config.index_name)
```
girishc13, line 233: Refresh is used to update internal lucene index so that the new documents are available for search. Let's make this as runtime config. Depending on the number of docs the refresh operation can be expensive and unnecessary. If you're creating an index for the first then doing a refresh after completing all indexing operations is more efficient.

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,420 @@
...
+        return fn
+
+else:
+    # static type checkers do not like callable objects as decorators
+    from docarray.doc_index.abstract_doc_index import composable
+
+TSchema = TypeVar('TSchema', bound=BaseDocument)
+T = TypeVar('T', bound='ElasticDocumentIndex')
+
+MAX_ES_RETURNED_DOCS = 10000
```
girishc13, line 50: This is the default value. The user can choose to increase the [index.max_result_window](https://www.elastic.co/guide/en/elasticsearch/reference/8.6/index-modules.html#dynamic-index-settings) to a different value.

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,420 @@
...
+
+        self._refresh(self._db_config.index_name)
+
+    def _get_items(self, doc_ids: Sequence[str]) -> Sequence[TSchema]:
+        accumulated_docs = []
+        accumulated_docs_id_not_found = []
+
+        for pos in range(0, len(doc_ids), MAX_ES_RETURNED_DOCS):
+
+            es_rows = self._client.mget(
```
girishc13, line 196: Users should ideally use the [Scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/8.6/paginate-search-results.html#scroll-search-results) for fetching large results. We should throw an error if someone is using the `_get_items` for fetching all items or items more than some X% of the data size. I believe the `_get_items` must be a performant fetch using the current logic.

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,420 @@
...
+
+    @composable
+    def _find(
+        self, query: np.ndarray, search_field: str, limit: int, **kwargs
+    ) -> FindResult:
+        knn_query = {
+            'field': search_field,
+            'query_vector': query,
+            'k': limit,
+            'num_candidates': 10000
```
girishc13, line 227: Why is not a runtime config?

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,420 @@
...
+    def _text_search(
+        self,
+        query: str,
+        search_field: str,
+        limit: int,
+    ) -> FindResult:
+        query = {
+            "bool": {
+                "must": [
+                    {"match": {search_field: query}},
```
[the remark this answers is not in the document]
girishc13: As I found out yesterday, the user should use the `execute` method to execute advanced queries.

docarray/doc_index/backends/elastic_doc_index.py
AnneYang720: I removed the `MAX_ES_RETURNED_DOCS` here because I notice there already is a `max_result_window` in ElasticSearch. I  don't know when should we throw the error, like what if there are only 20 documents and the user wants them all.

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,454 @@
...
+
+ELASTIC_PY_VEC_TYPES = [list, tuple, np.ndarray]
+ELASTIC_PY_TYPES = [bool, int, float, str, docarray.typing.ID]
+if torch_imported:
+    import torch
+
+    ELASTIC_PY_VEC_TYPES.append(torch.Tensor)
+
+
+def _raise_has_substitue(name):
```
JohannesMessner, line 48: I think we have a global helper for this, no need to re-define it

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,454 @@
...
+        hosts: Union[str, List[str], None] = 'http://localhost:9200'
+        index_name: Optional[str] = None
+        es_config: Dict[str, Any] = field(default_factory=dict)
+        index_settings: Dict[str, Any] = field(default_factory=dict)
+
+    @dataclass
+    class RuntimeConfig(BaseDocumentIndex.RuntimeConfig):
+        default_column_config: Dict[Type, Dict[str, Any]] = field(
+            default_factory=lambda: {
+                np.ndarray: {
```
JohannesMessner, line 169: Would be nice to also handle torch and tensorflow, maybe also list of numbers etc

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,454 @@
...
+    ###############################################
+
+    def python_type_to_db_type(self, python_type: Type) -> Any:
+        """Map python type to database type."""
+        for allowed_type in ELASTIC_PY_VEC_TYPES:
+            if issubclass(python_type, allowed_type):
+                return np.ndarray
+
+        if python_type in ELASTIC_PY_TYPES:
+            return python_type
```
JohannesMessner, line 194: I am not sure this is done the way it is intended (so maybe the guidance need to be revised). What this should output is the type _as it is represented in the database_. So `docarray.typing.ID` would not be a valid type in the database, right? This should be mapped to something like 'string' or 'keyword'. I think you have put that info in the mapping above, but actually it should go here so it can automatically be parsed into `_ColumnInfo` :)

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,454 @@
...
+            self._search = Search(
+                using=outer_instance._client, index=outer_instance._index_name, **kwargs
+            )
+
+        def build(self, *args, **kwargs) -> Any:
+            return self
+
+        def query(self, *args, **kwargs):
+            self._search = self._search.query(*args, **kwargs)
+            return self
```
JohannesMessner, line 126: Maybe it would be more in the spirit to create a new `QueryBuilder` with updated state at each operation? that way every step remains intact and is not modifed in-place. I think you @JoanFM argue for this pattern, right?

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,454 @@
...
+    @dataclass
+    class RuntimeConfig(BaseDocumentIndex.RuntimeConfig):
+        default_column_config: Dict[Type, Dict[str, Any]] = field(
+            default_factory=lambda: {
+                np.ndarray: {
+                    'type': 'dense_vector',
+                    'dims': 128,
+                },
+                docarray.typing.ID: {'type': 'keyword'},
+                bool: {'type': 'boolean'},
```
JohannesMessner, line 174: I can't find where these `'type'` configuration are actually used, can you point me to it?

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,454 @@
...
+
+    @dataclass
+    class DBConfig(BaseDocumentIndex.DBConfig):
+        hosts: Union[str, List[str], None] = 'http://localhost:9200'
+        index_name: Optional[str] = None
+        es_config: Dict[str, Any] = field(default_factory=dict)
+        index_settings: Dict[str, Any] = field(default_factory=dict)
+
+    @dataclass
+    class RuntimeConfig(BaseDocumentIndex.RuntimeConfig):
```
JohannesMessner, line 166: Is there any other stuff worth configering? consistency levels, chunk/batch sizes, stuff like that? then it should also go intot this `RuntimeConfig`, with a sensible default

JohannesMessner (changes requested):

docarray/doc_index/backends/elastic_doc_index.py
AnneYang720: `'type'` is one of `self._index_init_params`, and is used in `_creat_index`:
`index = dict((k, col.config[k]) for k in self._index_init_params)`

docarray/doc_index/backends/elastic_doc_index.py
AnneYang720: I added `chunk_size`. @girishc13 What else do u think should be added here?

girishc13: I can't think of anything right now. We can just rely on the ES defaults and let the user customize additional configs.

tests/doc_index/elastic/v7/test_find.py
```diff
@@ -0,0 +1,271 @@
...
+    )
+
+    for docs, scores, query in zip(docs_batched, scores_batched, queries):
+        assert len(docs) == 5
+        assert len(scores) == 5
+        assert docs[0].id == query.id
+        assert np.allclose(docs[0].tens, query.tens)
+
+
+def test_filter():
```
violenil, line 129: Can you add some test cases for `range` and `terms` filters? For example, selecting multiple filter terms for one field, and a range of integers/floats for another field.

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,484 @@
...
+                'You are using Elasticsearch 8.0+ and the current client is 7.10.1. HNSW based vector search is not supported and the find method has a default implementation.'
+            )
+
+        body = {
+            'size': limit,
+            'query': {
+                'script_score': {
+                    'query': {'match_all': {}},
+                    'script': {
+                        'source': f'cosineSimilarity(params.query_vector, \'{search_field}\') + 1.0',
```
violenil, line 318: Should we make it possible for the user to select the similarity metric here and have a default set to `cosineSimilarity`?

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,484 @@
...
+                    [query], search_field
+                )[0]
+            else:
+                query_vec = query
+            query_vec_np = BaseDocumentIndex._to_numpy(self._outer_instance, query_vec)
+            self._query['size'] = limit
+            self._query['query']['script_score'] = {
+                'query': {'match_all': {}},
+                'script': {
+                    'source': f'cosineSimilarity(params.query_vector, \'{search_field}\') + 1.0',
```
violenil, line 145: Just a thought: the awesome thing about script score query is that we can make multiple vector similarity comparisons in one query. Do you think we can enable this here too? (Or maybe this is possible and I am missing it.) For example, I have a query with multiple embeddings and I want to compare each to a specific field in my index. Perhaps we can also make this a follow-up ticket.

violenil (commented):
Great work! This is exciting 🥳

docarray/doc_index/backends/elastic_doc_index.py
AnneYang720: Yeah it's possible. But now the `find` function only accepts one vector. Maybe we can discuss about this feature in the future.

docarray/doc_index/backends/elastic_doc_index.py
AnneYang720: Here each `Similarity` has different script, for example `l2_norm` should be 
`'source': '1 / (1 + l2norm(params.query_vector, \'tens\'))'`
I think the conclusion of this discussion is we have a default implementation here, and user can use QueryBuilder if they want to pass other score standard.

docarray/doc_index/backends/elastic_doc_index.py
```diff
@@ -0,0 +1,484 @@
...
+    np.ndarray,
+]
+
+if torch_imported:
+    import torch
+
+    ELASTIC_PY_VEC_TYPES.append(torch.Tensor)
+
+
+class ElasticDocumentIndex(BaseDocumentIndex, Generic[TSchema]):
```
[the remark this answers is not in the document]
AnneYang720: I think it's possible to use current QueryBuilder to execute aggregation. It also use search API.

violenil (approved):
LGTM!🚀

tests/index/elastic/v7/test_index_get_del.py
```diff
@@ -0,0 +1,232 @@
+import numpy as np
+import pytest
+
+from docarray import DocumentArray
+from docarray.index import ElasticDocumentIndex
+from tests.index.elastic.fixture import start_storage_v7  # noqa: F401
+from tests.index.elastic.fixture import DeepNestedDoc, FlatDoc, NestedDoc, SimpleDoc
+
+
```
JohannesMessner, line 9: It is probably best to do this in (a) separte file(s), but can we add tests for:
- persisting data and loading it with a new instance of our index class
- column configurations, where the user for example changes the DB column type of one field (from `keyword` to `text` or something like that)
- a test where we index an `ImageDoc` with some image tensor, make sure that that can be handled properly
- tensorflow and torch support (after you rebase this should work automatically, but good to test anyways)
- maybe other stuff that I am forgetting

This is my fault, I didn't write all necessary tests on the HNSLib implementation. But unfortunately I think we need more, I am also currently working on this for HNSWLib :)

JohannesMessner (changes requested):

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,480 @@
...
+            search_field: str = 'embedding',
+            limit: int = 10,
+        ):
+            if isinstance(query, BaseDocument):
+                query_vec = BaseDocumentIndex._get_values_by_column(
+                    [query], search_field
+                )[0]
+            else:
+                query_vec = query
+            query_vec_np = BaseDocumentIndex._to_numpy(self._outer_instance, query_vec)
```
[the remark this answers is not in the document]
AnneYang720: self here is `QueryBuilder` and `AttributeError: 'QueryBuilder' object has no attribute '_to_numpy'`

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,480 @@
...
+                index=self._index_name, body=self._db_config.index_settings
+            )
+
+        self._refresh(self._index_name)
+
+    ###############################################
+    # Inner classes for query builder and configs #
+    ###############################################
+
+    class QueryBuilder(BaseDocumentIndex.QueryBuilder):
```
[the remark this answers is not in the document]
AnneYang720: It supports. I change user's query structure in `def build` if `filter` and `find` both exist.

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,480 @@
...
+        raise ValueError(f'Unsupported column type for {type(self)}: {python_type}')
+
+    def _index(
+        self,
+        column_to_data: Dict[str, Generator[Any, None, None]],
+        refresh: bool = True,
+        chunk_size: Optional[int] = None,
+    ):
+
+        data = self._transpose_col_value_dict(column_to_data)  # type: ignore
```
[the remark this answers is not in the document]
AnneYang720: the `column_to_data` of `_index` is `Dict[str, Generator[Any, None, None]]`
and `_transpose_col_value_dict` accepts `Dict[str, Iterable[Any]]`
should I change the type hint of `_transpose_col_value_dict`?

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,480 @@
...
+                '_index': self._index_name,
+                '_id': row['id'],
+            }
+            for col_name, col in self._column_infos.items():
+                if col.db_type == np.ndarray and np.all(row[col_name] == 0):
+                    row[col_name] = row[col_name] + 1.0e-9
+                request[col_name] = row[col_name]
+            requests.append(request)
+
+        _, warning_info = self._send_requests(requests, chunk_size)
```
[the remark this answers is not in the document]
AnneYang720: I handled it in `_send_requests`.

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,480 @@
...
+            else:
+                accumulated_docs_id_not_found.append(row['_id'])
+
+        # raise warning if some ids are not found
+        if accumulated_docs_id_not_found:
+            warnings.warn(f'No document with id {accumulated_docs_id_not_found} found')
+
+        return accumulated_docs
+
+    def execute_query(self, query: Dict[str, Any], *args, **kwargs) -> Any:
```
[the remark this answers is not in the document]
AnneYang720: yes

docarray/index/backends/elastic.py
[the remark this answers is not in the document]
JohannesMessner: fair, my mistake

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,502 @@
...
+
+        elastic_py_types = {
+            bool: 'boolean',
+            int: 'integer',
+            float: 'float',
+            str: 'text',
+            docarray.typing.ID: 'keyword',
+        }
+
+        if python_type in elastic_py_types:
```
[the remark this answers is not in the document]
JohannesMessner: same as `ID` btw, it is also a string but is treated separately :)

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,546 @@
...
+if is_tf_available():
+    import tensorflow as tf  # type: ignore
+
+    from docarray.typing import TensorFlowTensor
+
+    ELASTIC_PY_VEC_TYPES.append(tf.Tensor)
+    ELASTIC_PY_VEC_TYPES.append(TensorFlowTensor)
+
+
+class ElasticDocIndex(BaseDocumentIndex, Generic[TSchema]):
```
JohannesMessner, line 59: I think we should call this one the `ElasticV7DocIndex` and the other one simply `ElasticDocIndex`, since v7 is a legacy version.

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,546 @@
...
+
+if is_tf_available():
+    import tensorflow as tf  # type: ignore
+
+    from docarray.typing import TensorFlowTensor
+
+    ELASTIC_PY_VEC_TYPES.append(tf.Tensor)
+    ELASTIC_PY_VEC_TYPES.append(TensorFlowTensor)
+
+
```
JohannesMessner, line 58: Do you think this index could inherit frome the v8 index and only change the methods and properties that need to be changed, so that we can share more code? Or is the difference between the two too big?

docarray/index/backends/elastic.py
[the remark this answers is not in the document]
JohannesMessner: hmm this is weird, a genarator should also be an iterable, so this should work, no?

JohannesMessner (commented):
just some small comments :)

docarray/index/backends/elastic.py
AnneYang720: v8 has more specific API (create, search, put_mapping etc.). For example
```python
body = {
    'size': limit,
    'query': {
        'bool': {
            'must': {'match': {search_field: query}},
        }
    },
}

#v7
resp = self._client.search(index=self._index_name,body=body,)
#v8
resp = self._client.search(index=self._index_name, query=search_query, size=limit,)
```

But I think the difference is not too big, they share many common helper functions.

docarray/index/backends/elastic.py
```diff
@@ -0,0 +1,546 @@
...
+        self._client = Elasticsearch(
+            hosts=self._db_config.hosts,
+            **self._db_config.es_config,
+        )
+
+        # compatibility
+        self._server_version = self._client.info()['version']['number']
+        if int(self._server_version.split('.')[0]) >= 8:
+            os.environ['ELASTIC_CLIENT_APIVERSIONING'] = '1'
+
```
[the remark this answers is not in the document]
JohannesMessner: perfect

docarray/index/backends/elastic.py
JohannesMessner: ok sounds good, we can optimize code re-use once this one is merged

JohannesMessner (approved):

Issue #1189, opened 2023-02-28
Initialization and construction of a `Document` and `DocumentArray` takes too long


anna-charlotte:
**Describe the bug**
Initialization and construction of a `Document` and `DocumentArray` takes a lot of time compared to simple list initialization:

```python
from time import time 
n_docs = 1_000_000

# list initialisation
start_time = time()
numbers = [i for i in range(n_docs)]
duration = time() - start_time
print(f"duration list init = {duration}")

# DocumentArray initialisation
start_time = time()
numbers = DocumentArray[MyDoc]([MyDoc(number=i) for i in range(n_docs)])
duration = time() - start_time
print(f"duration da init = {duration}")

# construction of list of Documents
start_time = time()
numbers = [MyDoc.construct(number=i) for i in range(n_docs)]
duration = time() - start_time
print(f"duration list construct = {duration}")

# construction of DocumentArray of Documents
start_time = time()
numbers = DocumentArray[MyDoc]([MyDoc.construct(number=i) for i in range(n_docs)])
duration = time() - start_time
print(f"duration da construct = {duration}")
```

output:
```
duration list init         =  0.035775423049926 s
duration da init           = 11.883232831954956 s
duration list construct    = 10.953930139541626 s
duration da construct      =  9.914959907531738 s
```

**Expected behavior**
Quicker.

samsja:
@anna-charlotte it would be nice to see what kind of speed up pydantic v2 would offer.

you could do the same experiment but instead of creating a document you create a pedantic base model instance using pydantic v2.


Overall I think we cannot make this faster. That is on pydantic, and pydantic is doing nothing fancy there so that is just on python being slow for these custom things.

Could you post the results with dataclass as well ?

JoanFM:
So we do a lot more than `Pydantic`?

samsja:
> So we do a lot more than `Pydantic`?

we do nothing more than pydantic :thinking:

JoanFM:
then where does the difference come from?

JohannesMessner:
I'm not sure where else, but at least in the Document Indexes we depend on `.id` quite heavily. So how about having an id by default, but exposing a `Config` (same way pydantic has configs) to turn off id generation, as an optimization step that users can take?

Also, does this explain all the difference? That plain python dataclass is so much faster is surprising to me.

Also also, I agree that this seems looks problematic, but as long as we are faster than docarray v1 (which we seem to be) this should not block us, right?

[...]

PR #1197, opened 2023-03-01
fix: remove parsing of id in `BaseDocument` and make id optional

anna-charlotte:
Goals:

To make the performane of Document's init better, remove parsing of id and make id optional.
Related to this issue: #1189 

Removing `parse_as_obj` improves performance by factor 1.6.
Removing `parse_as_obj` and setting the `id=None` inside the class improves performance by factor 2.8


samsja (approved):

samsja (changes requested):
lets remove the sreenshot. This value will be obsolete quite fast. Lets just mention the factor of improvment

samsja (approved):

PR #1200, opened 2023-03-01
Fix: replaced ndarray with str

### LINKED ISSUE #1195: Incorrect type hint for `value` in `Url.validate()`
Currently the type hint for `value` in `.validate()` of the url classes is `value: Union[T, np.ndarray, Any]`. 
It should be `value: Union[T, str, Any]` though, since we actually expect a string here, not a numpy array.

This is the case for: 
- `AudioUrl`
- `AnyUrl`
- `ImageUrl`
- `VideoUrl`
- `Url3D`

hrik2001:
Should fix #1195 

Changed `np.ndarray` to `str` in validation methods of URL types

### DIFF
--- a/docarray/typing/url/audio_url.py
+++ b/docarray/typing/url/audio_url.py
@@ -26,7 +26,7 @@ class AudioUrl(AnyUrl):
     @classmethod
     def validate(
         cls: Type[T],
-        value: Union[T, np.ndarray, Any],
+        value: Union[T, str, Any],
         field: 'ModelField',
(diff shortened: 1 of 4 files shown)


hrik2001:
Hey @anna-charlotte! Thank you so much for leaving behind a very helful review and giving a heads up!
I have signed the commit as you mentioned. Do let me know if some other changes needs to be made :)

Thank You

JohannesMessner (approved):

Issue #1166, opened 2023-02-23
Check that we are doing proper file extension validation for all data types


JohannesMessner:
[Here](https://github.com/docarray/docarray/pull/1155) we make sure to do proper file extension validation for audio files paths, meaning:

- we have an exhaustive list of all audio file extensions
- if a different extension is given to the auio url, fail validation
- no _no_ extension is given, pass

We need to apply the same logic to all other modalities as well, such as image, text, video, ...

hrik2001:
Hi! I am eager to work on this issue and have also given a heads up at the docarray discord :)

I will be making separate PRs for each modality and will be linking them to this issue.

JohannesMessner:
After some consideration, we have changed our mind on this ticket. We no longer think it is a good idea to do this. You can see the reasons for this, and the follow-up work that needs to be done, here: https://github.com/docarray/docarray/issues/1309

PR #1207, opened 2023-03-04
feat: support other text formats

hrik2001:
This PR is to add file extension validation for text URL.
_(This PR is related to the issue #1166.)_

**Note:**
The protobuf test for text URL didn't test for remote URL (it only had local URL), moreover, it didn't check if the generated protobuf has `text_url` in it or not. I added both these features and also marked the test `slow` and `internet`, just how it is being done in other tests (for example test for `audio_url`). Do let me know if such change is not desired (as this change was not part of the issue), then I will add a commit removing that and squash.


hrik2001:
Looks like my changes broke other tests, hence converting this PR to a draft for the time being, until it's all fixed! :D

JohannesMessner (approved):
Looks good to me, once the tests pass ;)

JohannesMessner (dismissed):
Approved accidentally, the test need to pass of course, but looks good so far!

docarray/typing/url/text_url.py
```diff
@@ -11,6 +18,27 @@ class TextUrl(AnyUrl):
...
+    ):
+        import os
+        from urllib.parse import urlparse
+
+        url = super().validate(value, field, config)  # basic url validation
+        path = urlparse(url).path
+        ext = os.path.splitext(path)[1][1:].lower()
+
+        # pass test if extension is valid or no extension
+        has_text_extension = ext in TEXT_FILE_FORMATS or ext == ''
```
samsja, line 36:
```suggestion
        has_valide_text_extension = ext in TEXT_FILE_FORMATS or ext == ''
```

tests/units/typing/url/test_text_url.py
```diff
@@ -57,5 +77,17 @@ def test_json_schema():
 
 
 def test_dump_json():
```
samsja, line 79:
```suggestion
@pytest.mark.internet
def test_dump_json():
```

samsja (approved):

PR #1215, opened 2023-03-07
feat(index): automatically convert a dict to Document

AnneYang720:
Goals:

Add a common helper function to BaseDocumentIndex that converts a dict to Document.


docarray/doc_index/abstract_doc_index.py
```diff
@@ -342,14 +344,22 @@ def __getitem__(
             return_singleton = False
         # retrieve data
         doc_sequence = self._get_items(key)
+        # check data
+        if return_singleton and len(doc_sequence) == 0:
```
JohannesMessner, line 348: Why is this check only relevant if `return_singleton` is true?

docarray/doc_index/abstract_doc_index.py
```diff
@@ -693,3 +703,35 @@ def _to_numpy(self, val: Any) -> Any:
...
+        return schema_cls(**doc_dict)
+
+    def _convert_to_doc_list(self, docs: List[Dict[str, Any]]) -> List[BaseDocument]:
+        """Convert a list of docs in dict type to a list of Document objects."""
+
+        doc_list = []
+        for doc_dict in docs:
+            doc = self._convert_dict_to_doc(doc_dict, self._schema)
+            doc_list.append(doc)
+
```
JohannesMessner, line 736: I think all of this can be re-written to:
```python
return [self._convert_dict_to_doc(doc_dict, self._schema) for doc_dict in docs]
```

docarray/doc_index/abstract_doc_index.py
```diff
@@ -693,3 +703,35 @@ def _to_numpy(self, val: Any) -> Any:
             return val.numpy()
         else:
             raise ValueError(f'Unsupported input type for {type(self)}: {type(val)}')
+
+    def _convert_dict_to_doc(
```
JohannesMessner, line 707: I wonder if pydantic already offers something liket this out of the box.
And if not, maybe this should be a global and public helper function that we offer?

JohannesMessner (changes requested):
We also need to update the guide. Specifically we need to tell people what this dictionary needs to look like

JohannesMessner (changes requested):
And we need tests :)

docarray/doc_index/abstract_doc_index.py
AnneYang720: Because the last line is `return out_da[0] if return_singleton else out_da`.

AnneYang720:
> And we need tests :)

added

docarray/doc_index/abstract_doc_index.py
```diff
@@ -39,7 +39,7 @@
 
 
 class FindResultBatched(NamedTuple):
-    documents: List[DocumentArray]
+    documents: Union[List[DocumentArray], List[List[Dict[str, Any]]]]
```
JohannesMessner, line 42: the `FindResultBatched` is the thing that the user gets in the end, and us such it should always carry Documents, not dicts. So i wold undo that change

docarray/doc_index/abstract_doc_index.py
```diff
@@ -411,10 +421,18 @@ def find(
...
         query_vec_np = self._to_numpy(query_vec)
-        return self._find(
-            query_vec_np, search_field=search_field, limit=limit, **kwargs  # type: ignore
+        docs, scores = self._find(
+            query_vec_np, search_field=search_field, limit=limit, **kwargs
         )
 
+        if isinstance(docs, List):
+            da_cls = DocumentArray.__class_getitem__(
+                cast(Type[BaseDocument], self._schema)
```
JohannesMessner, line 430: why is this cast necessary?

docarray/doc_index/abstract_doc_index.py
```diff
@@ -458,7 +484,15 @@ def filter(
...
         :return: a DocumentArray containing the documents that match the filter query
         """
-        return self._filter(filter_query, limit=limit, **kwargs)  # type: ignore
+        docs = self._filter(filter_query, limit=limit, **kwargs)
+
+        if isinstance(docs, List):
+            da_cls = DocumentArray.__class_getitem__(
+                cast(Type[BaseDocument], self._schema)
+            )
+            docs = da_cls(self._convert_to_doc_list(docs))
```
JohannesMessner, line 493: It might be worth breaking this logic into a separate private method, something like `_list_to_docarray()`.

JohannesMessner (changes requested):

docarray/doc_index/abstract_doc_index.py
```diff
@@ -342,14 +344,22 @@ def __getitem__(
...
         doc_sequence = self._get_items(key)
+        # check data
+        if return_singleton and len(doc_sequence) == 0:
+            raise KeyError(f'No document with id {key} found')
+
         # cast output
         if isinstance(doc_sequence, DocumentArray):
             out_da: DocumentArray[TSchema] = doc_sequence
         else:
+            if type(doc_sequence) is Sequence[Dict]:
```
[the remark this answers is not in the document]
AnneYang720: We cannot use `issubclass` or `isinstance` between `doc_sequence` and `Sequence[Dict]`. It causes `TypeError: Subscripted generics cannot be used with class and instance checks`.

We can check change to `if instance(doc_sequence[0], dict)`, since we already check length above.

JohannesMessner (approved):
LGTM, good job!

PR #1217, opened 2023-03-09
feat: Support other image extensions

hrik2001:
Goals:
This PR is to add file extension validation for image URL.
_(This PR is related to the issue https://github.com/docarray/docarray/issues/1166.)_

**Note:**
Made some changes to the test to make it more in line with the other tests


docarray/typing/url/filetypes.py
```diff
@@ -117,3 +117,67 @@
...
+    '.iim',
+    '.ftu',
+    '.vst',
+    '.pnm',
+    '.h5',
+    '.j2k',
+    '.xpm',
+    '.ftc',
+    '.icns',
+    '.gif',
```
JohannesMessner, line 149: Where does this list of extensions come from? Is it the extension supported by Pillow? Are we sure that something like `.gif` works for our purposes?

JohannesMessner (changes requested):

docarray/typing/url/filetypes.py
hrik2001: Yes these are extensions supported by pillow. I had run this code to find a list of all the supported extensions
```
from PIL import Image

exts = Image.registered_extensions()
supported_extensions = {ex for ex, f in exts.items() if f in Image.OPEN}
```

The list consisted of formats like `mpeg` and `mpg` too, but I removed them since they were video formats. Looks like I forgot to remove the GIF format. Pushing a commit to remove gif. Thanks a lot :)

docarray/typing/url/image_url.py
```diff
@@ -33,10 +32,7 @@ def validate(
         url = super().validate(value, field, config)  # basic url validation
         has_image_extension = any(url.endswith(ext) for ext in IMAGE_FILE_FORMATS)
```
JohannesMessner, line 33: To align with other modalities, we should also allow no extension at all. So to pass validation the rule should be:
- Either there is an extension, and the extension is in `IMAGE_FILE_FORMATS`
- OR there is not extension at all
You can take a look at how this is handled for audio data, for example.

And please also adjust the error message below accordingly :) Thanks!

JohannesMessner (changes requested):

docarray/typing/url/filetypes.py
JohannesMessner: perfect

hrik2001:
Hey @JohannesMessner, have committed according to the review. Feel free to take a look at it :D

samsja:
closing as we don't want this anymore

PR #1220, opened 2023-03-10
feat(index): index data with union types

JohannesMessner:
Goals:

Before this PR it was impossible to index data if it came from a Document with Union types, including Optional types:

```python
class MyDoc(BaseDocument):
    tensor: NdArray[128]

class MyUnionDoc(BaseDocument):
        tensor: Union[NdArray[128], str]

class MyOptionalDoc(BaseDocument):
        tensor: Optional[NdArray[128]]


index = MyDocIndex[ImageDoc](...)

in_union = [MyUnionDoc(tensor=np.random.rand(10)]  # all the data is there, should be no problem
index.index(in_union)  # fails :/

in_optional = [MyOptionalDoc(tensor=np.random.rand(10)]  # all the data is there, should be no problem
index.index(in_optional)  # fails :/
```

After this PR these cases will work, while pathological cases still fail:

```python
class MyDoc(BaseDocument):
    tensor: NdArray[128]

class MyUnionDoc(BaseDocument):
        tensor: Union[NdArray[128], str]

class MyOptionalDoc(BaseDocument):
        tensor: Optional[NdArray[128]]


index = MyDocIndex[ImageDoc](...)

# good cases

in_union = [MyUnionDoc(tensor=np.random.rand(10)]  # all the data is there, should be no problem
index.index(in_union)  # this now works!

in_optional = [MyOptionalDoc(tensor=np.random.rand(10)]  # all the data is there, should be no problem
index.index(in_optional)  # this now works!

# bad cases

in_union = [MyUnionDoc(tensor='hello']  # wrong kind of data, should fail
index.index(in_union)  # still fails

in_optional = [MyOptionalDoc(tensor=None]  # no data, should fail
index.index(in_optional)  # still fails
```

This is achieved by leveraging pydantic to parse the input data. If that succeeds despite, we allow to proceed.

**Also** this PR contains a fix for equality and hashing on parametrized types; part of this PR bc it is need for the tests:

```python
# the following were `False` before the fix but are now `True`
NdArray[128] == NdArray[128]
hash(NdArray[128]) == hash(NdArray[128])
```


docarray/doc_index/backends/hnswlib_doc_index.py
```diff
@@ -161,9 +161,9 @@ def index(self, docs: Union[BaseDocument, Sequence[BaseDocument]], **kwargs):
         """Index a document into the store"""
         if kwargs:
             raise ValueError(f'{list(kwargs.keys())} are not valid keyword arguments')
-        doc_seq = docs if isinstance(docs, Sequence) else [docs]
-        data_by_columns = self._get_col_value_dict(doc_seq)
-        hashed_ids = tuple(self._to_hashed_id(doc.id) for doc in doc_seq)
+        docs_validated = self._validate_docs(docs)
```
samsja, line 164: where is this function define ?

docarray/doc_index/abstract_doc_index.py
```diff
@@ -626,31 +628,60 @@ def build_query(self) -> QueryBuilder:
         """
         return self.QueryBuilder()  # type: ignore
 
-    def _create_columns(self, schema: Type[BaseDocument]) -> Dict[str, _ColumnInfo]:
-        columns: Dict[str, _ColumnInfo] = dict()
+    def _flatten_schema(
+        self, schema: Type[BaseDocument], name_prefix: str = ''
+    ) -> List[Tuple[str, Type, 'ModelField']]:
+        """Flatten the schema of a Document into a list of column names and types.
```
samsja, line 634: maybe add a line in the docstring about that fact that it handle the nested schema in a recursive manner by adding `__`

docarray/doc_index/abstract_doc_index.py
```diff
@@ -626,31 +628,60 @@ def build_query(self) -> QueryBuilder:
         """
         return self.QueryBuilder()  # type: ignore
 
-    def _create_columns(self, schema: Type[BaseDocument]) -> Dict[str, _ColumnInfo]:
-        columns: Dict[str, _ColumnInfo] = dict()
+    def _flatten_schema(
```
samsja, line 631: it can be a classmethod no ?

docarray/doc_index/abstract_doc_index.py
```diff
@@ -626,31 +628,60 @@ def build_query(self) -> QueryBuilder:
...
+
             if is_union_type(t_):
+                union_args = get_args(t_)
+                if len(union_args) == 2 and type(None) in union_args:
+                    # simple "Optional" type, treat as special case:
+                    # treat as if it was a single non-optional type
+                    for t_arg in union_args:
+                        if t_arg is type(None):
+                            pass
+                        elif issubclass(t_arg, BaseDocument):
```
samsja, line 653: what if it is not a Union of document ? Example `Union[BaseDocument, int]` what should happen ? Failing ?

samsja (commented):
added a couple of comments

docarray/doc_index/abstract_doc_index.py
samsja: can we add unit test for this function ?

docarray/doc_index/abstract_doc_index.py
```diff
@@ -672,30 +703,48 @@ def _create_single_column(self, field: 'ModelField', type_: Type) -> _ColumnInfo
...
+        out_docs = []
+        for i in range(len(docs)):
+            # validate the data
+            try:
+                out_docs.append(
+                    cast(Type[BaseDocument], self._schema).parse_obj(docs[i])
+                )
+            except (ValueError, ValidationError):
+                raise ValueError(
+                    'The schema of the input Documents is not compatible with the schema of the Document Index.'
```
jupyterjazz, line 740: should we maybe print this once and not for every invalid document?

docarray/doc_index/abstract_doc_index.py
```diff
@@ -672,30 +703,48 @@ def _create_single_column(self, field: 'ModelField', type_: Type) -> _ColumnInfo
...
-        """Flatten a DocumentArray into a DocumentArray of the schema type."""
-        reference_col_db_types = [
-            (name, col.db_type) for name, col in self._column_infos.items()
-        ]
-        if isinstance(docs, AnyDocumentArray):
-            input_columns = self._create_columns(docs.document_type)
-            input_col_db_types = [
-                (name, col.db_type) for name, col in input_columns.items()
-            ]
+    def _validate_docs(
```
jupyterjazz, line 706: would be nice to have docstrings for this and `_create_column_infos`

docarray/doc_index/backends/hnswlib_doc_index.py
jupyterjazz: @samsja inside the abstract class

tests/doc_index/base_classes/test_base_doc_store.py
```diff
@@ -151,81 +151,115 @@ class OtherFlatDoc(FlatDoc):
...
+    in_list = [OptionalDoc(tens=np.random.random((10,)))]
+    assert isinstance(store._validate_docs(in_list), DocumentArray[BaseDocument])
+    in_da = DocumentArray[OptionalDoc](in_list)
+    assert store._validate_docs(in_da) == in_da
+
+    with pytest.raises(ValueError):
+        store._validate_docs([OptionalDoc(tens=None)])
+
+    # OTHER UNION
+    store = DummyDocIndex[SimpleDoc]()
```
jupyterjazz, line 255: should we add tests for `HnswDocumentIndex` too?

jupyterjazz (commented):
Nice! 
left some comments

docarray/doc_index/abstract_doc_index.py
```diff
@@ -672,30 +703,48 @@ def _create_single_column(self, field: 'ModelField', type_: Type) -> _ColumnInfo
...
-                # see schema translation ideas in the design doc
-                if reference_col_db_types != input_col_db_types:
-                    return False
-            return True
+            names_compatible = reference_names == input_names
+            types_compatible = all(
+                (not is_union_type(t2) and issubclass(t1, t2))
+                for (t1, t2) in zip(reference_types, input_types)
+            )
+            if names_compatible and types_compatible:
```
jupyterjazz, line 729: so if this is False, docs get checked again and only after the error is thrown, right? should we somehow separate these cases so that we don't have to check docs twice? especially since the second check is a costly operation as we do it for each doc

tests/doc_index/base_classes/test_base_doc_store.py
JohannesMessner: I will do that separately, since work on that index is still ongoing

docarray/doc_index/abstract_doc_index.py
JohannesMessner: it is just returned like that, without recursive flattening. I don't think we can do anything else. But I wouldn't fail here, since this only does the flattening as much as it can, and having this kind of thing returned might still be useful somewhere. But I guess it should fail at the call site

docarray/doc_index/abstract_doc_index.py
jupyterjazz: > Only if this is not successful we iterate over the docs

Do we need to iterate over docs if we already know that it's not successful? this was my point

But I guess if the schema is not compatible, error will be thrown right away for the first doc at line 767 so it's not really an iteration over all docs

JohannesMessner: yes, we still need to check because the actual data might be compatible even if the schema says otherwise.

For example, the input could be of type `DocumentArray[BaseDocument]`, but contain only `ImageDoc`. This is valid since `ImageDoc` is a subclass of `BaseDocument`.
If the Document Index has schema `ImageDoc`, then the fast validation will fail, because `BaseDoc != ImageDoc`, but actually it should work, since all Documents are actually `ImageDoc`. So we have to do the exhaustive check over the data.

So basically the validation check based on the DocumentArray parametrization can be seen as just a shortcut that may or may not work.

jupyterjazz: makes sense, thanks for clarification!

jupyterjazz (approved):
LGTM!

PR #1222, opened 2023-03-10
feat: weaviate document index

### LINKED ISSUE #1210: DocumentIndex: support for Weaviate

hsm207:
Goals:

This PR will implement Weaviate Document Store.

Fixes #1210 

ToDo

- [ ]  __init__ and build index
- [ ] getsetdel
- [ ] find
- [ ] filter
- [ ] text search
- [ ] debug &optimization
- [ ] tests
- [ ] ci


docarray/doc_index/backends/weaviate_doc_index.py
```diff
@@ -0,0 +1,525 @@
...
+                f"Found index {self._db_config.index_name} with schema {schema}. Will reuse existing schema."
+            )
+        else:
+            self._client.schema.create_class(schema)
+
+    @dataclass
+    class DBConfig(BaseDocumentIndex.DBConfig):
+        host: str = 'http://weaviate:8080'
+        index_name: str = 'Document'
+        batch_config: Dict[str, Any] = field(
```
JoanFM, line 137: this `batch_config` should go to the `RuntimeConfig` right?

docarray/doc_index/backends/weaviate_doc_index.py
```diff
@@ -0,0 +1,525 @@
...
+
+        if python_type in WEAVIATE_PY_TYPES:
+            return python_type
+
+        raise ValueError(f'Unsupported column type for {type(self)}: {python_type}')
+
+    def build_query(self) -> BaseDocumentIndex.QueryBuilder:
+        return self.QueryBuilder(self)
+
+    class QueryBuilder(BaseDocumentIndex.QueryBuilder):
```
JoanFM, line 457: What I see here now is that there is no `Hybrid Search` implemented right?

hsm207: yes, that is not implemented yet.

docarray/doc_index/backends/weaviate_doc_index.py
```diff
@@ -0,0 +1,594 @@
...
+        default_column_config: Dict[Type, Dict[str, Any]] = field(
+            default_factory=lambda: {
+                np.ndarray: {
+                    'dataType': ['number[]'],
+                },
+                docarray.typing.ID: {'dataType': ['string']},
+                bool: {'dataType': ['boolean']},
+                int: {'dataType': ['int']},
+                float: {'dataType': ['number']},
+                str: {'dataType': ['text']},
```
JohannesMessner, line 150: This dataType information should not be encoded here, rather it should be in the python_type_to_db_type method.
There are two reasons for this:
1) if done that way, and if the user manually wants to specify a different `col_type` in the `Field(..)` notation, the abstract class can automatically use the user defined type instead of the default one. This will then be reflected in the corresponding _ColumnInfo, and you don't have to worry about it
2) it allows you to use `issubclass()` etc to handle multiple classes and subclasses etc at once, so less writing for you :)

You can take a look at the ES PR and how it is done there

docarray/doc_index/backends/weaviate_doc_index.py
```diff
@@ -0,0 +1,594 @@
...
+            for doc_id in doc_ids
+        ]
+        where_filter = {
+            "operator": "Or",
+            "operands": operands,
+        }
+
+        # do a loop because there is a limit to how many objects can be deleted at
+        # in a single query
+        # see: https://weaviate.io/developers/weaviate/api/rest/batch#maximum-number-of-deletes-per-query
```
JohannesMessner, line 174: Your choice of course, but imo it would also be ok to let the user run into that limit here

docarray/doc_index/backends/weaviate_doc_index.py
```diff
@@ -0,0 +1,594 @@
...
+        score_name: Literal["certainty", "distance"] = "certainty",
+        score_threshold: Optional[float] = None,
+    ) -> _FindResultBatched:
+        if search_field != '':
+            logging.warning(
+                'Argument search_field is not supported for WeaviateDocumentIndex. Ignoring.'
+            )
+
+        qs = []
+        for i, query in enumerate(queries):
```
JohannesMessner, line 284: Can weaviate to a native batch operation here, without looping over the queries?

docarray/doc_index/backends/weaviate_doc_index.py
```diff
@@ -0,0 +1,594 @@
...
+                result[self.embedding_column] = additional_fields['vector']
+
+        return result
+
+    def _index(self, column_to_data: Dict[str, Generator[Any, None, None]]):
+        docs = self._transpose_col_value_dict(column_to_data)
+        index_name = self._db_config.index_name
+
+        with self._client.batch as batch:
+            for doc in docs:
```
JohannesMessner, line 371: Same here, a native batched implementation would be nice, if it exists

docarray/doc_index/backends/weaviate_doc_index.py
```diff
@@ -0,0 +1,594 @@
...
+            .do()
+        )
+
+        return self._format_response(results["data"]["Get"][index_name], "score")
+
+    def _text_search_batched(
+        self, queries: Sequence[str], search_field: str, limit: int
+    ) -> _FindResultBatched:
+        qs = []
+        for i, query in enumerate(queries):
```
JohannesMessner, line 400: batching again ^^

docarray/doc_index/backends/weaviate_doc_index.py
```diff
@@ -0,0 +1,594 @@
...
+        # TODO: decorator to check for errors
+        total_docs = result["data"]["Aggregate"][index_name][0]["meta"]["count"]
+
+        return total_docs
+
+    def python_type_to_db_type(self, python_type: Type) -> Any:
+        """Map python type to database type."""
+        for allowed_type in WEAVIATE_PY_VEC_TYPES:
+            if issubclass(python_type, allowed_type):
+                return np.ndarray
```
JohannesMessner, line 453: So linking back to my comment above, this should return the weaviate specific column type, not `np.ndarray`. Maybe this is not well explained in the manual, because in the ES implementation the same thing happened in the beginning...

tests/doc_index/weaviate/test_index_get_del.py
```diff
@@ -0,0 +1,319 @@
+import logging
+
```
JohannesMessner, line 2: we'll need a few more tests but you probably kne that :)

JohannesMessner (changes requested):
Looks like we are on a good track here!

docarray/index/backends/weaviate.py
```diff
@@ -0,0 +1,806 @@
...
+                'text': {},
+                'int': {},
+                'number': {},
+                'boolean': {},
+                'number[]': {},
+                'blob': {},
+            }
+        )
+
+        batch_config: Dict[str, Any] = field(
```
JohannesMessner, line 218: so here instead of having a batch_config we could have all keys that are currently inside of batch_config, directly in the RuntimeConfig

JohannesMessner (approved):

hsm207:
Closing since this is superseded by #1367

hsm207:
Closing since this is superseded by #1367

Issue #1177, opened 2023-02-24
Fix == for Document


samsja:

JoanFM:
I am not so sure about number 1.

samsja:
> I am not so sure about number 1.

which one ?

JoanFM:
about overlooking `id`. Not sure if it has some dangerous implications

samsja:
> about overlooking `id`. Not sure if it has some dangerous implications

The way I see it is that we mimmic python `==` which does not care if the actual data is under the same `id`

samsja:
maybe we could have a config option if people care about equality on `id`

```python

class MyDoc(BaseDocument):
    a: str
    
    class Config:
         equal_with_id: True


assert not MyDoc(a='hello') == MyDoc(a='hello)
```

JoanFM:
no, if we go like this we change it. We do not add so much complexity

JoanFM:
lets keep equality ignoring ID

JohannesMessner:
Let's make sure to document this properly

RStar2022:
Hi @samsja can I do this? 
Actually I am new to open source and trying to apply it in my VSCode, but when I try to execute the program, metaTorchAndNode class got a type error which cannot create a consistent MRO for bases type, ABCMeta. Is there something wrong about the code? 

```
Exception has occurred: TypeError
Cannot create a consistent method resolution
order (MRO) for bases type, ABCMeta
  File "docarray\docarray\typing\tensor\torch_tensor.py", line 27, in <module>
    class metaTorchAndNode(
  File "docarray\docarray\typing\tensor\embedding\torch.py", line 5, in <module>
    from docarray.typing.tensor.torch_tensor import TorchTensor
  File "docarray\docarray\typing\tensor\embedding\embedding.py", line 8, in <module>
    from docarray.typing.tensor.embedding.torch import TorchEmbedding
  File "docarray\docarray\typing\tensor\embedding\__init__.py", line 1, in <module>
    from docarray.typing.tensor.embedding.embedding import AnyEmbedding
  File "docarray\docarray\typing\tensor\__init__.py", line 1, in <module>
    from docarray.typing.tensor.embedding import AnyEmbedding, NdArrayEmbedding
  File "docarray\docarray\typing\__init__.py", line 3, in <module>
    from docarray.typing.tensor import ImageNdArray, ImageTensor
  File "docarray\docarray\base_document\io\json.py", line 4, in <module>
[...]

PR #1224, opened 2023-03-12
feat: implement == for document and document array

RStar2022:
Goals :
- Add Rich Comparison `__eq__` to check the equal type between dictionaries.
- Looping the key variable inside dictionary and compare the dictionary key. 
- Exclude `ID` from the key
- Check the value between `ndarray` key types and confirming shape. 
- Implementing fixed document of pydantic error to DocumentArray level `da1==da2`
- Add test feature

#1177


docarray/base_document/document.py
```diff
@@ -97,3 +99,24 @@ def __setattr__(self, field, value) -> None:
             for key, val in self.__dict__.items():
                 dict_ref[key] = val
             object.__setattr__(self, '__dict__', dict_ref)
+
+    def __eq__(self, cls) -> bool:
+        for key1,key2 in zip(self.dict(),cls.dict()):
```
samsja, line 104: I think you need to check that self and other have the same number of field

docarray/base_document/document.py
```diff
@@ -97,3 +99,24 @@ def __setattr__(self, field, value) -> None:
...
+            if key1 != key2 :
+                return False
+            if key1 == "id" and key2 == "id":
+                continue
+            
+            value1 = self.dict()[key1]
+            value2 = cls.dict()[key2]
+
+            if isinstance(value1, np.ndarray) and isinstance(value2, np.ndarray):
+                if not np.array_equal(value1, value2):
```
samsja, line 114: this will not cover TorchTensor and tensorflow. You need to rely on the abstracTensor class and the Computational backend. Probably you need to add `array_equal` in the computational backend

samsja (changes requested):
Thanks for contributing :rocket:  I added some comments in the codebase.

Please remember to add test as well

JoanFM (changes requested):
Please make sure to add tests for these features

docarray/base_document/document.py
RStar2022: I will be using len() to determine whether it has the same length or not. About the key I think I will be using set() to determine if it has the same key or not.

docarray/base_document/document.py
RStar2022: I don't understand what you mean by adding `array_equal` in the computational backend.
Do you mean that I need to add `__eq__` and `array_equal` func to all of the Computational backend (`torch_backend` , `AbstractComputationalBackend` , etc)?

RStar2022: Hey @samsja about relying on the Computational Backend and the abstractTensor class, do you mean by adding the `__eq__` in the backend file? 
If adding `__eq__` in the `abstract_computational_backend` class, is it enough by just adding the code below? If it is not, what should I compare in the backend and what should I add?
```
def __eq__(self : TTensor, other : TTensor) -> bool :
        if self.shape(TTensor) != other.shape(TTensor) :
            return False
```

samsja: yes exactly adding this the to backend !

Shape checking is not enough. probably this should look like this

```python
def __eq__(self : TTensor, other : TTensor) -> bool :
        return (self == other).all()
```

RStar2022: Thanks for the reply! Didn't know that actually works!

docarray/base_document/document.py
```diff
@@ -97,3 +99,24 @@ def __setattr__(self, field, value) -> None:
             for key, val in self.__dict__.items():
                 dict_ref[key] = val
             object.__setattr__(self, '__dict__', dict_ref)
+
+    def __eq__(self, cls) -> bool:
```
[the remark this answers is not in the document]
RStar2022: Fix it!

RStar2022:
Hey @samsja @JoanFM added the test!

docarray/base_document/document.py
```diff
@@ -97,3 +98,27 @@ def __setattr__(self, field, value) -> None:
             for key, val in self.__dict__.items():
                 dict_ref[key] = val
             object.__setattr__(self, '__dict__', dict_ref)
+
+    def __eq__(self, other) -> bool:
+        if self.dict().keys() != other.dict().keys() :
+            return False
+        
+        for key1,key2 in zip(self.dict(), other.dict()):
```
samsja, line 106:
```suggestion
        for field1,field2 in zip(self.dict(), other.dict()):
```

docarray/base_document/document.py
```diff
@@ -97,3 +98,27 @@ def __setattr__(self, field, value) -> None:
...
+            
+            value1 = self.dict()[key1]
+            value2 = other.dict()[key2]
```
samsja, line 109:
```suggestion
            value1 = getattr(self, field1)
            value2 = getattr(self, field2)
```

docarray/base_document/document.py
```diff
@@ -97,3 +98,27 @@ def __setattr__(self, field, value) -> None:
...
+            
+            if key1 == "id" and key2 == "id":
+                continue
```
samsja, line 112:
```suggestion
            if field_name == 'id'
                continue
```

docarray/computation/abstract_comp_backend.py
```diff
@@ -129,6 +129,12 @@ def isnan(cls, tensor: 'TTensor') -> 'TTensor':
         """Check element-wise for nan and return result as a boolean array"""
         ...
 
+    @classmethod
+    def equal(cls, self, other: 'TTensor') -> bool:
+        if cls.shape(self) != cls.shape(other):
+            return False
```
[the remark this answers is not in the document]
RStar2022: There are possibilities if either of the tensor shape is different.

docarray/base_document/document.py
RStar2022: Fix them!

docarray/base_document/document.py
RStar2022: Changed them!

docarray/array/array/array.py
```diff
@@ -147,7 +147,14 @@ def construct(
         da._data = docs if isinstance(docs, list) else list(docs)
         da.tensor_type = tensor_type
         return da
-
+    
+    def __eq__(self, other : Any) -> bool:
+        if self.__len__() != other.__len__() :
+            return False
+        for value_1, value_2 in zip(self._data(), other._data()):
```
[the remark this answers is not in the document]
RStar2022: Changed!

docarray/base_document/document.py
RStar2022: Changed them!

docarray/base_document/document.py
```diff
@@ -97,3 +98,28 @@ def __setattr__(self, field, value) -> None:
...
+
+    def __eq__(self, other) -> bool:
+        if self.dict().keys() != other.dict().keys():
+            return False
+
+        for field_name in self.__fields__:
+            value1 = getattr(self, field_name)
+            value2 = getattr(other, field_name)
+
+            if field_name == "id":
```
samsja, line 110:
```suggestion
            if field_name == 'id':
```

docarray/typing/tensor/tensorflow_tensor.py
```diff
@@ -298,3 +298,6 @@ def unwrap(self) -> tf.Tensor:
 
     def __len__(self) -> int:
         return len(self.tensor)
+
+    def __eq__(self: T, other: T) -> np.ndarray:
+        return tf.equal(self.tensor, other.tensor).numpy()
```
samsja, line 303:
```suggestion

    def __eq__(self: T, other: T):
        return self.tensor == other.tensor
```

samsja (changes requested):
Almost the end !

Can you please add a test with DocumentArray ? And please verify that you use single quote `'` everywhere instead of double quote `"`

RStar2022:
Added DocumentArray test and changed the single quote.

docarray/typing/tensor/tensorflow_tensor.py
RStar2022: Changed!

docarray/base_document/document.py
```diff
@@ -97,3 +98,28 @@ def __setattr__(self, field, value) -> None:
...
+                continue
+
+            if isinstance(value1, AbstractTensor) and isinstance(
+                value2, AbstractTensor
+            ):
+                comp_be = value1.get_comp_backend()
+                if not comp_be.equal(value1, value2):
+                    return False
+            else:
+                if value1 != value2:
```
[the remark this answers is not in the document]
RStar2022: Added if case that compare the type of value1 and value2 before comparing the tensor type value.

docarray/typing/tensor/tensorflow_tensor.py
RStar2022: @samsja after changing to your suggestion, the error occured which tells `TensorFlowTensor` class didn't have tensor attribute. Is it necessary to change it become tensor attribute?

RStar2022:
Hey @samsja need a little help with mypy test. 
When runs `abstract_tensor.py` file, it got error within the `__eq__` function, below is the error occured :
```
@abc.abstractmethod
def __eq__(self, other: object) -> bool:  # type: ignore
    ...
```
eventhough putting #type:ignore still mypy got the same error
```
error : Definition of `__eq__` in base class `_TensorBase` is incompatible with the definition in base class `AbstractTensor` 
```
And also when run `tensorflow_tensor.py` file, it got the same error within the `__eq__` function. 
```
    def __eq__(self: T, other: T) -> bool:  # type :ignore
        if not isinstance(other, TensorFlowTensor):
            return NotImplemented
        return (self.tensor == other.tensor).numpy()
```
And the errors : 
```
This violates the Liskov substitution principle
See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
It is recommended for "__eq__" to work with arbitrary objects, for example:
def __eq__(self, other: object) -> bool:
　if not isinstance(other, TensorFlowTensor):
　　　return NotImplemented
　return <logic to compare two TensorFlowTensor instances>
Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"
```
What should be added to those code so that the code can pass mypy test or is it better to ignore mypy by adding `type:ignore`? (eventhough I add `type:ignore` mypy still check the code)

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -239,6 +239,10 @@ def _docarray_stack(cls: Type[T], seq: Union[List[T], Tuple[T]]) -> T:
         # but mypy doesn't know that, so we ignore it here
         return cls._docarray_from_native(comp_backend.stack(seq))  # type: ignore
 
+    @abc.abstractmethod
+    def __eq__(self, other: object) -> bool:  # type: ignore
```
samsja, line 243: why do you need this one actualy ? I think since it is define at the computational backend level it should work wo

samsja:
> Hey @samsja need a little help with mypy test. When runs `abstract_tensor.py` file, it got error within the `__eq__` function, below is the error occured :
> 
> ```
> @abc.abstractmethod
> def __eq__(self, other: object) -> bool:  # type: ignore
>     ...
> ```
> 
> eventhough putting `#type:ignore` still mypy got the same error
> 
> ```
> error : Definition of `__eq__` in base class `_TensorBase` is incompatible with the definition in base class `AbstractTensor` 
> ```
> 
> And also when run `tensorflow_tensor.py` file, it got the same error within the `__eq__` function.
> 
> ```
>     def __eq__(self: T, other: T) -> bool:  # type :ignore
>         if not isinstance(other, TensorFlowTensor):
>             return NotImplemented
>         return (self.tensor == other.tensor).numpy()
> ```
> 
> And the errors :
> 
> ```
> This violates the Liskov substitution principle
> See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
> It is recommended for "__eq__" to work with arbitrary objects, for example:
> def __eq__(self, other: object) -> bool:
> 　if not isinstance(other, TensorFlowTensor):
> 　　　return NotImplemented
> 　return <logic to compare two TensorFlowTensor instances>
> Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"
> ```
> 
> What should be added to those code so that the code can pass mypy test or is it better to ignore mypy by adding `type:ignore`?

so it looks like you cannot overload `__eq__` :thinking: because it calsh with tensorflow / torch internal

Anyway  I think it might not be needed to add the `__eq__` at this level. I believe having it at the computational backend is enough and it will solve the problem

docarray/typing/tensor/abstract_tensor.py
RStar2022: I delete those code and just implement it in computational backend which actually works! The test also works fine without it!

RStar2022:
Yes, try to implement it and works, been stuck at those abstract tensor and tensorflowtensor for days, and it works just by using computational backend equal function. 
Thanks for the suggestion.

RStar2022:
Yes @samsja , still working with it, sorry for the late response.

tests/units/array/test_array.py
```diff
@@ -80,6 +82,79 @@ class Text(BaseDocument):
...
+
+    da1 = DocumentArray([Text(text='hello')])
+    da2 = DocumentArray([Text(text='hello')])
+
+    assert da1 == da2
+    assert da1 == [Text(text='hello') for _ in range(len(da1))]
+    assert da2 == [Text(text='hello') for _ in range(len(da2))]
+
+
+def test_tensorflowtensor_equality():
```
samsja, line 125:
```suggestion
pytest.mark.tensorflow
def test_tensorflowtensor_equality():
```

tests/units/array/test_array.py
```diff
@@ -80,6 +82,79 @@ class Text(BaseDocument):
...
+    tensor4 = Text(tensor=tf.constant([[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]]))
+
+    assert tensor1 == tensor2
+    assert tensor1 != tensor3
+    assert tensor1 != tensor4
+
+
+def test_text_tensor():
+    class Text(BaseDocument):
+        tensor: TensorFlowTensor
```
samsja, line 141: you can remove this one

samsja (changes requested):
I suggest some change to make the test pass.

RStar2022:
Add the pytest mark and removed the tensorflow type in test_tensor function

samsja:
@RStar2022 it seems that there is still a couple of test failing, I am not sure why.

FYI: don't worry about the test failing because of some token problem. We will merge the PR even if they fail

PR #1228, opened 2023-03-14
feat: ellipsis in tensor shape definition

jupyterjazz:
Introduces the first feature proposed here https://github.com/docarray/docarray/issues/1185

**Generic behavior for tensor shapes**
It's now possible to use ellipsis when defining tensor shapes. 
Example:
```python
def fn(batch: NdArray[100, ...])
```
which means `batch` is expected to have first dimension of size 100, and can have other arbitrary dimensions which won't be checked while validating the shape.

Other examples:
```python
# ellipsis in the beginning
def fn(batch: NdArray[..., 100]):
    pass
# ellipsis in the middle
def fn(batch: NdArray[3, ..., 100]):
    pass
```

**How it works**
If ellipsis is detected in the shape, remove it, and replace by the amount independent variables needed to make the number of dimensions equal for the actual shape and the shape definition. Then, proceed to check if the two shapes are compatible. 

### DIFF
--- a/docarray/typing/tensor/abstract_tensor.py
+++ b/docarray/typing/tensor/abstract_tensor.py
@@ -130,6 +130,7 @@ class AbstractTensor(Generic[TTensor, T], AbstractType, ABC, Sized):
                         f'Cannot use Ellipsis (...) more than once for the shape {shape}'
                     )
                 ellipsis_pos = ellipsis_occurrences[0]
+                # Calculate how many dimensions to add. Should be at least 1.
                 dimensions_needed = max(len(tshape) - len(shape) + 1, 1)
                 shape = (
                     shape[:ellipsis_pos]
--- a/docarray/typing/tensor/abstract_tensor.py
+++ b/docarray/typing/tensor/abstract_tensor.py
@@ -83,7 +83,6 @@ class _ParametrizedMeta(type):
 
 
 class AbstractTensor(Generic[TTensor, T], AbstractType, ABC, Sized):
-
     __parametrized_meta__: type = _ParametrizedMeta
     __unparametrizedcls__: Optional[Type['AbstractTensor']] = None
     __docarray_target_shape__: Optional[Tuple[int, ...]] = None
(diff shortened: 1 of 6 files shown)


tests/units/typing/tensor/test_tensor.py
```diff
@@ -52,6 +51,60 @@ def test_unwrap():
     assert (ndarray == np.zeros((3, 224, 224))).all()
 
 
+def test_ellipsis_in_shape():
+    # ellipsis in the end, two extra dimensions needed
+    tensor = parse_obj_as(NdArray[3, ...], np.zeros((3, 128, 224)))
```
samsja, line 56: can you make this test for torch tensor and tensorflow as well ?

tests/units/typing/tensor/test_tensor.py
```diff
@@ -52,6 +51,60 @@ def test_unwrap():
...
+    assert isinstance(tensor, NdArray)
+    assert isinstance(tensor, np.ndarray)
+    assert tensor.shape == (3, 128, 224)
+
+    # ellipsis in the end, no extra dimensions needed
+    tensor = parse_obj_as(NdArray[3, 128, 224, ...], np.zeros((3, 128, 224)))
```
samsja, line 62: I am not sure, should not this one fail instead?

tests/units/typing/tensor/test_tensor.py
jupyterjazz: should I parametrize it in this file or add tests into their separate test files?

tests/units/typing/tensor/test_tensor.py
jupyterjazz: yeah, makes sense to expect at least 1 dimension when users type ellipsis

tests/units/typing/tensor/test_tensor.py
samsja: you can parametrize for TorchTensor and NDarray. For tensorflow tensor you can add one or two test but not all of them.

Ideally we would only need to validate this feature at a unit test level and kind of be sure that it is working for all of the backend. But I think it is tough to do the unit test properly there

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -121,7 +120,25 @@ def __docarray_validate_shape__(cls, t: T, shape: Tuple[Union[int, str]]) -> T:
...
+            ellipsis_occurrences = [
+                pos for pos, dim in enumerate(shape) if dim == Ellipsis
+            ]
+            if ellipsis_occurrences:
+                if len(ellipsis_occurrences) > 1:
+                    raise ValueError(
+                        f'Cannot use Ellipsis (...) more than once for the shape {shape}'
+                    )
+                ellipsis_pos = ellipsis_occurrences[0]
+                dimensions_needed = max(len(tshape) - len(shape) + 1, 1)
```
samsja, line 133: can u add a comment on what is happening here ? I am not sure to understand

jupyterjazz: Can you check now? 

The motivation is not to allow validating shapes like [3, ...] against [3]. That's why I specified adding at least 1 extra dimension.

samsja: looks good now

samsja (approved):

PR #1244, opened 2023-03-15
feat: shift to mkdocs

samsja:
# Context

WIP: this pr try out mkdocs for docarray documentation. Before we make the call on what to use

## TODO

- [ ] Inherited member (not working right now : https://github.com/mkdocstrings/pytkdocs/issues/58#issuecomment-1475249881)
- [x] Docstring syntax highlight
- [ ] Something like {ref}`foo`
- [x] All classes in API ref


alexcg1:
@samsja what do you need from my end here? Just go through and check it all works ok? I haven't done any more playing with MkDocs beyond the test repo I did before

samsja:
> @samsja what do you need from my end here? Just go through and check it all works ok? I haven't done any more playing with MkDocs beyond the test repo I did before

IMO this should be done in another PR.

I think I just need you to take a brief look and tell me if it is coherent. We will polish, i.e, making all of the docstring looks better later. (I just need an approval to merge it )

PR #1252, opened 2023-03-17
feat: add user defined mapping for python type to db type

anna-charlotte:
When a user creates a Document Index with a schema, it is not always unambiguous what types should be used in the database:

```python
class MySchema(BaseDocument):
    text: str  # should this be a string or a varchar? if varchar, what length?
    embedding: NdArray  # should this be float tensor or boolean tensor?
    num: float  # float32 or float64?

index = MyDocIndex[MySchema]
```

Right now, there is the method `python_type_to_db_type` that disambiguates these things, but it leaves no user choice.

We should enable an **optional** feature like this:

```python
class MySchema(BaseDocument):
    text: str  = Field(..., col_type='varchar', max_len=2048)
    embedding: NdArray  = Field(col_type='boolean_tensor')
    num: float  = Field(col_type='float64')

index = MyDocIndex[MySchema]
```

To do so,  check in  `_create_single_column` if `col_type` is present, and if so override `db_type`.

- [x] add user defined mapping
- [x] add test

### DIFF
--- a/docarray/doc_index/abstract_doc_index.py
+++ b/docarray/doc_index/abstract_doc_index.py
@@ -654,8 +654,14 @@ class BaseDocumentIndex(ABC, Generic[TSchema]):
 
     def _create_single_column(self, field: 'ModelField', type_: Type) -> _ColumnInfo:
         custom_config = field.field_info.extra
+
         if 'col_type' in custom_config.keys():
             db_type = custom_config['col_type']
+            custom_config.pop('col_type')
+            if db_type not in self._runtime_config.default_column_config.keys():
+                raise ValueError(
+                    f'The given col_type is not a valid db type: {db_type}'
+                )
         else:
             db_type = self.python_type_to_db_type(type_)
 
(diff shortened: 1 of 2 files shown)


JohannesMessner (changes requested):
Woohoo, @anna-charlotte is back and delivering immediately!
Just one thing, can we add a sentence or two to the implementation guide (`docs/tutorials/add_doc_index.md`)?

Where it talks about `_ColumnInfo` and `self._column_infos` and/or where it talks about `python_type_to_db_type()`.
Something along the lines of

> "`_ColumnInfo.db_type` contains the data type of a particular database column, whereas `_ColumnInfo.docarray_type` contains the python type as specified in `self._schema`. By default, it holds that `_ColumnInfo.docarray_type == self.python_type_to_db_type(_ColumnInfo.db_type)`. However, you should not rely on this, because a user can manually specify a different `db_type`. Therefore, your implementation should rely on `ColumnInfo.db_type` and not directly call `python_type_to_db_type()`."

Or something like this

JohannesMessner:
Ah, and maybe it's also good to explain that in cases like this:
```python
class MySchema(BaseDocument):
    text: str  = Field(..., col_type='varchar', max_len=2048)
```
`max_len` kind of relates to the type, but it will simply appear in `_ColumnInfo.config` (maybe double check that this is actually the case?)

JohannesMessner (approved):

PR #1254, opened 2023-03-20
feat: add minimal logger

jupyterjazz:
Closes https://github.com/docarray/docarray/issues/1218

Initialize a simple logger in __init__.py, get it as an attribute of Document Index class, and add logs in few places.

Default logging level is WARNING. However, let's set it to INFO during tests so that we can monitor things better (could be set to DEBUG if needed which would show even more logs)

- [x] add a minimal logger


tests/doc_index/conftest.py
girishc13, line 1: :+1:

docarray/doc_index/abstract_doc_index.py
girishc13, line 1: LGTM. I think that having the info logging on the `delete` method is useful and having debug level on the other methods gives more flexibility. Otherwise the frequent searching and indexing might produce too much noise by default.

girishc13 (changes requested):

CONTRIBUTING.md
```diff
@@ -271,6 +271,36 @@ This allows:
...
+<a name="-enable-logging"></a>
+## Enable logging
+See more logs by setting the log level to `DEBUG`.
+
+Example:
+```python
+import ...
+# import logging and set the level to DEBUG
+import logging
+logging.getLogger('docarray').setLevel(logging.DEBUG)
```
samsja, line 283: Not sure if it should be in the contributing file.

I think the part that says to the user how to enable logging should not be in the contributing guideline but rather in the readme for now ( we need to build documentation later anyway)

THough we might need something in the contributing guideline that say: if you need to do some logging you can use docarray logger

docarray/doc_index/abstract_doc_index.py
```diff
@@ -85,10 +86,13 @@ def __init__(self, db_config=None, **kwargs):
                 'A DocumentIndex must be typed with a Document type.'
                 'To do so, use the syntax: DocumentIndex[DocumentType]'
             )
+        self._logger = logging.getLogger('docarray')
```
JohannesMessner, line 89: shouldn't we import the logger from where it is initially set up instead of calling `getLogger()` again? Maybe it's just me but that would feel nicer to me

docarray/doc_index/abstract_doc_index.py
```diff
@@ -85,10 +86,13 @@ def __init__(self, db_config=None, **kwargs):
...
         self._db_config = db_config or self.DBConfig(**kwargs)
         if not isinstance(self._db_config, self.DBConfig):
             raise ValueError(f'db_config must be of type {self.DBConfig}')
+        self._logger.info('DB config created')
```
samsja, line 93: isn't info too much intrusive ? I would say 'debug' by default

docarray/doc_index/abstract_doc_index.py
Jackmin801: importing increase the risk of cyclic imports though. `getLogger()` would not? I am not sure about the general guidelines around this.

samsja: maybe @girishc13 you have an opinion their? Otherwise we can stick with sentry blog reco

jupyterjazz: > I see [sentry blog](https://blog.sentry.io/2022/07/19/logging-in-python-a-developers-guide/#python-logging-best-practices) recommending this: logging.getLogger(__name__)

I've also only encountered this approach, however, in that example, they create separate loggers for each module (i.e. they're using `getlogger` to create a new logger and not to get existing one), which is not similar to our case because currently we only have 1 logger ('docarray'). 

But this makes me wonder if we should initialize this one logger in the abstract docindex class and not in init.py since we only use it for docindex stuff. Thus, we won't have to call `getlogger` to get an existing logger.

girishc13: Sentry is a tool used for central server side logging. It makes sense to use `logging.getLogger(__name__)` when running a server with a specific log level and configuration for the whole application. This type of logging is a global control and reduces the flexibility. The main reason for using module level logging is to zoom in on specific functionalities of the docarray library.

I also think that we should not force the import and just use `logging.getLogger('docarray')`. The users are of course also allowed to import the logger.

docarray/doc_index/abstract_doc_index.py
jupyterjazz: Agree with Jack.

I was following python's [tutorial about logging](https://docs.python.org/3/howto/logging.html), and based on the definition on what should be logged as info and debug (DEBUG | Detailed information, typically of interest only when diagnosing problems. INFO | Confirmation that things are working as expected.) I think we should either keep it as INFO or remove the log (in case it doesn't bring much value) 

I'd say let's keep it, and if the tests are too verbose we can leave the log level to default (WARNING)

CONTRIBUTING.md
jupyterjazz: ok makes sense, logs are not for contributors only

docarray/doc_index/abstract_doc_index.py
jupyterjazz: Hmm, you mean too much noise in the tests? because we set the level to INFO?
In that case maybe we should leave it to WARNING?
Otherwise I'm not sure if I understand or agree with your suggestion, because logs for delete and other actions (such as index, search) look like the same importance/level to me. 
wdyt?

docarray/doc_index/abstract_doc_index.py
jupyterjazz: Default logging level is WARNING, so you won't see this by default, if that's your concern here.
I set logging level to INFO for tests only, so you'll see this when running tests.

samsja: Ah okay then makes sense !

CONTRIBUTING.md
```diff
@@ -271,6 +271,14 @@ This allows:
 * the reviewer to be very confident that the feature does what it is supposed to do before merging it into the code base.
 * the contributors to be sure that they don't break already-merged features when refactoring or modifying the code base.
 
+<a name="-enable-logging"></a>
+## Enable logging
+Monitor and debug your code by enabling docarray logging:
```
samsja, line 276:
```suggestion
If you need to monitor and debug your code you can do it by enabling docarray logging:
```

docarray/doc_index/abstract_doc_index.py
jupyterjazz: Sure, and maybe it’s because I’m inexperienced with logging, but I don’t think we should differentiate levels only based on how frequently something is called

However, if you think that’s more practical, I’m ok with that. This is still just an initial version and there’ll be more iterations to improve it

README.md
```diff
@@ -444,6 +444,35 @@ match = store.find(
 )
 ```
 
+## Enable logging
+See more logs by setting the log level to `DEBUG`.
+
+Example:
+```python
+import ...
```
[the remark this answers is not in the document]
jupyterjazz: ok, first let's see if it makes sense with actual imports. can you check?

docarray/doc_index/abstract_doc_index.py
samsja: lets go with @girishc13 suggestion I guess then, I don't have strong opinion neither here

samsja (approved):

PR #1262, opened 2023-03-21
fix: disable pycharm da property detection

anna-charlotte:
Goals:

PyCharm does currently not detect properties from DocumentArray and complains with `Unresolved attribute reference 'title' for class 'DocumentArray'`:

image

For now, we want to silence the error message from pycharm. This can be done by simply implementing `__getattr__()` in DocumentArray as described [here](https://youtrack.jetbrains.com/issue/PY-47991).

- [ ] add `__getattr__()`


samsja:
I think we should rather override `__get_atr_` otherwise pycharm cannot work with "normal" method of document array because they are hidden in the `__getatratributes__`. Whereas `__get_atr__` will only be called after the initial look up fail

PR #1266, opened 2023-03-21
refactor: doc index structure

jupyterjazz:
Two suggestions:

### shorter imports
Since developers will most likely only interact with specific backend implementations, i.e. will only import those classes, let's make the import shorter.

Before:
```python
from docarray.doc_index.backends.hnswlib_doc_index import HnswDocumentIndex
```
Now:
```python
from docarray.index import HnswDocumentIndex
```

### less verbose file names
Being inside `doc_index` module, it's already clear that `abstract_doc_index.py` (same with `hnswlib_doc_index.py`) is about `doc_index`, so let's make it less verbose.

Before:
```python
doc_index/abstract_doc_index.py
doc_index/backend/hnswlib_doc_index.py
```

Now:
```python
index/abstract.py
index/backend/hnswlib.py
```


JohannesMessner:
```python
from docarray.doc_index import HnswDocumentIndex
```
I think we should also rename the `doc_index` part, i find it a bit verbose and i don't like having an underscore. Just `index` might be not so clear though, any other suggestions?

samsja:
> ```python
> from docarray.doc_index import HnswDocumentIndex
> ```
> 
> I think we should also rename the `doc_index` part, i find it a bit verbose and i don't like having an underscore. Just `index` might be not so clear though, any other suggestions?

fully agree

```python
from docarray.vectordb import HnswDocumentIndex
``` ?

```python
from docarray.index import HnswDocumentIndex
``` ?

```python
from docarray.search import HnswDocumentIndex
``` ?

```python
from docarray.db import HnswDocumentIndex
``` ?

**Comment by jupyterjazz** [comment]
@samsja @JohannesMessner 
Let's go with
```python
from docarray.index import HnswDocumentIndex
```
for now and if it doesn't suit docarray we can refine it later. wdyt?

samsja (approved):

PR #1267, opened 2023-03-22
feat: add `get_paths()` instead of v1 `from_files()`

anna-charlotte:
Goals:

In docarray v1 we had the classmethod `.from_files()`.
For v2 we want to have something similar, but not to the same extend, but leave some part to the user.

Calling `.from_files()` would not be as straight forward in v2, since for example it is not clear which url field a file path should be assigned to.
Therefore we introduce only `get_paths()` which give you all the paths according to a pattern.

The part, where the content of a file is loaded, will be adressed in a separate PR.

```python
da = DocumentArray[MyDoc]([MyDoc(url=url) for url in get_paths(''./urls/*'')])
da.text = da.url.load(). # this would be nice, but not possible right now, since da.url is a list. Will be adressed in separate PR
```


samsja:
I think all we need is this `get_paths(patterns='*.*')` that returns a generator of url. Then it is up to the user to deal with it IMO. This changes the original behavior of `from_files` but this once just expects that URL is always defined in a Document which is not aligned with v2 philosophy anymore.


doing this might be interesting: `da.text = da.url.load()` but I think this will involve a bigger reflection on our type.

Basically, we could imagine that for each of our type we have a vector extension of it on which we could call the exact same function but it will do it on all of the element at once.

But this is bigger that this PR IMO.

I would go first with having this `get_paths` function. We can always bring back a `from_files` as a method later if we find out that it is better. The other way is not true if we commit to a design for `from_files` we cannot remove it since it is a method

JohannesMessner:
Hmm I see the problems. Maybe we have to stick to something closer to `from_files()` after all...
What about something like this?

```python
da = DocumentArray[MyDoc].from_files(url='./urls/*', other_url='./others/*')  # need to specify the fields, but as kwargs. Optional url fields do not need to be specified

da.text = da.url.load()  # we return a subclass of list (`UrlList`) that allows this
```

JohannesMessner:
> Basically, we could imagine that for each of our type we have a vector extension of it on which we could call the exact same function but it will do it on all of the element at once.

> But this is bigger that this PR IMO.

For this PR we only need it for URL, we can lazily extend the same idea to other types as we need it in the future

samsja:
Lets see in term of number of verbositiy

```python
da = DocumentArray[MyDoc].from_files(url='./urls/*')  
```

vs


```python
da = DocumentArray[MyDoc]([MyDoc(url=url) for url in get_path(''./urls/*'')])
```

JohannesMessner:
> Lets see in term of number of verbositiy
> 
> ```python
> da = DocumentArray[MyDoc].from_files(url='./urls/*')  
> ```
> 
> vs
> 
> ```python
> da = DocumentArray[MyDoc]([MyDoc(url=url) for url in get_path(''./urls/*'')])
> ```

actually i like the second option here, let's not re-invent too many wheels.
But we should still combine it with this, imo:

```python
da.text = da.url.load()  # we return a subclass of list (`UrlList`) that allows this
```

Otherwise it becomes a pain to load stuff

docarray/helper.py
```diff
@@ -150,3 +164,51 @@ def _is_lambda_or_partial_or_local_function(func: Callable[[Any], Any]) -> bool:
...
+
+
+def get_paths(
+    patterns: Union[str, List[str]],
+    recursive: bool = True,
+    size: Optional[int] = None,
+    exclude_regex: Optional[str] = None,
+) -> Generator[str, None, None]:
+    """
+    Yield file paths described by `patterns`.
```
samsja, line 176: can you provide a code example ?

docarray/helper.py
```diff
@@ -150,3 +164,51 @@ def _is_lambda_or_partial_or_local_function(func: Callable[[Any], Any]) -> bool:
...
+            raise ValueError(f'`{exclude_regex}` is not a valid regex.')
+
+    def _iter_file_extensions(ps):
+        return itertools.chain.from_iterable(
+            glob.iglob(os.path.expanduser(p), recursive=recursive) for p in ps
+        )
+
+    num_docs = 0
+    for file_path in _iter_file_extensions(patterns):
+        print(f"file_path = {file_path}")
```
samsja, line 206:
```suggestion
        print(f'file_path = {file_path}')
```

samsja (approved):

PR #1268, opened 2023-03-22
refactor(da): remove tensor type from `DocumentArray` init

samsja:
# Context

This pr removes the `tensor_type` argument from `DocumentArray` init and put it rather at the `da.stack(tensor_type=TorchTensor)` level.

Indeed this argument is only relevant for `DocumentArrayStacked` and was used inside `DocumentArray` only when the stack method was called

### DIFF
--- a/docarray/array/stacked/array_stacked.py
+++ b/docarray/array/stacked/array_stacked.py
@@ -323,7 +323,9 @@ class DocumentArrayStacked(AnyDocumentArray[T_doc]):
                     f'{value} schema : {value.document_type} is not compatible with '
                     f'this DocumentArrayStacked schema : {self.document_type}'
                 )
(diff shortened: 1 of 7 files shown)


docarray/array/stacked/array_stacked.py
```diff
@@ -82,8 +82,9 @@ class DocumentArrayStacked(AnyDocumentArray[T_doc]):
      numpy/PyTorch.
 
     :param docs: a DocumentArray
-    :param tensor_type: Class used to wrap the stacked tensors
-
+    :param tensor_type: Tensor Class used to wrap the stacked tensors. This is usefull
```
jupyterjazz, line 85: another one here 😄 

```suggestion
    :param tensor_type: Tensor Class used to wrap the stacked tensors. This is useful
```

jupyterjazz (approved):

PR #1270, opened 2023-03-22
fix: bytes type in `TextDoc` and `VideoDoc`

anna-charlotte:
There is a bug related to the type of the `bytes` field in the predefined Documents, such as TextDoc.
The bytes field is of type `Optional[bytes]`, but when looking at `TextDoc.__fields__['bytes'].type_` it equals 'None' instead of `Optional[bytes]`

Because of this, the following fails:
```python
from docarray.documents import TextDoc, VideoDoc

da = DocumentArray[TextDoc]([TextDoc()])
for name, value in TextDoc.__fields__.items():
    print(f"name, type = {name, value.type_}")
da.summary()  # fails

da = DocumentArray[VideoDoc]([VideoDoc()])
da.summary()  # fails
```

```
name, type = ('id', <class 'docarray.typing.id.ID'>)
name, type = ('text', <class 'str'>)
name, type = ('url', <class 'docarray.typing.url.text_url.TextUrl'>)
name, type = ('embedding', typing.Union[docarray.typing.tensor.embedding.ndarray.NdArrayEmbedding, docarray.typing.tensor.embedding.torch.TorchEmbedding, docarray.typing.tensor.embedding.tensorflow.TensorFlowEmbedding, NoneType])
name, type = ('bytes', <class 'NoneType'>)  # <----------------------- NoneType instead of bytes!

╭─────── DocumentArray Summary ────────╮
│                                      │
│   Type     DocumentArray[VideoDoc]   │
│   Length   1                         │
│                                      │
╰──────────────────────────────────────╯
  File "/Users/charlottegerhaher/Library/Application Support/JetBrains/PyCharm2022.2/scratches/scratch_71.py", line 66, in <module>
    da.summary()
  File "/Users/charlottegerhaher/Desktop/jina-ai/docarray_v2/docarray/docarray/array/abstract_array.py", line 276, in summary
    DocumentArraySummary(self).summary()
  File "/Users/charlottegerhaher/Desktop/jina-ai/docarray_v2/docarray/docarray/display/document_array_summary.py", line 54, in summary
    self.da.document_type.schema_summary()
  File "/Users/charlottegerhaher/Desktop/jina-ai/docarray_v2/docarray/docarray/base_document/document.py", line 71, in schema_summary
    DocumentSummary.schema_summary(cls)
  File "/Users/charlottegerhaher/Desktop/jina-ai/docarray_v2/docarray/docarray/display/document_summary.py", line 41, in schema_summary
    DocumentSummary._get_schema(cls),
  File "/Users/charlottegerhaher/Desktop/jina-ai/docarray_v2/docarray/docarray/display/document_summary.py", line 76, in _get_schema
    for arg in field_type.__args__:
AttributeError: type object 'NoneType' has no attribute '__args__'

```

This can be solved by removing the ` = None` default for `TextDoc` and `VideoDoc`. Since they are optional, there default stays None, even without explicitly stating it.

I removed all other `= None` defaults to to keep it coherent, since sometimes this was explicitly stated, sometimes not.

### DIFF
--- a/docarray/documents/text.py
+++ b/docarray/documents/text.py
@@ -98,9 +98,9 @@ class TextDoc(BaseDocument):
         doc == doc2  # False, their ids are not equivalent
     """
 
-    text: Optional[str] = None
-    url: Optional[TextUrl] = None
-    embedding: Optional[AnyEmbedding] = None
+    text: Optional[str]
+    url: Optional[TextUrl]
+    embedding: Optional[AnyEmbedding]
     bytes: Optional[bytes]
 
     def __init__(self, text: Optional[str] = None, **kwargs):
--- a/docarray/documents/text.py
+++ b/docarray/documents/text.py
@@ -43,6 +43,7 @@ class TextDoc(BaseDocument):
         from docarray.typing import AnyEmbedding
         from typing import Optional
 
+
         # extend it
         class MyText(Text):
             second_embedding: Optional[AnyEmbedding]
@@ -61,6 +62,7 @@ class TextDoc(BaseDocument):
 
         from docarray import BaseDocument
         from docarray.documents import ImageDoc, TextDoc
+
 
         # compose it
         class MultiModalDoc(BaseDocument):
@@ -99,7 +101,7 @@ class TextDoc(BaseDocument):
     text: Optional[str] = None
     url: Optional[TextUrl] = None
     embedding: Optional[AnyEmbedding] = None
-    bytes: Optional[bytes] = None
+    bytes: Optional[bytes]
 
     def __init__(self, text: Optional[str] = None, **kwargs):
         if 'text' not in kwargs:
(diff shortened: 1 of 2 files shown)


jupyterjazz (commented):
Good catch! But what I don't understand is why setting to `None` would change the type and  make `TextDoc.__fields__['bytes'].type_` return `None`? Seems like an unexpected behavior to me

samsja (approved):

samsja:
I think `type_` is not safe you should use `outer_type` rather. Btw we have a `_get_field_type` method that you should use

jupyterjazz:
haha yes, we can handle it internally, but there's a real chance developers also use `Optional[bytes] = None` in their dataclasses which as we know would not work, so maybe it's worth further investigation?

samsja:
Then lets rename everything to not bytes. This is a python reserved keyword anyway I believe.

samsja:
and feel free to still open an issue on pydantic, it might still be a bug in pydantic code so better for them to be aware of it

PR #1276, opened 2023-03-22
feat: add magic typed list 

samsja:
# Context

allow to do this 

```python

from docarray.array.array.typed_sequence import TypedList

class A:    
    def __init__(self, param = None):
        self.param = param
    
    def f(self):
        return None
    
    def g(self):
        return 'is this really working ?'

data = TypedList[A]([A(param=1) for _ in range(3)])

data.f()
>>> [None, None, None]

data.g()

>>> ['is this really working ?', 'is this really working ?', 'is this really working ?']

data.param

>>> [1, 1, 1]

```
### Intended usecase

```python
da.tensor = da.url.load()

# or

da.bytes_ = da.tensor.to_bytes()


# instead of 

da.tensor = [item for item in da.url.load()]

```

### DIFF
--- a/tests/integrations/array/test_magic_list.py
+++ b/tests/integrations/array/test_magic_list.py
@@ -1,0 +1,22 @@
+import numpy as np
+import pytest
+
+from docarray import DocumentArray
+from docarray.documents import ImageDoc
+
+REMOTE_JPG = (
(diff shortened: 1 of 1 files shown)


samsja:
closing. This feature will be implemented later

Issue #1288, opened 2023-03-24
DocIndex: Fix flatten_schema


JohannesMessner:
Bug described here: https://github.com/docarray/docarray/pull/1277#issuecomment-1482280565
Ther are additional things to consider, though. The fix described in the PR woks, but as a side effect (?) it allows union types in the schema definition, which is not desired.

JohannesMessner:
While working on this issue i notices that it makes it so that union types in the index schema are not caught.

So, the problem with having union types in the index schema is the following:

Say we have a schema like this:
```python
class MyDoc(BaseDocument):
    tensor: Union[NdArray, str]
    
index = MyDocIndex[MyDoc](...)  # should fail
```
Then what what column type should the backend choose for this? It is not well defined, thus should not be allowed.

However, this would preclude pre-built docs to be used as schemas.
**Note** that union types can still be used as the _schema of the data_, so the following workaround would work:

```python
class MyDoc(BaseDocument):
    tensor: Union[NdArray, str]
    
class MySchema(BaseDocument):
    tensor: NdArray
    
index = MyDocIndex[MySchema](...)  # works
index.index([MyDoc(tensor=np.random.rand(...)) for _ in range(10)])  # also works
```

So we have two options here, as I see it:
1. as suggested above, disallow all union types _in the index schema_, and raise an Exception. This would require the workaround shown below, even for built-in docs such as `ImageDoc`. It is the clearest "rule", but annoying to the user.
2. do this, but have an exception: Union types that only contain tensor-like stuff should be allowed, since they all map to a tensor type in the backend column anyways. So, `Union[NdArray, TorchTensor]` would be allowed in the index schema, while `Union, NdArray, str]` would not be. We even have a helper `_is_tensor_union(..)` somewhere for this. This would allow us to use pre-built docs like `ImageDoc` to be used as a schema, which would be nice; but it makes the "rules" a bit less clear.

Personally I think I would go for the second option, but maybe as a first iteration the the first option works. @AnneYang720 up to you!

PR #1294, opened 2023-03-27
fix: flatten schema of abstract index

AnneYang720:
Goals:

issue #1288 

- [x] change `_flatten_schema`
- [x] change `_create_column_infos` 
- [x] other possible changes: `unwrap_optional_type`, `_convert_dict_to_doc`
- [x] tests
- [x] doc

The schema allows "basic" types, simple optionals and union of tensor types. And `_flatten_schema` will return `AbstractTensor` for all correct tensor related types. Here are the examples of the convertion that `_flatten_schema` does:

`Optional[str] -> str`
`Union[str, None] -> str`
`Optional[NdArray] -> AbstractTensor`
`Optional[ImageTorchTensor, str] -> ValueError`
`Union[ImageTorchTensor, ImageTensorFlowTensor] -> AbstractTensor`
`Union[ImageTorchTensor, str] -> ValueError`

After the schema is built, anything is allowed as long as the data itself has the right types, but choosing a properly mathing schema can have better performance. The following example work:
```python
class MySchema(BaseDoc):
    tensor: NdArray[10]

store = MyDocIndex[MySchema]()

class MyDoc(BaseDocument):
    tensor: Union[NdArray, str]

store.index([MyDoc(tensor=np.random.rand(10)) for _ in range(10)])
```


docs/how_to/add_doc_index.md
```diff
@@ -153,6 +153,9 @@ By default, it holds that `_ColumnInfo.docarray_type == self.python_type_to_db_t
 However, you should not rely on this, because a user can manually specify a different db_type. 
 Therefore, your implementation should rely on `_ColumnInfo.db_type` and not directly call `python_type_to_db_type()`.
 
+**Caution**
+`AbstractTensor` will be the `_ColumnInfo.docarray_type` if the field in `self._schema` is a subclass of `AbstractTensor` or a tensor Union.
+
```
JohannesMessner, line 158: I would rephrase this a little bit:

```suggestion
If a subclass of `AbstractTensor` appears in the Document Index's schema (i.e. `TorchTensor`, `NdArray`, or `TensorFlowTensor`), then `_ColumnInfo.docarray_type` will simply show `AbstractTensor` instead of the specific subclass. This is because the abstract class normalizes all input data of type `AbstractTensor` to `np.ndarray` anyways, which should make your life easier. Just be sure to properly handle `AbstractTensor` as a possible value or `_ColumnInfo.docarray_type`, and you won't have to worry about the differences between torch, tf, and np.

```

JohannesMessner (changes requested):
Just one small suggestion :)

Also, could you update the PR description to show the "rules" of the schema?

Stuff like: The index schema allows "basic" types and simple optionals, simple optionals are treated like basic types, unions are not allowed, unless they are a tensor union; in the schema of the data anything is allowed as long as the data itself has the right types, but choosing a properly mathing schema can have better performance; stuff like this and other edge cases i am missing here.

This will help us later when we are writing documentation about these behaviours, otherwise we will forget about it

JohannesMessner (approved):

PR #1297, opened 2023-03-28
fix: rename DocArrayProto to DocumentArrayProto

samsja:
# Context

* rename `DocArrayProto` to `DocumentArrayProto` to fix jina problem


JoanFM (approved):

jupyterjazz (commented):
what was the problem?

Issue #1234, opened 2023-03-14
parallel batching for `find()`


JoanFM:

prajakta-1527:
Hii @JoanFM, I would like to solve this issue. Can you guide me regarding details of the issue?

Animesh312:
Hii @JoanFM I would like to solve this issue

Puriva:
i would the like to solve the issue

JohannesMessner:
Hi @prajakta-1527 , since you were the first to ask, you can take the issue :) I updated the description, I hope it is clear. If not, don't hesitate to reach out!

PR #1302, opened 2023-03-29
Modified find function to enable parallel batching

prajakta-1527:
Fixes [#1234](https://github.com/docarray/docarray/issues/1234)
Tasks completed :
- [x] modified find and find_batched function
- [x] defined function to be passed to map_docs_batch()
- [x] leveraged map_docs_batch() function to enable parallel batching
- [x] handled case of query input as AnyTensor
- [x] handled negative batch_size input
- [x] added params description in docstring
- [x] added appropriate examples in docstrings


prajakta-1527:
Hi @JoanFM @JohannesMessner 
The approach for the above issue is to batch the input queries and pass them to the map_docs_batch() function for parallel computation of the result. If the batch size is not provided, the computation will proceed as before (without using multiprocessing pools). If the query input is a tensor, it is first passed as the embedding parameter to a docarray of size the same as the count of input queries. Then they are batched and processed as mentioned above. Is this the right way to proceed, or are some changes required? Thanks.

docarray/utils/find.py
```diff
@@ -176,34 +188,94 @@ class MyDocument(BaseDoc):
...
+        else:
+            batch_size = int(batch_size)
+    else:
+        top_indices, top_scores = _get_result(query)
+        res = []
+        for indices_per_query, scores_per_query in zip(top_indices, top_scores):
+            docs_per_query: DocList = DocList([])
+            for idx in indices_per_query:  # workaround until #930 is fixed
+                docs_per_query.append(index[idx])
+            docs_per_query = DocList(docs_per_query)
```
samsja, line 250: not sure that this line is needed

docarray/utils/find.py
```diff
@@ -110,10 +115,16 @@ def find_batched(
     index: AnyDocArray,
     query: Union[AnyTensor, DocList],
     embedding_field: str = 'embedding',
+    batch_size: Optional[int] = None,
     metric: str = 'cosine_sim',
     limit: int = 10,
     device: Optional[str] = None,
-    descending: Optional[bool] = None,
+    descending: bool = False,
```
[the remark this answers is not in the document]
prajakta-1527: Yes, I agree, the default value for descending must be None, it generates mypy error further down in line 232:
```(parameter) descending: bool
descending: sort the results in descending order.

Argument "descending" to "top_k" of "Retrieval" has incompatible type "Optional[bool]"; expected "bool"  [arg-type]mypy(error)
```
I found out that adding cast function in the _get_result function solves this error.

docarray/utils/find.py
```diff
@@ -176,34 +188,94 @@ class MyDocument(BaseDoc):
...
+        batch_size=batch_size,
+        backend=backend,
+        num_worker=num_worker,
+        shuffle=shuffle,
+        pool=pool,
+        show_progress=show_progress,
+    )
+
+    for indices_per_query, scores_per_query in it:
+        docs_per_query: DocList = DocList([])  # type: ignore
```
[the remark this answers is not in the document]
prajakta-1527: Yes, its not needed. It was added because mypy gave error for redefining the variable docs_per_query. It is much better to rename it.

docarray/utils/find.py
prajakta-1527: Are you refering to the comment "workaround until #930 is fixed" in line 248?

prajakta-1527:
> Looks great! Added some small comments. Some tests are failing due to the changed `descending` default I think.
> Could you also add some tests with the new parameters? So that we can make sure everything works as expected with the batchsize and parallelism for instance : )

Thanks for reviewing! I would add the required tests with the new parameters.

prajakta-1527:
> Or we could just add an optional param to map_docs_batched: `func_args: Optional[Dict[str, Any]]`

Yes, but that wont allow us us to pass the `func_args` to the function which is to be mapped because curently we are using `pool.map()` whose inputs are target function and an iterable while the `pool.apply()` additionally supports multiple arguments to the target function.

docarray/utils/find.py
```diff
@@ -23,6 +26,10 @@ class _FindResult(NamedTuple):
     scores: AnyTensor
 
 
+class Doc(BaseDoc):
```
JoanFM, line 29: why is this definition needed?

JoanFM (changes requested):
Hey @prajakta-1527 thanks for the contribution, would it be possible for you to add tests to this behavior?

docarray/utils/find.py
prajakta-1527: This is used down in line 278. When the query data is in the form of AnyTensor, for incorporating parallelism, the queries are separated and then passed in a batch. The Doc class is created to store these individual queries.

prajakta-1527:
> Hey @prajakta-1527 thanks for the contribution, would it be possible for you to add tests to this behavior?

I had modified the `test_find.py` file to ensure parallel batching works properly. Do I have to add more tests to it?

docarray/utils/find.py
JoanFM: can you create another class with a private more descriptive name? Doc seems too generic

prajakta-1527: yes, I will change it.

Issue #1264, opened 2023-03-21
DocIndex: Validate if `search_field` is valid


JohannesMessner:
When a user passes a `search_field` we should check in the abstract class if it correspons to one of the columns that was parsed from the schema. That we the backend implementer does not have to worry about it, and we can give a uniform error message.

srini047:
@JohannesMessner I would like to take up the issue.

srini047:
Update on the issue🔔

I feel to have found the right file where the code needs to be added i.e. [here](https://github.com/docarray/docarray/blob/feat-rewrite-v2/docarray/index/abstract.py)
And I have a brief idea about what code needs to be added as well. That goes like this:

```
class DocIndex:
    def __init__(self, schema):
        self.columns = [column['name'] for column in schema['columns']]

    def validate_search_field(self, search_field):
        if search_field in self.columns:
            return True
        else:
            raise ValueError(f{search_field} is not a valid search field. Valid search fields are: {', '.join(self.columns)})
```

Though this is not the actual code this has to be the logic. But I cannot rightly insert the code.
Any input is appreciated.

P.S.: This is my 1st contribution to this project.

JohannesMessner:
Hey @srini047, sorry for the delayed response!

Your logic looks about right!
The only thing is that `__init__()` doesn't actually accept a schema, but the abstract class already stores `self._column_infos`.
So your code could look something like this:

```python
class DocIndex:
    def validate_search_field(self, search_field):
        if search_field in self._column_infos.keys():
            return True
        else:
            raise ValueError(f{search_field} is not a valid search field. Valid search fields are: {', '.join(self._column_infos.keys())})
```
Don't hesitate to open a (draft) PR for this, even before you are completely done!

srini047:
> Hey @srini047, sorry for the delayed response!
> 
> Your logic looks about right! The only thing is that `__init__()` doesn't actually accept a schema, but the abstract class already stores `self._column_infos`. So your code could look something like this:
> 
> ```python
> class DocIndex:
>     def validate_search_field(self, search_field):
>         if search_field in self._column_infos.keys():
>             return True
>         else:
>             raise ValueError(f{search_field} is not a valid search field. Valid search fields are: {', '.join(self._column_infos.keys())})
> ```
> 
> Don't hesitate to open a (draft) PR for this, even before you are completely done!

Thanks for such a humble reply. I'm in the process of making a draft PR.

PR #1319, opened 2023-03-30
feat: add validate search field method

srini047:
Fixes: #1264 

Goals:

- Add a validate search field method to check if the column name matches with the schema.
- Ultimate aim is to provide a uniform error message.


JohannesMessner:
We should have an excpetion for search fields that are empty (`''`) or `None`. Some backends, like weaviate, don't want a search field and should not fail.

srini047:
> We should have a special case for search fields that are empty (`''`) or `None`. Some backends, like weaviate, don't want a search field and should not fail.

@JohannesMessner Mind creating a new issue just in case it's a priority so I can fix work on it?

jupyterjazz:
@srini047 I took over your PR, thanks for the contribution

PR #1320, opened 2023-03-30
feat: docarray fastapi simple integration

jupyterjazz:
Presenting a simple way to use docarray with FastAPI as other approaches turned out to be more complicated.

```python
    # Create a docarray
    docs = DocArray[ImageDoc]([ImageDoc(tensor=np.zeros((3, 224, 224)))])

    app = FastAPI()
    # Always use our custom response class (needed to dump tensors)
    @app.post("/doc/", response_class=DocArrayResponse)
    async def func(fastapi_docs: List[ImageDoc]) -> List[ImageDoc]:
        # The docs FastAPI will receive will be treated as List[ImageDoc]
        # so you need to cast it to DocArray
        docarray_docs = DocArray[ImageDoc].construct(fastapi_docs)
        ...
        # In case you want to return a DocArray, return it as a list
        return list(docarray_docs)

    async with AsyncClient(app=app, base_url="http://test") as ac:
        response = await ac.post("/doc/", data=docs.to_json()) # sending docs as json

    assert response.status_code == 200
    # You can read FastAPI's response in the following way
    docs = DocArray[ImageDoc].from_json(response.content.decode())
```

Basically, FastAPI will treat our request as a `List[DocType]` so you'll have to cast it to `DocArray` inside the function, and in case you want to return your `DocArray`, return it as a `List` and read it with `from_json`.

We can definitely improve this procedure. I'll create a followup ticket and write down all the learnings I got from experimenting with different approaches this week. 


docarray/array/array/io.py
```diff
@@ -311,21 +310,22 @@ def to_base64(
...
         json_docs = json.loads(file)
-        return cls([cls.document_type.parse_raw(v) for v in json_docs])
+        return cls([cls.document_type(**v) for v in json_docs])
 
     def to_json(self) -> str:
         """Convert the object into a JSON string. Can be loaded via :meth:`.from_json`.
         :return: JSON serialization of DocArray
         """
-        return json.dumps([doc.json() for doc in self])
+        doc_jsons = ', '.join([doc.json() for doc in self])
```
samsja, line 327: I think here we could just use orson dump on `self._data`

docarray/base_doc/doc.py
```diff
@@ -28,7 +29,7 @@ class BaseDoc(BaseModel, IOMixin, UpdateMixin, BaseNode):
     class Config:
         json_loads = orjson.loads
         json_dumps = orjson_dumps_and_decode
-        json_encoders = {dict: orjson_dumps}
+        json_encoders = {AbstractTensor: lambda x: x}
```
samsja, line 32: can u add a comment explaning why the `lambda x: x` is needed here ?

Smth like: 

Our custom orjson response is able to handle tensor by itself, therefore we don't need FastAPI to do any transformation to the tensor object but just gave it to the response class

samsja (changes requested):
minor comment

docarray/array/array/io.py
```diff
@@ -311,21 +310,22 @@ def to_base64(
...
+        file: str,
     ) -> T:
         """Deserialize JSON strings or bytes into a DocArray.
 
         :param file: JSON object from where to deserialize a DocArray
         :return: the deserialized DocArray
         """
         json_docs = json.loads(file)
-        return cls([cls.document_type.parse_raw(v) for v in json_docs])
+        return cls([cls.document_type(**v) for v in json_docs])
```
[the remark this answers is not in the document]
jupyterjazz: `parse_raw` was expecting a string but now it receives a dict.

This is because the way we were doing json encoding -> '["{text: smth}"]' so `parse_raw` was taking "{text: smth}"

Now it would be '[{text: smth}]' so we pass a dict instead of a string.

docarray/array/array/io.py
jupyterjazz: I don't see how. Then docs are treated as strings (double-encoded)
Would be nice if you could specify custom encoders for specific types, like 
```python
orjson.dumps(docs, custom_encoders={BaseDoc: lambda x: x.json()}
```
but i don't think that's possible

docarray/array/array/io.py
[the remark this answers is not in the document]
samsja: okay great

docarray/array/array/io.py
samsja: okay I see. I think this is feasible with the default thing

docarray/array/array/io.py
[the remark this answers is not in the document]
samsja: yes my bad you are right

docarray/array/array/io.py
jupyterjazz: Ok so this also works but not sure if it's better because we're not using pydantic's json() method anymore.

```python

class CustomEncoder(json.JSONEncoder):
    def default(self, val):
        if isinstance(val, BaseDoc):
            return val.dict()

def to_json(self) -> str:
    return json.dumps([doc for doc in self], cls=CustomEncoder)
```

wdyt?

samsja: This way is better performance wise. We delegate everything to orjson. Otherwsie the `b''.join` need to copy paste everything into the new bytes. And this is expensive

docarray/array/array/io.py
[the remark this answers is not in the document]
samsja: but should not we use orjson load rather than json.load ? Otherwise tensor loading will be problematic or slow

docarray/array/array/io.py
jupyterjazz: We're not doing bytes here.

Basically it's between these two:

```python
def to_json(self) -> str:
    doc_jsons = ', '.join([doc.json() for doc in self])
    return f'[{doc_jsons}]'
```
```python
class CustomEncoder(json.JSONEncoder):
    def default(self, val):
        if isinstance(val, BaseDoc):
            return val.dict()

def to_json(self) -> str:
    return json.dumps([doc for doc in self], cls=CustomEncoder)
```

Performance-wise I think it's the same and we should go with a cleaner approach which to me is the second one but i don't really have a strong opinion here

samsja: I think performance wise the first one is worse. We should rely as much on orsjon as possible

samsja: join is super slow

docarray/base_doc/base_node.py
```diff
@@ -20,3 +24,14 @@ def _to_node_protobuf(self) -> 'NodeProto':
...
+    @classmethod
+    @abstractmethod
+    def from_protobuf(cls: Type[T], pb_msg: T) -> T:
+        ...
+
+    def _docarray_to_json_compatible(self):
+        """
+        Convert itself into a json compatible object
+        """
+        ...
```
samsja, line 37: should not it still return self by default ?

samsja: or maybe not

jupyterjazz: I think it's better to leave this abstract and define for each class, otherwise we say self by default by override it in both IOmixin and BaseDoc class

PR #1321, opened 2023-03-30
feat: qdrant document index

kacperlukawski:
Goals:

This PR implements Qdrant as a document index.

- [x] init and build index
- [x] find
- [x] find batched
- [x] filter
- [x] filter batched
- [x] text search
- [x] text search batched
- [x] num docs
- [x] get items
- [x] delete items
- [x] query builder
- [x] execute query


samsja:
:rocket: :rocket: :rocket: :rocket:

kacperlukawski:
@JohannesMessner I have some doubts regarding the batched methods. Things are obvious for `_find_batched`, as any two documents will always be similar. But should we return scores as a single `np.array` in the case of `_text_search_batched` or `_filter_batched`? Each query may have a different number of results. Wouldn't it be better to return a list of `np.array` instances instead? So each of them might be of a different shape. WDYT?

JohannesMessner:
> @JohannesMessner I have some doubts regarding the batched methods. Things are obvious for `_find_batched`, as any two documents will always be similar. But should we return scores as a single `np.array` in the case of `_text_search_batched` or `_filter_batched`? Each query may have a different number of results. Wouldn't it be better to return a list of `np.array` instances instead? So each of them might be of a different shape. WDYT?

This is a good point, you are right. Feel free to do it this way and change the type hint as needed! If this change requires some cleanup/adjustment in one of the other backends we can take care of that afterwards

docarray/index/backends/qdrant.py
```diff
@@ -0,0 +1,318 @@
...
+        quantization_config: Optional[types.QuantizationConfig] = None
+
+    @dataclass
+    class RuntimeConfig(BaseDocIndex.RuntimeConfig):
+        default_column_config: Dict[Type, Dict[str, Any]] = field(
+            default_factory=lambda: {
+                'id': {},
+                'vector': {},
+                'payload': {},
+                np.ndarray: {},
```
JohannesMessner, line 97: Is `np.ndarray` really needed here?

JohannesMessner (changes requested):
Looking good!
I would just like to see a few more tests:
- Testing the configuration, especially the users changing a configuration for a specific column. For example, they should be able to change the column type in the DB, by passing a `col_type` to `Field()`. A possible scenario would be that someone wants a vector to not be indexed, so they want to store it as `payload` instead of `vector`. E.g.: https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/integrations/doc_index/elastic/v7/test_column_config.py
- Something with tensorflow (sorry if I missed it), this should work out of the box, but let's make sure that there arent' any issues. E.g.: https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/integrations/doc_index/elastic/v7/test_find.py#L140
- Something that uses our built-in Documents, they have a lot of unions and optionals in their schema, so let's just double check that everything is good there. E.g.: https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/integrations/doc_index/elastic/v7/test_index_get_del.py#L270

kacperlukawski:
@JohannesMessner Thanks for the comments - I'll implement the changes soon! I just added the `QueryBuilder`, so functional-wise, we're done.

docarray/index/backends/qdrant.py
kacperlukawski: I reused the following test from hnswlib:

```
def test_schema_with_user_defined_mapping(tmp_path):
    class MyDoc(BaseDoc):
        tens: NdArray[10] = Field(dim=1000, col_type=np.ndarray)

    store = HnswDocumentIndex[MyDoc](work_dir=str(tmp_path))
    assert store._column_infos['tens'].db_type == np.ndarray
```

When the `col_type` is defined for a field, it's passed directly without calling `python_type_to_db_type`. I wondered how to overcome it but couldn't find a better way. How should I handle that?

kacperlukawski:
>     * Testing the configuration, especially the users changing a configuration for a specific column. For example, they should be able to change the column type in the DB, by passing a `col_type` to `Field()`. A possible scenario would be that someone wants a vector to not be indexed, so they want to store it as `payload` instead of `vector`. E.g.: https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/integrations/doc_index/elastic/v7/test_column_config.py

Qdrant is schemaless, so it doesn't really matter what `col_type` will be defined. We store the data as JSON, so that makes no difference. I'll review the ES7 approach, but I'm not sure if that applies to Qdrant as well.
 
>     * Something with tensorflow (sorry if I missed it), this should work out of the box, but let's make sure that there arent' any issues. E.g.: https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/integrations/doc_index/elastic/v7/test_find.py#L140

No, it was not tested yet, but I'll extend the suite. I just didn't find Tensorflow in the dependencies. Shouldn't it be defined in `pyproject.toml` as well?

>     * Something that uses our built-in Documents, they have a lot of unions and optionals in their schema, so let's just double check that everything is good there. E.g.: https://github.com/docarray/docarray/blob/feat-rewrite-v2/tests/integrations/doc_index/elastic/v7/test_index_get_del.py#L270

Sure, I'll cover those scenarios as well. Thanks for the references!

docarray/index/backends/qdrant.py
JohannesMessner: I believe passing `ndarray` as a column type only really makes sense for hnswlib, not for Qdrant. The user should pass a column type that is a valid type in Qdrant, i.e. one of the ones that you define in your mapping ('vector', 'payload', etc.). No special handling should be necessary

JohannesMessner:
>No, it was not tested yet, but I'll extend the suite. I just didn't find Tensorflow in the dependencies. Shouldn't it be defined in pyproject.toml as well?

There are some issues with tf and the protobuf version that it is pinned to, so we don't have it in our toml file, and no need to add it. you can just add the tests and mark them as tensorflow using the pytest marker as shown in the example, then our CI will install tf and run the test appropriately.

docarray/index/backends/qdrant.py
```diff
@@ -0,0 +1,478 @@
...
+            return QdrantDocumentIndex.QueryBuilder(
+                vector_search_field=self._vector_search_field,
+                vector_filters=self._vector_filters,
+                payload_filters=self._payload_filters,
+                text_search_filters=self._text_search_filters + [(search_field, query)],
+            )
+
+        find_batched = _raise_not_composable('find_batched')
+        filter_batched = _raise_not_composable('find_batched')
+        text_search_batched = _raise_not_composable('text_search')
```
JohannesMessner, line 163:
```suggestion
        filter_batched = _raise_not_composable('filter_batched')
        text_search_batched = _raise_not_composable('text_search_batched')
```

docarray/index/backends/qdrant.py
```diff
@@ -0,0 +1,478 @@
...
+                    ),
+                ],
+            ),
+            limit=len(doc_ids),
+            with_payload=True,
+            with_vectors=True,
+        )
+        return [self._convert_to_doc(point) for point in response]
+
+    def execute_query(self, query: Query, *args, **kwargs) -> DocList:
```
JohannesMessner, line 279: here the users should also be able to pass in a raw query that does not come from our query builder, i.e. a python dict or string like this (copied from your docs):

```python
{
    "filter": {
        "must": [
            {
                "key": "city",
                "match": {
                    "value": "London"
                }
            }
        ]
    },
    "params": {
        "hnsw_ef": 128,
        "exact": false
    },
    "vector": [0.2, 0.1, 0.9, 0.7],
    "limit": 3
}
```

This is meant as a fallback option for any functionality that may not be covered by the DocArray API, we don't want to lock away Qdrant functionality from our users. Is that possible without too much hassle?

kacperlukawski: I'm still working on that one - I'd like to avoid sending raw HTTP requests, and we need to do some conversion.

docarray/index/backends/qdrant.py
kacperlukawski: Done, the np.array is not here anymore. The three types are left though, as we need to differentiate the behaviour for ids, vectors and metadata attributes (those are stored differently).

docarray/index/backends/qdrant.py
kacperlukawski: @JohannesMessner This is already implemented, so raw queries might be passed to `.execute_query`

kacperlukawski:
Should I reformat the files with black and commit? Those don't seem related, as none of the problematic files was changed in that PR.

samsja:
> Those don't seem related

you can do `black -s`

kacperlukawski:
@samsja @JohannesMessner Already one - I'd be grateful for a review ;)

JohannesMessner:
> @samsja @JohannesMessner Already done - I'd be grateful for a review ;)

Awesome! I think we were in a good place already, but making another pass now

samsja:
lgtm. I will let @JohannesMessner do the final approve as I was less involved on this PR

samsja:
But good job ! :rocket:

JohannesMessner (approved):
Nice job, thanks a ton for your contribution!! ❤️

PR #1325, opened 2023-04-02
Ignore tests which has dependency on tensorflow if tensorflow is missing

agaraman0:
Goals:

- ignore tests which are using tensorflow but tensorflow is missing from device ignore tests instead of tests getting failed


samsja:
You can just do `pytest -m "not tensorflow"` when you run the test. No need to mark them skip. This is what we are doing in CI

agaraman0:
Thanks for highlighting, was not aware about this. this is helpful

samsja:
no worries, thanks for contributing!

PR #1335, opened 2023-04-04
feat: save image tensor to file

anna-charlotte:
Goals:
Save an image tensor to file, such as `.jpg` or `.png`

### DIFF
--- a/docarray/typing/tensor/image/abstract_image_tensor.py
+++ b/docarray/typing/tensor/image/abstract_image_tensor.py
@@ -1,14 +1,11 @@
 import io
 import warnings
 from abc import ABC
-from typing import TypeVar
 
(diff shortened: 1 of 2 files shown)


docarray/typing/tensor/image/abstract_image_tensor.py
```diff
@@ -31,6 +36,22 @@ def to_bytes(self, format: str = 'PNG') -> bytes:
 
         return img_byte_arr
 
+    def save(self: 'T', file_path: str) -> None:
```
samsja, line 39: I don't think you need the T here. Self is always from the class
```suggestion
    def save(self, file_path: str) -> None:
```

samsja (approved):

PR #1339, opened 2023-04-05
refactor: change return type of find batched

### LINKED ISSUE #1310: change return type of `find_batched()` utility function
Currently, `find_batched()` has the following signature:

```python
def find_batched(
    ...
) -> List[FindResult]:
```

In order to align with `DocumentIndex`, it should be switched to:

```python
def find_batched(
    ...
) -> FindResultBatched:
```

**TODOs:**
- move the definition of `FindResultBatched` from `index/abstract.py` to `utils/find.py`
- change the signature as described above
- fix the imports on all Document Index classes
- Probably `FindResultBatched` needs a tweak as well: `scores: np.ndarray` probably needs to be changed to allow a union of ndarray, torch.Tensor and tf.Tensor

npitsillos:
Goals:

- Closes #1310

### DIFF
--- a/docarray/index/backends/elastic.py
+++ b/docarray/index/backends/elastic.py
@@ -28,17 +28,12 @@ from pydantic import parse_obj_as
 
 import docarray.typing
 from docarray import BaseDoc
-from docarray.index.abstract import (
-    BaseDocIndex,
-    _ColumnInfo,
-    _FindResultBatched,
-    _raise_not_composable,
-)
+from docarray.index.abstract import BaseDocIndex, _ColumnInfo, _raise_not_composable
 from docarray.typing import AnyTensor
 from docarray.typing.tensor.abstract_tensor import AbstractTensor
 from docarray.typing.tensor.ndarray import NdArray
 from docarray.utils._internal.misc import is_tf_available, is_torch_available
-from docarray.utils.find import _FindResult
+from docarray.utils.find import _FindResult, _FindResultBatched
 
 TSchema = TypeVar('TSchema', bound=BaseDoc)
 T = TypeVar('T', bound='ElasticDocIndex')
--- a/docarray/index/backends/elastic.py
+++ b/docarray/index/backends/elastic.py
@@ -440,7 +440,7 @@ class ElasticDocIndex(BaseDocIndex, Generic[TSchema]):
         das, scores = zip(
             *[self._format_response(resp) for resp in responses['responses']]
         )
-        return _FindResultBatched(documents=list(das), scores=np.array(scores))
+        return _FindResultBatched(documents=list(das), scores=scores)
 
     ###############################################
     # Helpers                                     #
@@ -537,7 +537,7 @@ class ElasticDocIndex(BaseDocIndex, Generic[TSchema]):
             docs.append(doc_dict)
             scores.append(result['_score'])
 
-        return docs, parse_obj_as(NdArray, scores)
+        return docs, [parse_obj_as(NdArray, s) for s in scores]
 
     def _refresh(self, index_name: str):
         self._client.indices.refresh(index=index_name)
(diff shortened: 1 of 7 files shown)


docarray/utils/find.py
```diff
@@ -23,6 +23,16 @@ class _FindResult(NamedTuple):
...
 
 
+class FindResultBatched(NamedTuple):
+    documents: List[DocList]
+    scores: AnyTensor
+
+
+class _FindResultBatched(NamedTuple):
+    documents: Union[List[DocList], List[List[Dict[str, Any]]]]
+    scores: AnyTensor
```
samsja, line 33: where do we use this class ?

docarray/utils/find.py
```diff
@@ -114,7 +124,7 @@ def find_batched(
     limit: int = 10,
     device: Optional[str] = None,
     descending: Optional[bool] = None,
-) -> List[FindResult]:
+) -> FindResultBatched:
```
JohannesMessner, line 127: You need to adapt the implementation to actually return a `FindResultBatdched` :)

JohannesMessner (changes requested):

docarray/utils/find.py
npitsillos: @JohannesMessner how would I handle `scores` here?

npitsillos: I am not sure if this is the correct way to go about it but I used `parse_obj_as()` passing the embedding type so that scores are then parse as the correct tensor. Was this the correct way to go about this?

JohannesMessner (changes requested):
I would like to wait with merging this PR until weaviate and qdrant are merged.

The reason is that they both had to do a teeny tiny change to `FindRestulBatched` (the same change for both of them) that will conflict with this.

So since these are big PR coming from external partners, I would rather not cause them additional troubles, so I would merge their PRs first, then merge this one and resolve the conflicts

docarray/utils/find.py
```diff
@@ -23,6 +23,16 @@ class _FindResult(NamedTuple):
     scores: AnyTensor
 
 
+class FindResultBatched(NamedTuple):
+    documents: List[DocList]
+    scores: AnyTensor
```
JohannesMessner, line 28: This should be `List[AnyTensor]` or `List[np.ndarray]` like in the qdrant and weaviate PRs. The reason is that for batched search operations, especially if filtering is involved, not every query is guaranteed to yield the same number of results. If this happens, then the scores cannot be stacked up into one 2-axis array/tensor. Thus the different type.

JohannesMessner (changes requested):

docarray/utils/find.py
npitsillos: Just made these changes let me know if something else is needed.

docarray/utils/find.py
```diff
@@ -197,14 +208,17 @@ class MyDocument(BaseDoc):
...
+    ):
         docs_per_query: DocList = DocList([])
         for idx in indices_per_query:  # workaround until #930 is fixed
             docs_per_query.append(index[idx])
-        docs_per_query = DocList(docs_per_query)
-        results.append(FindResult(scores=scores_per_query, documents=docs_per_query))
-    return results
+        batched_docs.append(DocList(docs_per_query))
+        scores.append(scores_per_query)
+    return FindResultBatched(documents=batched_docs, scores=comp_backend.stack(scores))
```
JohannesMessner, line 221: I think stacking the scores is not required here anymore

docarray/index/backends/elastic.py
```diff
@@ -445,9 +440,7 @@ def _text_search_batched(
         das, scores = zip(
             *[self._format_response(resp) for resp in responses['responses']]
         )
-        return _FindResultBatched(
-            documents=list(das), scores=np.array(scores, dtype=object)
-        )
+        return _FindResultBatched(documents=list(das), scores=np.array(scores))
```
JohannesMessner, line 443: scores should be a list of NdArray here

JohannesMessner (changes requested):

JohannesMessner (approved):

PR #1358, opened 2023-04-11
refactor: rename from_pandas to from_dataframe

samsja:
# Context 

rename from_pandas to from_dataframe


samsja:
> maybe same for `.to_pandas()` -> `.to_dataframe()`

good point

PR #1362, opened 2023-04-11
feat: read from remote csv file

anna-charlotte:
Goals:

Currently, only local urls work. Let's allow remote ones too

### DIFF
--- a/docarray/array/doc_list/io.py
+++ b/docarray/array/doc_list/io.py
@@ -372,6 +372,15 @@ class IOMixinArray(Iterable[T_doc]):
         doc_type = cls.doc_type
         docs = DocList.__class_getitem__(doc_type)()
 
+        if file_path.startswith('http'):
+            import urllib3
+
+            http = urllib3.PoolManager()
+
+            response = http.request('GET', file_path)
+            print(f"response.data = {response.data}")
+            file_path = response.data
+
         with open(file_path, 'r', encoding=encoding) as fp:
             rows = csv.DictReader(fp, dialect=dialect)
             field_names: List[str] = (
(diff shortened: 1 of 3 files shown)


docarray/array/doc_list/io.py
```diff
@@ -361,40 +362,52 @@ def from_csv(
...
+            import urllib.request
+
+            with urllib.request.urlopen(file_path) as f:
+                file_as_string = StringIO(f.read().decode(encoding))
+                rows = csv.DictReader(file_as_string, dialect=dialect)
+                docs = cls._from_csv_dict_reader(rows)
+        else:
+            with open(file_path, 'r', encoding=encoding) as fp:
+                rows = csv.DictReader(fp, dialect=dialect)
+                docs = cls._from_csv_dict_reader(rows)
```
samsja, line 381: you can factorise your code so that the line csv.DictReader and from_csv_dict_reader is not repeated

samsja: or maybe it is not that easy with the context manager ?

samsja: you can create a function that you call inside both context

samsja (changes requested):
small comment

docarray/array/doc_list/io.py
```diff
@@ -372,13 +372,12 @@ def from_csv(
...
             with urllib.request.urlopen(file_path) as f:
-                file_as_string = StringIO(f.read().decode(encoding))
-                rows = csv.DictReader(file_as_string, dialect=dialect)
-                docs = cls._from_csv_dict_reader(rows)
+                file = StringIO(f.read().decode(encoding))
         else:
-            with open(file_path, 'r', encoding=encoding) as fp:
-                rows = csv.DictReader(fp, dialect=dialect)
-                docs = cls._from_csv_dict_reader(rows)
+            file = open(file_path, 'r', encoding=encoding)
```
samsja: arf, but here if you don't use the context manager then you might not close the file properly. better to use the context manager to be safer

samsja (approved):

PR #1401, opened 2023-04-17
docs: final round of fixes

alexcg1:
Still some [things](https://discord.com/channels/1062677585999429653/1097520320195330088/1097542697474076735) that need to be resolved, but think these are for someone else.

Signed-off-by: Alex C-G <alexcg@outlook.com>


samsja (changes requested):
can u remove the docs/readme ?

PR #1426, opened 2023-04-20
fix: passes max_element when load index in hnswlib

AnneYang720:
This is the additional fix for PR #1424 .
We need to pass `max_elements` in method `load_index()`, otherwise it's reset to 0.

### DIFF
--- a/tests/index/hnswlib/test_persist_data.py
+++ b/tests/index/hnswlib/test_persist_data.py
@@ -23,6 +23,10 @@ def test_persist_and_restore(tmp_path):
 
     # create index
     store = HnswDocumentIndex[SimpleDoc](work_dir=str(tmp_path))
+
+    # load existing index file
+    store = HnswDocumentIndex[SimpleDoc](work_dir=str(tmp_path))
+    assert store.num_docs() == 0
     store.index([SimpleDoc(tens=np.random.random((10,))) for _ in range(10)])
     assert store.num_docs() == 10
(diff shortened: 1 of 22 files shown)


JoanFM (changes requested):
Please use this PR to add the tests discussed in discord

tests/index/hnswlib/test_persist_data.py
```diff
@@ -81,5 +81,7 @@ def test_persist_and_restore_nested(tmp_path):
 
 
 def test_persist_index_file(tmp_path):
-    _ = HnswDocumentIndex[SimpleDoc](work_dir=str(tmp_path))
-    _ = HnswDocumentIndex[SimpleDoc](work_dir=str(tmp_path))
+    store = HnswDocumentIndex[SimpleDoc](work_dir=str(tmp_path))
```
JoanFM, line 84: can u call them `index` better, to have a clear distinction between indexes and stores

JoanFM (changes requested):

tests/index/hnswlib/test_persist_data.py
AnneYang720: > can u call them `index` better, to have a clear distinction between indexes and stores

`store` is used in all existing test files. Can I change them all in this PR? @JoanFM

JoanFM: yes

JoanFM (approved):

Issue #1235, opened 2023-03-14
Subindex search + navigate and traverse nested structures with `id`, `parent_id` and `root_id`


JoanFM:

JohannesMessner:
@hsm207 fyi

PR #1428, opened 2023-04-20
feat: subindex for document index

AnneYang720:
This PR is related to issue #1235 

**ToDo**:
Subindex
- [x] __init__
   - [x] abstract
   - [x] change schema (add `parent_id` dynamically)
   - [x] hnswlib
   - [x] elastic
- [x] index 
   - [x] abtract
   - [x] hnswlib 
   - [x] elastic
- [x] get
- [x] del
- [x] find
- [x] find (return both root and subindex results)
- [x] filter
- [x] tests
   - [x] abtract common methods
   - [x] hnswlib
   - [x] elastic (v7&v8)
- [x] weaviate
- [x] qdrant
- [x] Documentation

**Example**
```python
class SimpleDoc(BaseDoc):
    simple_tens: NdArray[10]
    simple_text: str

class MyDoc(BaseDoc):
    docs: DocList[SimpleDoc]

my_docs = [MyDoc(id='0', docs=DocList[SimpleDoc]([SimpleDoc(simple_tens=np.ones(10) * (j + 1), simple_text=f'hello {j}',)for _ in range(10)])]
index = ElasticDocIndex[MyDoc](index_name='idx')
index.index(my_docs)# index with name 'idx' and 'idx__docs' will be generated

doc = index['0] #this doc should be complete
print(doc) 
print(doc.docs[0])

# you can find on subindex level, and return results will be subindex docs
docs, scores = index.find(query, search_field='docs__simple_tens', limit=1) 
# or you can use find_subindex, which can return both root and sub results
root, sub, stores = index.find_subindex(query, search_field='docs__simple_tens', limit=5)

# filter_subindex is for filter on subindex level, return results are subindex docs
query = {'match': {'simple_text': 'hello 0'}}
docs = index.filter_subindex(query, subindex='docs', limit=5)

del index['0'] # delete all subindex docs as well
```


docarray/base_doc/doc.py
```diff
@@ -73,6 +74,43 @@ class Config:
         validate_assignment = True
         _load_extra_fields_from_protobuf = False
 
+    @classmethod
+    def _add_fields(cls, **field_definitions):
```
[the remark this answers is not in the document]
samsja:
```suggestion
    def _add_fields(cls, **field_definitions) -> None:
```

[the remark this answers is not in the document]
samsja: I guess this is used for add the root it and removing it right ?

I am not a fan of this method because it change the original class.

We should rather create a new class that inherit from this class and add the root_id/parent_id information

samsja (changes requested):
this looks great !

But we need to change how we add the root id to the original class. We cannot modify the class entirely for this purpose even if it is only temporary. I suggest to rather create a new class each time

JohannesMessner (changes requested):
Sorry, didn't get to review everything before going on vacation. So only commenting on part of the code changes ^^

docarray/index/abstract.py
```diff
@@ -85,12 +86,20 @@ class BaseDocIndex(ABC, Generic[TSchema]):
...
-    def __init__(self, db_config=None, **kwargs):
+    def __init__(self, db_config=None, subindex: bool = False, **kwargs):
         if self._schema is None:
             raise ValueError(
                 'A DocumentIndex must be typed with a Document type.'
                 'To do so, use the syntax: DocumentIndex[DocumentType]'
             )
+        if subindex:
+
+            class _new_schema(self._schema):  # type: ignore
```
samsja, line 97:
```suggestion
            class _NewSchema(self._schema):  # type: ignore
```

docarray/index/abstract.py
```diff
@@ -85,12 +86,20 @@ class BaseDocIndex(ABC, Generic[TSchema]):
...
+                parent_id: Optional[ID] = None
+
+            self._ori_schema = self._schema
+            self._schema = cast(Type[BaseDoc], _new_schema)
```
samsja, line 101:
```suggestion
            self._schema = cast(Type[BaseDoc], _NewSchema)
```

docarray/index/abstract.py
```diff
@@ -427,6 +473,40 @@ def find(
 
         return FindResult(documents=docs, scores=scores)
 
+    def find_subindex(  # TODO change method name
```
samsja, line 476: what methd name should we pick ?

docarray/index/abstract.py
```diff
@@ -427,6 +473,40 @@ def find(
...
+        self,
+        query: Union[AnyTensor, BaseDoc],
+        search_field: str = '',
+        limit: int = 10,
+        **kwargs,
+    ) -> SubindexFindResult:
+        self._logger.debug(
```
samsja, line 483: can u add the docstring ?

samsja (commented):
code lgtm.

One last things is do we really name the other find function `find_subindex` ? 
I think the name is a bit confusing but I don't have a better name in mind

AnneYang720:
> code lgtm.
> 
> One last things is do we really name the other find function `find_subindex` ? I think the name is a bit confusing but I don't have a better name in mind

Me neither, that's why I left the TODO

samsja:
@JohannesMessner @JoanFM what do you think ?

JohannesMessner:
> # you can find on subindex level, and return results will be subindex docs

Can we change it such that this one return on the root level, not the subindex level? Or is that difficult to do?

The reason I am asking is because normal nested search (not subindex) returns on the root level, so this should do the same, otherwise it will be confusing imo

samsja:
but is `subindex` clear outside of our world ? Is it a common term ?

docarray/index/abstract.py
```diff
@@ -116,6 +127,8 @@ def build(self, *args, **kwargs) -> Any:
             """
             ...
 
+        # TODO support subindex in QueryBuilder
```
JohannesMessner, line 130: I guess this will be left for future iterations? That is fine, let's just document it properly :)

docarray/index/abstract.py
```diff
@@ -890,17 +1053,64 @@ def _convert_dict_to_doc(
...
+            parent_id = doc['id']
+            nested_docs_id = self._subindices[field_name]._filter_by_parent_id(
+                parent_id
+            )
+            if nested_docs_id:
+                doc[field_name] = self._subindices[field_name].__getitem__(
+                    nested_docs_id
+                )
+
+    def _get_root_doc_id(self, id: str, root: str, sub: str) -> str:
```
JohannesMessner, line 1100: can we get docstrings for these?

docarray/index/backends/elastic.py
```diff
@@ -90,6 +94,13 @@ def __init__(self, db_config=None, **kwargs):
         mappings.update(self._db_config.index_mappings)
 
         for col_name, col in self._column_infos.items():
+            if issubclass(col.docarray_type, AnyDocArray):
+                sub_db_config = copy.deepcopy(self._db_config)
+                sub_db_config.index_name = f'{self._index_name}__{col_name}'
+                self._subindices[col_name] = self.__class__[col.docarray_type.doc_type](
+                    db_config=sub_db_config, subindex=True
+                )
+                continue
```
JohannesMessner: I think this same code is repeated in all backends. Can't it be done in the abstract base class?

I think in v1 we managed to do this sort of thing in the base class, any only exposed the `index_name` as something that the backend implementation needs to tackle, while the rest was "automated". I think the same should be possible here

JohannesMessner (changes requested):

docarray/index/abstract.py
samsja: yes lets open issue if we have todo left

samsja (approved):
lgtm very goo job !

docarray/index/backends/elastic.py
JohannesMessner: This implies that `index_name` is now an alias that can be used with any backend, right?
If so, this needs to be documented. In parts of the docs I tell people to look at the fields of `DBConfig` to see all available configs; if they now see `collection_name` and `index_name` that both do the same it will be confusing

docarray/index/abstract.py
```diff
@@ -141,6 +156,10 @@ class RuntimeConfig(ABC):
         # Example: `default_column_config['VARCHAR'] = {'length': 255}`
         default_column_config: Dict[Type, Dict[str, Any]] = field(default_factory=dict)
 
+    @property
+    def index_name(self):
+        ...
```
JohannesMessner, line 161: Doctring please :)

docarray/index/backends/hnswlib.py
```diff
@@ -184,9 +198,24 @@ def python_type_to_db_type(self, python_type: Type) -> Any:
 
         return None  # all types allowed, but no db type needed
 
-    def _index(self, column_data_dic, **kwargs):
+    # TODO fix mypy here
```
JohannesMessner, line 201: I think this comment can be removed?

docarray/index/backends/weaviate.py
```diff
@@ -358,6 +370,47 @@ def find(
 
         return FindResult(documents=docs, scores=scores)
 
+    def find_subindex(
```
JohannesMessner, line 373: Why does weaviate need to implement its own `find_subindex` instead of using the one from the abstract class?

docs/how_to/add_doc_index.md
```diff
@@ -402,6 +402,16 @@ in your backend, if such a concept exists in your case. In your implementation y
...
 If your implementation encounters a duplicate `.id`, it is okay to fail and raise an Exception.)
 
+### The `_filter_by_parent_id()` method
+
+The default implementatin return `None`. You can choose to override this function with database specific filter API when needed. 
+This function should return a list of ids of subindex level documents given the id of root document.
+
+### The `index_name()` property
+
+The `index_name` property is used in the initialization of subindices, and the default implementation is empty. This function should return the name of the index. And if the property of the index name in your backend is not `index_name`, you need to convert it as the first step in `__init__()`, like `index_name` is assigned to `work_dir` in `docarray/index/backends/hnswlib.py`.
```
JohannesMessner, line 412: For non-subindex the default behaviour is now to use the class name of the schema, see here: https://github.com/docarray/docarray/pull/1486

I think for subindex something similar should apply:

```python
class Doc(BaseDoc):
    inner = DocList[Bla]
```
I think the subindex for `inner` here should have a reasonable default index name.
Maybe `inner_Bla` to avoid conflict with other indices of the `Bla` class?

@samsja @JoanFM I'm als curious about your input here

docs/user_guide/storing/docindex.md
```diff
@@ -639,3 +641,81 @@ docs, scores = doc_index.find(query_doc, search_field='thumbnail__tensor', limit
 # find by the `video` tensor; neseted level
 docs, scores = doc_index.find(query_doc, search_field='video__tensor', limit=3)
 ```
+
+### Nested data with subindex
+**Index**
```
JohannesMessner, line 646:
```suggestion
### Nested data with subindex

Documents can be nested by containing a `DocList` of other documents, which is a slightly more complicated scenario than the one [above][## Nested data].

If a Document contains a DocList, it can still be stored in a Document Index.
In this case, the DocList will be represented as a new index (or table, collection, etc., depending on the database backend), that is linked with the parent index (table, collection, ...).

This still lets index and search through all of your data, but if you want to avoid the creation of additional indexes you could try to refactor your document schemas without the use of DocList.


**Index**
```
please make sure that the markdown above is rendered properly :)

JohannesMessner (changes requested):

JohannesMessner:
> > you can find on subindex level, and return results will be subindex docs
> 
> Can we change it such that this one return on the root level, not the subindex level? Or is that difficult to do?
> 
> The reason I am asking is because normal nested search (not subindex) returns on the root level, so this should do the same, otherwise it will be confusing imo

@AnneYang720 what is the status on this?

docs/how_to/add_doc_index.md
JoanFM: it would not be a default however, it should be defined by thr parent indrx right? I think the name of the subindex should be fully determined by the name og the root and the class and name of the subindex

JohannesMessner: > it would not be a default however, it should be defined by thr parent indrx right? I think the name of the subindex should be fully determined by the name og the root and the class and name of the subindex

Yes that would be nice, however there are differences in what the different backends accept as a valid name. Therefore just doing `rootname + subname` or `rootname + "_" + subname` may not work for all of them; but great if it does!

docarray/index/backends/weaviate.py
AnneYang720: Because the `search_field` for other backend is like `docs__simple_tens`, which includes the nested field name. But it should be only `docs` for weaviate.

> Weaviate is configured to allow one vector per data object

AnneYang720:
> > > you can find on subindex level, and return results will be subindex docs
> > 
> > 
> > Can we change it such that this one return on the root level, not the subindex level? Or is that difficult to do?
> > The reason I am asking is because normal nested search (not subindex) returns on the root level, so this should do the same, otherwise it will be confusing imo
> 
> @AnneYang720 what is the status on this?

This one is hard to change for two reasons:
1. The `find()` is recursively called. For example `search_field='list_docs__docs__simple_tens'`, when `find` is called on subindex `docs` level, we already lose the root level info.
2. `find_subindex()` use `find()` to get the subindex results as well.

But we can let `find()` return on the root level and implement another new function like `find_subindex_only()` as a substitue.

JohannesMessner:
> > > > you can find on subindex level, and return results will be subindex docs
> > > 
> > > 
> > > Can we change it such that this one return on the root level, not the subindex level? Or is that difficult to do?
> > > The reason I am asking is because normal nested search (not subindex) returns on the root level, so this should do the same, otherwise it will be confusing imo
> > 
> > 
> > @AnneYang720 what is the status on this?
> 
> This one is hard to change for two reasons:
> 
>     1. The `find()` is recursively called. For example `search_field='list_docs__docs__simple_tens'`, when `find` is called on subindex `docs` level, we already lose the root level info.
> 
>     2. `find_subindex()` use `find()` to get the subindex results as well.
> 
> 
> But we can let `find()` return on the root level and implement another new function like `find_subindex_only()` as a substitue.

I see. Maybe we should drop the subindex functionality in `.find()` altogether? I would rather it not being there than it being confusing. We could detect if a user passes a field that belongs to a subindex and then raise an error clearly stating that they should use `find_subindex` instead.
@AnneYang720 @samsja what do you think?

docs/how_to/add_doc_index.md
AnneYang720: `index_name` property is used when getting the `rootname`:
```python
sub_db_config.index_name = f'{self.index_name}__{col_name}'
```

JohannesMessner (approved):

PR #1434, opened 2023-04-24
docs: index predefined documents

JohannesMessner:
Explains how to index predefined documents into a document index

### DIFF
--- a/docs/user_guide/storing/docindex.md
+++ b/docs/user_guide/storing/docindex.md
@@ -77,7 +77,7 @@ class MyDoc(BaseDoc):
 db = HnswDocumentIndex[MyDoc](work_dir='./my_test_db')
 ```
 
-**Schema definition:**
+### Schema definition
 
 In this code snippet, `HnswDocumentIndex` takes a schema of the form of `MyDoc`.
 The Document Index then _creates a column for each field in `MyDoc`_.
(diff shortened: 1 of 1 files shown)


samsja (approved):

docs/user_guide/storing/docindex.md
```diff
@@ -93,6 +93,69 @@ the database will store vectors with 128 dimensions.
     Instead of using `NdArray` you can use `TorchTensor` or `TensorFlowTensor` and the Document Index will handle that
     for you. This is supported for all Document Index backends. No need to convert your tensors to NumPy arrays manually!
 
+
+### Using a predefined Document as schema
+
+DocArray offers a number of predefined Documents, like [ImageDoce][docarray.documents.ImageDoc] and [TextDoc][docarray.documents.TextDoc].
```
alexcg1, line 99:
```suggestion
DocArray offers a number of predefined Documents, like [ImageDoc][docarray.documents.ImageDoc] and [TextDoc][docarray.documents.TextDoc].
```

docs/user_guide/storing/docindex.md
```diff
@@ -93,6 +93,69 @@ the database will store vectors with 128 dimensions.
...
+If you try to use these directly as a schema for a Document Index, you will get unexpected behavior:
+Depending on the backend, and exception will be raised, or no vector index for ANN lookup will be built.
```
alexcg1, line 101:
```suggestion
Depending on the backend, an exception will be raised, or no vector index for ANN lookup will be built.
```

docs/user_guide/storing/docindex.md
```diff
@@ -93,6 +93,69 @@ the database will store vectors with 128 dimensions.
...
+
+    class MyDoc(TextDoc):
+        embedding: AnyTensor = Field(n_dim=128)
+
+
+    db = HnswDocumentIndex[MyDoc](work_dir='test_db3')
+    ```
+
+Once the schema of your Document Index is defined in this way, the data that you are indexing can be either of the
+predefined Document type, or of your custom Document type.
```
alexcg1, line 138:
```suggestion
predefined Document types, or your custom Document type.
```

docs/user_guide/storing/docindex.md
```diff
@@ -136,6 +199,8 @@ need to have compatible schemas.
     - A and B have the same field names and field types
     - A and B have the same field names, and, for every field, the type of B is a subclass of the type of A
 
+    In particular this means that you can easily [index predefined Documents](#using-a-predefined-document-as-schema) into a Document Index.
```
alexcg1, line 202:
```suggestion
    In particular, this means that you can easily [index predefined Documents](#using-a-predefined-document-as-schema) into a Document Index.
```

alexcg1: What's the policy on capitalizing `Document` now that we don't use that class name? I think @samsja mentioned on Discord we don't do that any more.

alexcg1 (changes requested):

docs/user_guide/storing/docindex.md
JohannesMessner: I think we should still capitalize it, since it is a concept in our library. Lowercased it looks a bit weird and "unofficial" to me. Plus, I think the rule of thumb was always that "concepts" are capitalized, whereas classes go in between `backticks`

JohannesMessner:
@alexcg1 I was a bit fast on the trigger there, your suggested fixes are here: https://github.com/docarray/docarray/pull/1436/

docs/user_guide/storing/docindex.md
samsja: No strong feeling here. But tehcnically speaking Document is not a concept in term of code in the library

JohannesMessner: I think it is a concept but just not a class, otherwise "concept" and "class" would be synonyms. But I just checked the pydantic documentation, they don't capitalize "model". So no strong feeling either

PR #1464, opened 2023-04-27
fix: fix dataframe max recursion depth

JoanFM:


JoanFM:
@samsja actually my question is why `ID` is a subclass of `BaseDoc`?

samsja:
> @samsja actually my question is why ID is a subclass of BaseDoc?

it is not

samsja:
ID is a predefined type like NdArray, TorchTensor, ImageUrl, etc. Therefore they are not BaseDoc nor PydanticModel. They just define validator for pydantic to make it work

JoanFM:
We wil handle this in #1467

PR #1467, opened 2023-04-27
fix: raise error when calling to csv or to dataframe on non homegenou…

samsja:
# Context

 raise error when calling to csv or to dataframe on non homogenous DocList

fixing : https://github.com/docarray/docarray/issues/1463


JoanFM (changes requested):
Please make sure to adapt Documentation on how to do so and cast, also to reply to original user

samsja:
I don't think we need to update documentation here

JoanFM (approved):

PR #1470, opened 2023-04-27
fix: return doclist of same type as input index in find and findbatched

anna-charlotte:
Return a doclist of same type as the input index in `find()` and `find_batched()`.

### DIFF
--- a/docarray/utils/find.py
+++ b/docarray/utils/find.py
@@ -213,10 +213,11 @@ def find_batched(
     for _, (indices_per_query, scores_per_query) in enumerate(
         zip(top_indices, top_scores)
     ):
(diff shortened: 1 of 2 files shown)


docarray/utils/find.py
```diff
@@ -210,13 +210,14 @@ class MyDocument(BaseDoc):
 
     batched_docs: List[DocList] = []
     scores = []
-    for _, (indices_per_query, scores_per_query) in enumerate(
+    for i, (indices_per_query, scores_per_query) in enumerate(
```
JoanFM, line 213:
```suggestion
    for _, (indices_per_query, scores_per_query) in enumerate(
```

samsja (approved):

PR #1472, opened 2023-04-27
fix: allow doclist to have nested optional document

samsja:
# WIP

investigating this issue: https://github.com/docarray/docarray/issues/1468

### DIFF
--- a/docarray/array/doc_list/doc_list.py
+++ b/docarray/array/doc_list/doc_list.py
@@ -200,6 +200,7 @@ class DocList(
 
         if (
             not is_union_type(field_type)
+            and self.__class__.doc_type.__fields__[field].required
             and isinstance(field_type, type)
             and issubclass(field_type, BaseDoc)
         ):
--- a/tests/units/array/test_array.py
+++ b/tests/units/array/test_array.py
@@ -432,3 +432,23 @@ def test_sort():
     assert len(da) == 3
     assert da[0].url == 'http://url.com/foo_0.png'
     assert da[1].url == 'http://url.com/foo_1.png'
+
+
+def test_optional():
+    from typing import Optional
+
+    from docarray import BaseDoc, DocList
+    from docarray.typing import ImageUrl, NdArray
+
+    class Nested(BaseDoc):
+        tensor: NdArray
+
+    class Image(BaseDoc):
+        url: ImageUrl
+        features: Optional[Nested] = None
+
+    docs = DocList[Image]([Image(url='http://url.com/foo.png') for _ in range(10)])
+
+    assert docs.features == [None for _ in range(10)]
+    assert isinstance(docs.features, list)
+    assert not isinstance(docs.features, DocList)


JohannesMessner:
> This behavior is uniformized for the tensor field as well.

> This is a breaking change. Before we were casting the None value into Nan value of each of the tensor backends and we would store the whole column of Nan value.

is this breaking change necessary? can't we just keep nan values in the respective backend format?

docarray/array/doc_list/doc_list.py
```diff
@@ -102,6 +102,13 @@ class Image(BaseDoc):
     del docs[0:5]  # remove elements for 0 to 5 from DocList
     ```
 
+    !!! note
+        if the DocList is homogeneous and its schema contains nested BaseDoc
```
JohannesMessner, line 106:
```suggestion
        If the DocList is homogeneous and its schema contains nested BaseDoc
```

docarray/array/doc_list/doc_list.py
```diff
@@ -102,6 +102,13 @@ class Image(BaseDoc):
...
+        (i.e, BaseDoc inside a BaseDoc) but that it is an optional field. Calling
```
JohannesMessner, line 107:
```suggestion
        (i.e, BaseDoc inside a BaseDoc) where the nested Document is `Optional`, calling
```

docarray/array/doc_vec/column_storage.py
```diff
@@ -63,12 +64,22 @@ def __len__(self) -> int:
     def __getitem__(self: T, item: IndexIterType) -> T:
         if isinstance(item, tuple):
             item = list(item)
-        tensor_columns = {key: col[item] for key, col in self.tensor_columns.items()}
-        doc_columns = {key: col[item] for key, col in self.doc_columns.items()}
+        tensor_columns = {
+            key: col[item] if col is not None else col
```
JohannesMessner, line 68:
```suggestion
            key: col[item] if col is not None else None
```
It's the same, but a tad more readable imo, up to you if you want to do it this way :)
same for the lines below

docarray/array/doc_vec/column_storage.py
```diff
@@ -91,15 +102,35 @@ def __init__(self, index: int, storage: ColumnStorage):
...
+                    return col[self.index : self.index + 1]
+                else:
+                    return None
+
+        col = self.storage.columns[name]
+
+        if col is not None:
+            return col[self.index]
+        else:
+            return None
```
JohannesMessner, line 123:
```suggestion
        if col None:
            return None
        return col[self.index]
```

docarray/array/doc_vec/doc_vec.py
```diff
@@ -83,6 +84,15 @@ class DocVec(AnyDocArray[T_doc]):
     this Document "view" is similar to the behavior of `view = tensor[i]` in
     numpy/PyTorch.
 
+    !!! note
+        DocVec support optional fields. Nevertheless if a field is optional it needs to
```
JohannesMessner, line 88:
```suggestion
        DocVec supports optional fields. Nevertheless if a field is optional it needs to
```

docarray/array/doc_vec/doc_vec.py
```diff
@@ -83,6 +84,15 @@ class DocVec(AnyDocArray[T_doc]):
     this Document "view" is similar to the behavior of `view = tensor[i]` in
     numpy/PyTorch.
 
+    !!! note
```
JohannesMessner, line 87: Maybe explain that otherwise it cannot be stacked, otherwise this seems a bit arbitrary

docarray/array/doc_vec/doc_vec.py
```diff
@@ -83,6 +84,15 @@ class DocVec(AnyDocArray[T_doc]):
...
+        DocVec support optional fields. Nevertheless if a field is optional it needs to
+        be homogeneous. This means that if the first document has a None value all of the
+        other documents should have a None value as well.
+    !!! note
+        If one field is Optional the column will be stored
+        * as None if the first doc is as the field as None
```
JohannesMessner, line 93: `is as the field as None` I don't understand this

docarray/array/doc_vec/doc_vec.py
```diff
@@ -83,6 +84,15 @@ class DocVec(AnyDocArray[T_doc]):
...
+        * as a normal column otherwise that cannot contain None value
```
JohannesMessner, line 94:
```suggestion
        * otherwise, as a normal column
```

docarray/array/doc_vec/doc_vec.py
```diff
@@ -96,11 +106,17 @@ def __init__(
         docs: Sequence[T_doc],
         tensor_type: Type['AbstractTensor'] = NdArray,
     ):
+
+        if not hasattr(self, 'doc_type') or self.doc_type == AnyDoc:
+            raise TypeError(
+                f'{self.__class__.__name__} does not precise a doc_type. You probably should do'
+                f'docs = DocVec[MyDoc](docs) instead of DocVec(docs)'
+            )
```
JohannesMessner, line 114: This seems like a big breaking change? Or are we just failing earlier?

docarray/array/doc_vec/doc_vec.py
```diff
@@ -96,11 +106,17 @@ def __init__(
         docs: Sequence[T_doc],
         tensor_type: Type['AbstractTensor'] = NdArray,
     ):
+
+        if not hasattr(self, 'doc_type') or self.doc_type == AnyDoc:
+            raise TypeError(
+                f'{self.__class__.__name__} does not precise a doc_type. You probably should do'
```
JohannesMessner, line 112:
```suggestion
                f'{self.__class__.__name__} does not precise a doc_type. You need to do'
```
If it fails then it is not "probably", right?

docarray/array/doc_vec/doc_vec.py
```diff
@@ -116,64 +132,113 @@ def __init__(
...
+
+            first_doc_is_none = getattr(docs[0], field_name) is None
+
+            def _check_none_field():
+
+                if is_field_required:
+                    if first_doc_is_none:
+                        raise ValueError(
+                            f'Field {field_name} is None for {docs[0]} even though it is required'
+                        )
```
JohannesMessner, line 145:
```suggestion
                if is_field_required and first_doc_is_none:
                    raise ValueError(
                        f'Field {field_name} is None for {docs[0]} even though it is required'
                    )
```

docarray/array/doc_vec/doc_vec.py
```diff
@@ -116,64 +132,113 @@ def __init__(
...
-                        tf_stack.append(val.tensor)
 
-                    stacked: tf.Tensor = tf.stack(tf_stack)
-                    tensor_columns[field_name] = TensorFlowTensor(stacked)
+                    if first_doc_is_none:
+                        _check_none_field()
+                        tensor_columns[field_name] = None
+                    else:
+                        tf_stack = []
+                        for i, doc in enumerate(docs):
```
JohannesMessner, line 176: This loop seems really expensive to me, no? Iterating over all the data in python code? I think this is a problem with decent sized data

docarray/array/doc_vec/doc_vec.py
```diff
@@ -116,64 +132,113 @@ def __init__(
...
+                        )
+
+                if first_doc_is_none:
+                    for i, doc in enumerate(docs):
+                        if getattr(doc, field_name) is not None:
+                            raise ValueError(
+                                f'Field {field_name} is put to None for the first doc. This mean that all of the other docs should have this field set to None as well. This is not the case for {doc} at index {i}'
+                            )
+
+            def _check_field_not_none(field_name, doc):
```
JohannesMessner, line 154: This helper fn and the one above have almost the same name, i think it is a bit confusing

docarray/array/doc_vec/doc_vec.py
```diff
@@ -116,64 +132,113 @@ def __init__(
...
+            def _check_none_field():
+
+                if is_field_required:
+                    if first_doc_is_none:
+                        raise ValueError(
+                            f'Field {field_name} is None for {docs[0]} even though it is required'
+                        )
+
+                if first_doc_is_none:
+                    for i, doc in enumerate(docs):
```
JohannesMessner, line 148: I would prefer if this got `docs` passed as an argument, this seems a bit smelly. Plus it "hides" the fact that this fn iterates over all docs

docarray/array/doc_vec/doc_vec.py
```diff
@@ -116,64 +132,113 @@ def __init__(
...
                         )
-                    )
-
-                    for i, doc in enumerate(docs):
-                        val = getattr(doc, field_name)
-                        if val is None:
-                            val = tensor_type.get_comp_backend().none_value()
 
-                        cast(AbstractTensor, tensor_columns[field_name])[i] = val
+                        for i, doc in enumerate(docs):
```
JohannesMessner, line 207: another iteration over al the data :/

docarray/array/doc_vec/doc_vec.py
```diff
@@ -353,46 +421,65 @@ def _set_data_column(
...
+                    values,
+                )
+                self._storage.doc_columns[field] = values_
+
+            elif field in self._storage.docs_vec_columns.keys():
+                values_ = cast(Sequence[DocList[T_doc]], values)
+                # TODO here we should actually check if this is correct
+                self._storage.docs_vec_columns[field] = values_
+            elif field in self._storage.any_columns.keys():
+                # TODO here we should actually check if this is correct
```
JohannesMessner, line 478: unresolved TODOs

JohannesMessner (changes requested):

docarray/array/doc_vec/doc_vec.py
samsja: actually no. this happen only once I think. if the first data point is None I check that this is the case for the rest

docarray/array/doc_vec/column_storage.py
samsja: fair

docarray/array/doc_vec/doc_vec.py
samsja: renamed

docarray/array/doc_vec/doc_vec.py
samsja: yes but this is todo from even before

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,104 @@ assert not my_doc.is_view()  # False
...
+
+class ArticleDoc(BaseDoc):
+    image: Optional[ImageDoc]
+    title: str
+```
+
+In this example `ArticleDoc` has an optional field `image` which is an `ImageDoc`. This means that this field can either
+be None or be a `ImageDoc` instance.
+
+Remember for both DocList and DocVec calling `docs.image` will return a list like object of all the images of the documents.
```
JohannesMessner, line 472:
```suggestion
Remember that for both DocList and DocVec calling `docs.image` will return a list-like object of all the images of the documents.
```

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,104 @@ assert not my_doc.is_view()  # False
...
+
+For DocList it will just iterate over all the documents and collect the image attribute of each document in a sequence for DocVec it will return the column of the image attribute.
```
JohannesMessner, line 474:
```suggestion
For DocList this call will iterate over all the documents and collect the image attribute of each document in a sequence, and for DocVec it will return the already stacked column of the `.image` attribute.
```

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,104 @@ assert not my_doc.is_view()  # False
...
+
+The question which kind of sequence to you pick when the field is optional, i.e, some of the datapoint could be None ?
```
JohannesMessner, line 476: What does it mean "which kind of sequence to pick"?

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,104 @@ assert not my_doc.is_view()  # False
...
+    [
+        ArticleDoc(image=ImageDoc(tensor=np.ones((3, 224, 224))), title="Hello"),
+        ArticleDoc(image=None, title="World"),
+    ]
+)
+
+assert docs.image == [ImageDoc(tensor=np.ones((3, 224, 224))), None]
+```
+
+but for DocVec it is a bit different. Indeed, DocVec store the data for each filed as contiguous column.
```
JohannesMessner, line 494:
```suggestion
But for DocVec it is a bit different. Indeed, a DocVec stores the data for each filed as contiguous column. 
```

JohannesMessner, line 494: What does the fact that it stores it as a column to do with it not being a DocList? That column could in theory be a DocList, right?

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,104 @@ assert not my_doc.is_view()  # False
...
+        ArticleDoc(image=ImageDoc(tensor=np.ones((3, 224, 224))), title="Hello"),
+        ArticleDoc(image=None, title="World"),
+    ]
+)
+
+assert docs.image == [ImageDoc(tensor=np.ones((3, 224, 224))), None]
+```
+
+but for DocVec it is a bit different. Indeed, DocVec store the data for each filed as contiguous column. 
+This means that DocVec can create a column in only two case: either all the data for a field is None or all the data is not None.
```
JohannesMessner, line 495:
```suggestion
This means that DocVec can create a column in only two cases: either all the data for a field is None or all the data is not None.
```
I think in other places you have a name for this. Homogeneous? something like that? Would be good to introduce here I think.
Or did we take homogeneous to mean same doc type for every doc? Honestly I don't remember

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,104 @@ assert not my_doc.is_view()  # False
...
+docs = DocVec[ArticleDoc](
+    [
+        ArticleDoc(image=ImageDoc(tensor=np.zeros((3, 224, 224))), title="Hello")
+        for _ in range(10)
+    ]
+)
+assert (docs.image.tensor == np.zeros((3, 224, 224))).all()
+``` 
+
+Or it can be None
```
JohannesMessner, line 513:
```suggestion
Or it can be None:
```

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,104 @@ assert not my_doc.is_view()  # False
...
+
+but for DocVec it is a bit different. Indeed, DocVec store the data for each filed as contiguous column. 
+This means that DocVec can create a column in only two case: either all the data for a field is None or all the data is not None.
+
+For the first one the whole column will just be None. In the second The column will be a `DocList[ImageDoc]`
+
+
+
+```python
+from docarray import DocVec
```
JohannesMessner, line 502: maybe use tabs for these examples?

JohannesMessner (changes requested):
Sorry to be pain in the butt, a lot of suggestions :D

But honestly it seems like the entire doc section could be condensed down to a few sentences:
- Nested optionals are a bit tricky, watch out
- For doclist the return will be a `DocList` instead of `DocList[...]`
- For docvec the return will be `DocList[...]` if all documents are there, or `None` if all docs are None. No mix of docs and None allowed!

I would lead with these "rules", I think they are most important to the user. And after that we can go into an explanation of why that is, maybe even hidden away behind a collapsed box.

Otherwise I think it is difficult for the users to tease out what is the important information

samsja:
> Sorry to be pain in the butt, a lot of suggestions :D
> 
> But honestly it seems like the entire doc section could be condensed down to a few sentences:
> 
>     * Nested optionals are a bit tricky, watch out
> 
>     * For doclist the return will be a `DocList` instead of `DocList[...]`
> 
>     * For docvec the return will be `DocList[...]` if all documents are there, or `None` if all docs are None. No mix of docs and None allowed!
> 
> 
> I would lead with these "rules", I think they are most important to the user. And after that we can go into an explanation of why that is, maybe even hidden away behind a collapsed box.
> 
> Otherwise I think it is difficult for the users to tease out what is the important information

sure I will follow your suggestion

docs/user_guide/representing/array.md
samsja: I split it a bit differently now

docs/user_guide/representing/array.md
samsja: well not if some document are none ?

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,125 @@ assert not my_doc.is_view()  # False
     To summarize: you should use `DocVec` when you need to work with contiguous data, and you should use `DocList` when you need to rearrange
     or extend your data.
 
+
+## Dealing with Optional fields
```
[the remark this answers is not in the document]
samsja: yes

docs/user_guide/representing/array.md
```diff
@@ -439,6 +439,125 @@ assert not my_doc.is_view()  # False
...
+    class MyDoc(BaseDoc):
+        nested_doc: Optional[BaseDoc]
+    ```
+
+!!! note
+    Using optional field is slightly different in some case so watch out. But in a nutshell :
+    
+    For nested BaseDoc 
+    * DocList will return a list of document if the field is optional and a DocList if the field is not optional
+      * DocVec will return a DocList if all documents are there, or None if all docs are None. No mix of docs and None allowed!
```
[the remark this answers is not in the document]
samsja: shit you are right ofc

PR #1476, opened 2023-04-28
docs: change cosine similarity to distance for hnswlib doc index

anna-charlotte:
Change cosine similarity to distance in hnswlib doc index, because their `'cosine'` actually refers to distance, not similarity.

Opened issue in hnswlib repo: https://github.com/nmslib/hnswlib/issues/456


JoanFM (changes requested):
Can we add kind of a note highlighting this potential differences with other indexes? And also, a link or a formula to transform?

JoanFM (approved):

PR #1478, opened 2023-04-28
fix: use `pd.concat()` instead `df.append()` in `to_dataframe()` to avoid warning

anna-charlotte:
To avoid this warning, use `pd.concat()`:
image

### DIFF
--- a/docarray/array/doc_list/io.py
+++ b/docarray/array/doc_list/io.py
@@ -543,8 +543,8 @@ class IOMixinArray(Iterable[T_doc]):
         df = pd.DataFrame(columns=fields)
 
         for doc in self:
-            doc_dict = _dict_to_access_paths(doc.dict())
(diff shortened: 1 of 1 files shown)


JoanFM (approved):

docarray/array/doc_list/io.py
```diff
@@ -543,8 +543,9 @@ def to_dataframe(self) -> 'pd.DataFrame':
         df = pd.DataFrame(columns=fields)
 
         for doc in self:
-            df_doc = pd.DataFrame(_dict_to_access_paths(doc.dict()))
-            df = pd.concat([df, df_doc], ignore_index=True)
+            doc_dict = _dict_to_access_paths(doc.dict())
+            doc_dict = {k: [v] for k, v in doc_dict.items()}
```
JoanFM, line 547: why do u need to convert them to list?

PR #1479, opened 2023-05-01
fix: ndarray casting

alaeddine-13:
closes: https://github.com/docarray/docarray/issues/1439

### DIFF
--- a/docarray/typing/tensor/ndarray.py
+++ b/docarray/typing/tensor/ndarray.py
@@ -12,6 +12,7 @@ from typing import (
 )
 
 import numpy as np
+import torch
 
 from docarray.typing.proto_register import _register_proto
 from docarray.typing.tensor.abstract_tensor import AbstractTensor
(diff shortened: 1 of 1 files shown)


docarray/typing/tensor/ndarray.py
```diff
@@ -112,10 +112,13 @@ def validate(
         field: 'ModelField',
         config: 'BaseConfig',
     ) -> T:
+        torch = import_library('torch', raise_error=False)
         if isinstance(value, np.ndarray):
             return cls._docarray_from_native(value)
         elif isinstance(value, NdArray):
             return cast(T, value)
+        elif torch and isinstance(value, torch.Tensor):
+            return cast(T, value)
```
samsja, line 121: I think this can be a problem with or AnyTensor and how Pydantic deal with Union. cc @JohannesMessner .

Basically if someone define a BaseDoc with a AnyTensor and pass a torch.Tensor as input it will always be converted to a NdArray

alaeddine-13: can you share a code snippet showing when this is problematic ?

PR #1486, opened 2023-05-04
feat: index or collection name will default to doc-type name

JoanFM:


JohannesMessner (changes requested):
Since this is a breaking change I think we need to be careful to guide people to un-break their code.

Concretely what will happen is that collections/tables created with a docarray version before this will be called `Document`, and when a user tries to access their database with a new version of docarray it will assume a table called `MyDoc` or something like that, and it will fail.

I think for this case we should catch the error and add a message like:

> You are trying to connect to table {whatever}, but no such table exists. If you are trying to access a table that was created with DocArray version {one smaller than this one} this could be because the default table name has changed between versions. In that case, try adding the following `QdrantDocIndex(..., table_name='Document'`), or equivalently, adding `table_name='Document'` to your DBConfig

The details of this messag will depend on the particular backend i suppose

docarray/index/backends/weaviate.py
```diff
@@ -111,6 +111,17 @@ def __init__(self, db_config=None, **kwargs) -> None:
...
         self._create_schema()
 
+    @property
+    def index_name(self):
+        default_index_name = self._schema.__name__ if self._schema is not None else None
+        if default_index_name is None:
+            raise ValueError(
+                'A WeaviateDocumentIndex must be typed with a Document type.'
+                'To do so, use the syntax: WeaviateDocumentIndex[DocumentType]'
+            )
```
JohannesMessner, line 121: I don't think that this code is reachable because basically the same check happens already earlier, but maybe I am wrong

JoanFM: yes I know, but `mypy` complains otherwise

JoanFM:
> Since this is a breaking change I think we need to be careful to guide people to un-break their code.
> 
> Concretely what will happen is that collections/tables created with a docarray version before this will be called `Document`, and when a user tries to access their database with a new version of docarray it will assume a table called `MyDoc` or something like that, and it will fail.
> 
> I think for this case we should catch the error and add a message like:
> 
> > You are trying to connect to table {whatever}, but no such table exists. If you are trying to access a table that was created with DocArray version {one smaller than this one} this could be because the default table name has changed between versions. In that case, try adding the following `QdrantDocIndex(..., table_name='Document'`), or equivalently, adding `table_name='Document'` to your DBConfig
> 
> The details of this messag will depend on the particular backend i suppose

This would be really nice, but it does not seem so feasible, nowhere in the code I see this check being done, plus I am not sure what each of the Backends throw when this happens, or they may simply create the table/etc ...

I would simply add a breaking change note in the release notes once this is released

JohannesMessner (approved):

JoanFM:
@AnneYang720 why is it needed to change the names? why the tests do not pass with default naming?

AnneYang720:
> @AnneYang720 why is it needed to change the names? why the tests do not pass with default naming?

The schema name is used as default `index_name`, and sometimes this `index_name` is used in some tests before. For example:
```python
index = ElasticDocIndex[SimpleDoc]()
```
The `index_name` is `simpledoc` and is used in previous tests(`SimpleDoc` is used as schema), the `index.num_docs()` is not zero at the moment.

JoanFM:
> > @AnneYang720 why is it needed to change the names? why the tests do not pass with default naming?
> 
> The schema name is used as default `index_name`, and sometimes this `index_name` is used in some tests before. For example:
> 
> ```python
> index = ElasticDocIndex[SimpleDoc]()
> ```
> 
> The `index_name` is `simpledoc` and is used in previous tests(`SimpleDoc` is used as schema), the `index.num_docs()` is not zero at the moment.

Can we have some tests were the default name is used for indexing and retrieving in one shot?

AnneYang720:
> > > @AnneYang720 why is it needed to change the names? why the tests do not pass with default naming?
> > 
> > 
> > The schema name is used as default `index_name`, and sometimes this `index_name` is used in some tests before. For example:
> > ```python
> > index = ElasticDocIndex[SimpleDoc]()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The `index_name` is `simpledoc` and is used in previous tests(`SimpleDoc` is used as schema), the `index.num_docs()` is not zero at the moment.
> 
> Can we have some tests were the default name is used for indexing and retrieving in one shot?

Some tests with schema that is only used once(like `FlatSchema`) use default index name. And I just change `test_get_single`, `test_get_multiple` in elastic and `test_persist_and_restore` in qdrant so that they also use default index name.

PR #1510, opened 2023-05-08
feat: openapi tensor shapes

### LINKED ISSUE #1440: FastAPI openai spec shows "str" for tensor types and no shape info
When serving a fastapi app using a Document for the API schema, if one of the fields is a tensor-like the type shows up as "string". Further, the shape information is not shown.

It would be nice if we could
- display the shape of the tensor in the "schema" section
- show a plausible example in the "Example Value" section

**How to reproduce:**
In `main.py`:
```python
from docarray import BaseDoc
from fastapi import FastAPI

from docarray.base_doc import DocArrayResponse


class Doc(BaseDoc):
    embedding: TorchTensor[3, 256, 256]


app = FastAPI()


@app.post("/transcribe", response_model=Doc, response_class=DocArrayResponse)
async def transcribe(doc: Doc) -> Doc:
    return Doc(embedding=doc.embedding)
```
Then do:
```bash
uvicorn main:app --reload
```

And go to http://127.0.0.1:8000/docs#/default/transcribe_transcribe_post / http://127.0.0.1:8000/docs#/default/transcribe_transcribe_post

You will see:

Image

Image

JohannesMessner:
closes #1440 

This feature makes it s.t. shaped tensors show up with their appropriate shape in OpenAPI/SwaggerUI, in both the **examples** and the **schema**:

```python
from docarray import BaseDoc
from docarray.typing import TorchTensor, NdArray
from fastapi import FastAPI

from docarray.base_doc import DocArrayResponse


class Doc(BaseDoc):
    embedding_torch: TorchTensor[10]
    embedding_np: NdArray[3, 4]

app = FastAPI()


@app.post("/transcribe", response_model=Doc, response_class=DocArrayResponse)
async def transcribe(doc: Doc) -> Doc:
    return Doc(embedding=doc.embedding_np)

```
image
image

## TODO
- [x] tests
- [x] documentation


docarray/typing/tensor/abstract_tensor.py
```diff
@@ -225,6 +230,21 @@ def __docarray_validate_getitem__(cls, item: Any) -> Tuple[int]:
...
+        if cls.__docarray_target_shape__ is not None:
+            shape_info = (
+                '[' + ', '.join([str(s) for s in cls.__docarray_target_shape__]) + ']'
+            )
+            if reduce(mul, cls.__docarray_target_shape__, 1) <= 256:
+                # custom example only for 'small' shapes, otherwise it is too big to display
+                example_payload = orjson_dumps(np.zeros(cls.__docarray_target_shape__))
+                field_schema.update(example=example_payload)
+        else:
+            shape_info = 'not specified'
```
samsja, line 245:
```suggestion
            shape_info = 'any shape'
```

would not it be better like this ?

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -225,6 +230,21 @@ def __docarray_validate_getitem__(cls, item: Any) -> Tuple[int]:
             raise TypeError(f'{item} is not a valid tensor shape.')
         return item
 
+    @classmethod
+    def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
JohannesMessner: Strictly speaking no. It is used whenever someone generates a JSONSchema. But idk what other use cases for that are out there.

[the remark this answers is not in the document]
JoanFM: In Jina it is required to recreate a model from it in a remote service

docarray/typing/tensor/abstract_tensor.py
JohannesMessner: hmm maybe? I feel like it might give a false sense of confidence though, that fact that no shape was specified doesn't automatically mean that any shape will work

samsja: Okay, I see a parallel with python type hint:

```python
def f(x):
   ...
```
this is not specified

```python
def f(x: Any):
   ...
```

this means anything works.

Should we then have an Any shape stuff that would be different from note specified?

JohannesMessner: > Should we then have an Any shape stuff that would be different from not specified?

For mypy Any and not specified are the same thing, so if we want to stay within this analogy that you are drawing then i would say no actually

samsja: hmm okay fair

samsja: then lets keep like this

docarray/typing/tensor/abstract_tensor.py
```diff
@@ -225,6 +230,21 @@ def __docarray_validate_getitem__(cls, item: Any) -> Tuple[int]:
...
+        field_schema.update(type='array', items={'type': 'number'})
+        if cls.__docarray_target_shape__ is not None:
+            shape_info = (
+                '[' + ', '.join([str(s) for s in cls.__docarray_target_shape__]) + ']'
+            )
+            if reduce(mul, cls.__docarray_target_shape__, 1) <= 256:
```
JoanFM, line 240: make this 256 not a magic number

JoanFM (changes requested):

docarray/typing/tensor/abstract_tensor.py
JohannesMessner: I can pull it out as a constant, but it doesn't really come from anywhere honestly, so I think it'll have to stay hard coded

JoanFM (approved):

samsja (approved):

PR #1513, opened 2023-05-08
fix: fix json behavior of doclist to allow custom json encoder

samsja:
# Context

we need to let Doclist as they are inside BaseDoc otherwise custom json encoder will not worjk


JoanFM (changes requested):
let's add a test then

JoanFM (approved):

Issue #1504, opened 2023-05-08
Redundant parameter `search_field` in weaviate hybrid search


violenil:
The `search_field` parameter seems to be redundant in the hybrid search query on a Weaviate document index, since the text search will search all fields of the object and no specific text field may be supplied. However, when not supplied, the following error occurs:

<img width="821" alt="Screenshot 2023-05-08 at 11 45 20" src="https://user-images.githubusercontent.com/55404563/236793299-5c74e77c-b6b3-4efc-b3e1-c9025dcde2fc.png">

Although the [documentation](https://docs.docarray.org/user_guide/storing/index_weaviate/) also explains that this parameter is `necessary but has no effect`, since all text fields will be searched by the hybrid query, it seems there should be no requirement to pass a `None` value here.

 <img width="867" alt="Screenshot 2023-05-08 at 11 48 24" src="https://user-images.githubusercontent.com/55404563/236793705-ee0229f2-c91c-4907-91e4-ef3e84ddb4e9.png">

hsm207:
weaviate can now support searching in specific field(s) (the BM25 part) when doing hybrid search (see this [issue](https://github.com/weaviate/weaviate/issues/2747)). We should update the implementation in docarray accordingly.

JoanFM:
Hey @hsm207 , maybe u can create a ticket for this change?

AnneYang720:
Hi @hsm207 , I checked the lastest code of weaviate python client. It seems class `Hybrid` still doesn't have `properties` [here](https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/gql/get.py#L49).  Like in [weaviate go](https://github.com/weaviate/weaviate/pull/2968/files#diff-c323c8a4ee0021aaaa831b2c8d47abdf20b1722fc01e94f877a46413ccbe16c8), I guess this should be changed to:

```python
@dataclass
class Hybrid:
    query: str
    alpha: float
    vector: List[float]
    properties: Optional[List[str]]

    def __str__(self) -> str:
        ret = f'query: "{util.strip_newlines(self.query)}"'
        if self.vector is not None:
            ret += f", vector: {self.vector}"
        if self.alpha is not None:
            ret += f", alpha: {self.alpha}"
        if self.properties is not None and len(self.properties) > 0:
            props = '","'.join(self.properties)
            ret += f', properties: ["{props}"]'

        return "hybrid:{" + ret + "}"
```

JohannesMessner:
[...]

PR #1516, opened 2023-05-09
feat: search_field  should be optional in hybrid text search

AnneYang720:
This PR is related to issue #1504


hsm207 (changes requested):
Which PR has the update for the docs?

AnneYang720:
> Which PR has the update for the docs?

I just updated the doc here.

hsm207 (approved):

docs/user_guide/storing/index_weaviate.md
```diff
@@ -379,8 +380,8 @@ This will perform a hybrid search for the word "hello" and the vector [1, 2] and
 q = (
     store.build_query()
     .text_search(
-        "world", search_field=None  # Set as None as it is required but has no effect
-    )
+        "world"
+    )  # No need to set search_field as it has no effect in hybird search
```
JoanFM, line 384: I think there is no need to add this annotation here. I as a user I would not see why you mention something that is not there

JoanFM (changes requested):

JoanFM (approved):

PR #1519, opened 2023-05-09
feat: add DocDict

samsja:
# Context

early draft of DocDict

### DIFF
--- a/docarray/array/any_collections.py
+++ b/docarray/array/any_collections.py
@@ -1,0 +1,132 @@
+from abc import abstractmethod
+from typing import (
+    TYPE_CHECKING,
+    Dict,
+    Generic,
+    Iterable,
+    List,
+    MutableSequence,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+)
+
+from docarray.base_doc import BaseDoc
+from docarray.typing.abstract_type import AbstractType
+from docarray.utils._internal._typing import change_cls_name
+
+if TYPE_CHECKING:
+    from docarray.proto import DocListProto, NodeProto
+    from docarray.typing.tensor.abstract_tensor import AbstractTensor
+
+T = TypeVar('T', bound='AnyCollections')
+T_doc = TypeVar('T_doc', bound=BaseDoc)
+IndexIterType = Union[slice, Iterable[int], Iterable[bool], None]
+
+
+class AnyCollections(Generic[T_doc], AbstractType):
+    doc_type: Type[BaseDoc]
+    __typed_da__: Dict[Type['AnyCollections'], Dict[Type[BaseDoc], Type]] = {}
+
+    def __repr__(self):
+        return f'<{self.__class__.__name__} (length={len(self)})>'
+
+    @classmethod
+    def __class_getitem__(cls, item: Union[Type[BaseDoc], TypeVar, str]):
+        if not isinstance(item, type):
+            return Generic.__class_getitem__.__func__(cls, item)  # type: ignore
+            # this do nothing that checking that item is valid type var or str
+        if not issubclass(item, BaseDoc):
+            raise ValueError(
+                f'{cls.__name__}[item] item should be a Document not a {item} '
+            )
+
+        if cls not in cls.__typed_da__:
+            cls.__typed_da__[cls] = {}
+
+        if item not in cls.__typed_da__[cls]:
+            # Promote to global scope so multiprocessing can pickle it
+            global _DocArrayTyped
+
+            class _DocArrayTyped(cls):  # type: ignore
+                doc_type: Type[BaseDoc] = cast(Type[BaseDoc], item)
+
+            for field in _DocArrayTyped.doc_type.__fields__.keys():
+
+                def _property_generator(val: str):
+                    def _getter(self):
+                        return self._get_data_column(val)
+
+                    def _setter(self, value):
+                        self._set_data_column(val, value)
+
+                    # need docstring for the property
+                    return property(fget=_getter, fset=_setter)
+
+                setattr(_DocArrayTyped, field, _property_generator(field))
+                # this generates property on the fly based on the schema of the item
+
+            # The global scope and qualname need to refer to this class a unique name.
+            # Otherwise, creating another _DocArrayTyped will overwrite this one.
+            change_cls_name(
+                _DocArrayTyped, f'{cls.__name__}[{item.__name__}]', globals()
+            )
+
+            cls.__typed_da__[cls][item] = _DocArrayTyped
+
+        return cls.__typed_da__[cls][item]
+
+    def __getattr__(self, item: str):
+        # Needs to be explicitly defined here for the purpose to disable PyCharm's complaints
+        # about not detected properties: https://youtrack.jetbrains.com/issue/PY-47991
+        return super().__getattribute__(item)
+
+    @abstractmethod
+    def _get_data_column(
+        self: T,
+        field: str,
+    ) -> Union[MutableSequence, T, 'AbstractTensor', None]:
+        """Return all values of the fields from all docs this array contains
+
+        :param field: name of the fields to extract
+        :return: Returns a list of the field value for each document
+        in the array like container
+        """
+        ...
+
+    @abstractmethod
+    def _set_data_column(
+        self: T,
+        field: str,
+        values: Union[List, T, 'AbstractTensor'],
+    ):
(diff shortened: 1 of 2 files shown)


docarray/array/any_collections.py
```diff
@@ -0,0 +1,145 @@
...
+    from pydantic.fields import ModelField
+
+    from docarray.proto import DocListProto, NodeProto
+
+T = TypeVar('T', bound='AnyCollections')
+T_doc = TypeVar('T_doc', bound=BaseDoc)
+IndexIterType = Union[slice, Iterable[int], Iterable[bool], None]
+
+
+class AnyCollections(Generic[T_doc], AbstractType):
```
JohannesMessner, line 29:
```suggestion
class AnyCollection(Generic[T_doc], AbstractType):
```

docarray/array/any_collections.py
```diff
@@ -0,0 +1,145 @@
...
+    doc_type: Type[BaseDoc]
+    __typed_da__: Dict[Type['AnyCollections'], Dict[Type[BaseDoc], Type]] = {}
```
JohannesMessner, line 31:
```suggestion
    __typed_subclass__: Dict[Type['AnyCollections'], Dict[Type[BaseDoc], Type]] = {}
```

JohannesMessner: this is the class cache thing, right? maybe a small comment to explain it?

docarray/array/any_collections.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+
+        if cls not in cls.__typed_da__:
+            cls.__typed_da__[cls] = {}
+
+        if item not in cls.__typed_da__[cls]:
+            # Promote to global scope so multiprocessing can pickle it
+            global _DocArrayTyped
+
+            class _DocArrayTyped(cls):  # type: ignore
```
JohannesMessner, line 53: it's not rally a docarray anymore, more like a `_DocCollectionTyped`?

docarray/array/doc_dict/doc_dict.py
```diff
@@ -0,0 +1,88 @@
...
+        if not issubclass(self.doc_type, AnyDoc) and not isinstance(doc, self.doc_type):
+            raise ValueError(f'{doc} is not a {self.doc_type}')
+        return doc
+
+    @classmethod
+    def from_doc_list(cls: Type[T], docs: DocList) -> T:
+        return cls(**{doc.id: doc for doc in docs})
+
+    # here we need to ignore type as the DocDict as a signature incompatible with Dict (it is more restrictive)
+    def update(self, other: Union['DocDict', Dict[str, T_doc], DocList]):  # type: ignore
```
JohannesMessner, line 36: does `Sequence[Doc]` not work, only `DocList`?
Can't we in that case just do `super().update(DocDict.from_doc_list(DocList[self.doc_type](other)))`?

docarray/array/doc_dict/doc_dict.py
```diff
@@ -0,0 +1,88 @@
...
+            return DocDict.__class_getitem__(field_type)(  # todo skip validation
+                **{key: getattr(doc, field) for key, doc in self.items()},
+            )
+        else:
+            return {key: getattr(doc, field) for key, doc in self.items()}
+
+    def _set_data_column(
+        self: T,
+        field: str,
+        values: Union[Mapping[str, Any]],
```
JohannesMessner, line 66: this union only has one arg

tests/units/array/doc_dict/test_doc_dic.py
```diff
@@ -0,0 +1,102 @@
+import pytest
```
JohannesMessner, line 1: please rename this file ;) ;)

JohannesMessner (changes requested):

tests/units/array/doc_dict/test_doc_dic.py
samsja: lol yes

JohannesMessner:
Are we planning to have full support for the IO mixin?

Also, I think we should have `to_doclist` and `to_docvec`, and doclist and docvec should have `to_docdict` (in that case using id's as keys).

samsja:
Good point I will do

JohannesMessner:
> > Also, I think we should have to_doclist and to_docvec
> 
> in this case, what should happened when converting to DocList ? We just don't care about the dict key ? Or we override the key as the id of the Document ?
> 
> I would say option 1 but then I am not sure that this function is that usefull

I would ignore the key, yes. I guess that would make it identical to `l = DocList[MyDoc](doc_dict.keys())`? I think `l = doc_dict.to_doclist()` is a bit less verbose, especially bc of the `[]` notation, but yeah, the utility is not huge

docarray/array/doc_dict/doc_dict.py
```diff
@@ -0,0 +1,131 @@
...
+T = TypeVar('T', bound='DocDict')
+
+
+class DocDict(AnyCollection[T_doc], Dict[str, T_doc]):
+    doc_type: Type[BaseDoc] = AnyDoc
+
+    def __init__(self, **mapping):
+        super().__init__(
+            {key: self._validate_one_doc(val) for key, val in mapping.items()}
+        )
```
JohannesMessner, line 27: Didn't we want to have a default where if there is no key given, we use the id as key?

docarray/array/doc_dict/doc_dict.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def _validate_one_doc(self, doc: T_doc) -> T_doc:
+        """Validate if a Document is compatible with this `DocList`"""
+        if not issubclass(self.doc_type, AnyDoc) and not isinstance(doc, self.doc_type):
+            raise ValueError(f'{doc} is not a {self.doc_type}')
+        return doc
+
+    @classmethod
+    def from_doc_list(cls: Type[T], docs: Union[DocList, Iterable[BaseDoc]]) -> T:
```
JohannesMessner: Could this just happen in the constructor? Also, if an iter of docs is accepted, then the name is a bit misleading

docarray/array/doc_dict/doc_dict.py
```diff
@@ -0,0 +1,131 @@
...
+        :return: the deserialized `DocList`
+        """
+        json_docs = orjson.loads(file)
+        return cls(**{key: cls.doc_type(**doc) for key, doc in json_docs.items()})
+
+    def to_json(self) -> bytes:
+        """Convert the object into JSON bytes. Can be loaded via `.from_json`.
+        :return: JSON serialization of `DocList`
+        """
+        return orjson_dumps(self)
```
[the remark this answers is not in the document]
samsja: Is the goal to:

1- abstract away some part of the code to avoid to repeat our self.
or 
2- provide a base abstract class that will work as an Interface to show to the user what is available.


I thought about this while implementing and basically:

I think 1 is not needed bc most of the io code is cannot be factorized between different classes (or very few part of it and it will just be a pain to maintain)

for 2 I agree that we need it, But since we don't support json stuff on DocVec for now I was thinking that it is not really needed for now but will def'ly be when we harmonized all of the export

docarray/array/doc_dict/doc_dict.py
samsja: it is define in the `from_doc_list` method.

The init just take a mapping, like a normal dict

docarray/array/doc_dict/doc_dict.py
samsja: > Also, if an iter of docs is accepted, then the name is a bit misleading

yep I had the same feeling.

How should I named it them ?

1- `from_docs`
2- `from_seq_docs`
3- ???

docarray/array/doc_dict/doc_dict.py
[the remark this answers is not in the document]
JohannesMessner: I was thinking of 2, but makese sense, let's do that after IO for DocVec is merged

docarray/array/doc_dict/doc_dict.py
JohannesMessner: `from_docs` is good I think

docarray/array/doc_dict/doc_dict.py
JohannesMessner: I think normal dict takes more, no? Like list of tuples for example? Or am I wrong?

samsja: let me double check, but I thin the `dict()` function can just take a mapping. I will just follow whatever dict init do, Makes sense to you ?

PR #1523, opened 2023-05-10
chore: add protobuf to hnsw extra

jupyterjazz:
hnsw requires protobuf, so let's add it in the extra


samsja (approved):

samsja (changes requested):
need poetry lock no update

PR #1526, opened 2023-05-10
fix: detach the torch tensors

makram93:
Detach `TorchTensors` before converting to `numpy` for serialization. 

Change is only needed for PyTorch as Tensorflow operates differently afaik. I have tried different initialization for tensorflow and the below serializing script worked without any error.

```python

import time

from docarray import BaseDoc
from docarray.typing import TensorFlowTensor
import tensorflow as tf


class MyDoc(BaseDoc):
    x: TensorFlowTensor
    y: TensorFlowTensor


# 1. Eager execution with GradientTape
x = tf.Variable(3.0)  # trainable: Default: `True`, GradientTapes automatically watch uses of this variable.

with tf.GradientTape() as tape:
    y = x**2

doc = MyDoc(x=x, y=y)
doc.json()


# 2. GradientTape with tf.function
@tf.function
def square(a):
    return a**2


x = tf.Variable(tf.random.normal((3, 2)))

with tf.GradientTape() as tape:
    y = square(x)

doc = MyDoc(x=x, y=y)
doc.json()

# 3. TF constants
x = tf.Variable(tf.random.normal((3, 2)))
y = tf.constant(2.0)

doc = MyDoc(x=x, y=y)
doc.json()

# 4. Full example
w = tf.Variable(tf.random.normal((3, 2)), name="w")
b = tf.Variable(tf.zeros(2, dtype=tf.float32), name="b")
x = [[1.0, 2.0, 3.0]]

with tf.GradientTape(persistent=True) as tape:
    y = x @ w + b
    loss = tf.reduce_mean(y**2)

[dl_dw, dl_db] = tape.gradient(loss, [w, b])

doc = MyDoc(x=dl_dw, y=dl_db)
doc.json()

```

### DIFF
--- a/docarray/typing/tensor/torch_tensor.py
+++ b/docarray/typing/tensor/torch_tensor.py
@@ -137,7 +137,7 @@ class TorchTensor(
         Convert `TorchTensor` into a json compatible object
         :return: a representation of the tensor compatible with orjson
         """
-        return self.numpy()  ## might need to check device later
+        return self.detach().numpy()  # might need to check device later
 
     def unwrap(self) -> torch.Tensor:
         """
--- a/tests/units/typing/tensor/test_torch_tensor.py
+++ b/tests/units/typing/tensor/test_torch_tensor.py
@@ -63,7 +63,7 @@ def test_wrong_but_reshapable():
         parse_obj_as(TorchTensor[3, 224, 224], torch.zeros(224, 224))
 
 
-def test_inependent_variable_dim():
+def test_independent_variable_dim():
     # test independent variable dimensions
     tensor = parse_obj_as(TorchTensor[3, 'x', 'y'], torch.zeros(3, 224, 224))
     assert isinstance(tensor, TorchTensor)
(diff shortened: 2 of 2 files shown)


JoanFM (changes requested):
can you please add a test for the case reported in the issue?

docarray/typing/tensor/torch_tensor.py
```diff
@@ -137,7 +137,7 @@ def _docarray_to_json_compatible(self) -> np.ndarray:
         Convert `TorchTensor` into a json compatible object
         :return: a representation of the tensor compatible with orjson
         """
-        return self.numpy()  ## might need to check device later
+        return self.detach().numpy()  # might need to check device later
```
JohannesMessner, line 140: This only seems to tackle json serialization. Do al the other ones work? Protobuf, pickle, etc?

JohannesMessner (changes requested):

tests/units/typing/tensor/test_torch_tensor.py
```diff
@@ -177,3 +177,16 @@ class MMdoc(BaseDoc):
...
     assert not (doc.embedding == doc_copy.embedding).all()
+
+
+def test_serialization():
+    from docarray import BaseDoc
+
+    class MyDoc(BaseDoc):
+        tens: TorchTensor
+
+    doc = MyDoc(tens=torch.rand(10, requires_grad=False))
```
JoanFM, line 188: can we please use `parametrization`? Like this, when a test fails is easier to identify which failed.

Also, can u do some sort of assertion on the output of this `json` method?

JoanFM (changes requested):

JoanFM (approved):

PR #1550, opened 2023-05-17
[wip ]feat: support redis

jupyterjazz:


docarray/index/backends/redis.py
```diff
@@ -93,7 +93,7 @@ def _random_name():
         return uuid.uuid4().hex
 
     def _create_index(self):
-        if not self._check_index_exists(self._db_config.index_name):
+        if not self._check_index_exists(self._db_config.index_name):  # type: ignore[arg-type]
```
JoanFM, line 96: we cannot simply ignore everything

JoanFM (changes requested):

docarray/index/backends/redis.py
jupyterjazz: removed some

JoanFM (changes requested):
Please remove all the typing ignores

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,606 @@
...
+
+        :yield: A list of dictionaries where each item consists of a column name and
+            an item from the corresponding generator. Yields until all generators
+            are exhausted.
+        """
+        keys = list(column_to_data.keys())
+        iterators = [iter(column_to_data[key]) for key in keys]
+        batch: List[Dict[str, Any]] = []
+
+        while True:
```
JoanFM, line 305: is this `while True` needed?

JoanFM (changes requested):

docarray/index/backends/redis.py
jupyterjazz: yes because I don't know the size in advance

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,606 @@
...
+    'DISMAX',
+    'DOCSCORE',
+    'HAMMING',
+]
+
+
+class RedisDocumentIndex(BaseDocIndex, Generic[TSchema]):
+    def __init__(self, db_config=None, **kwargs):
+        """Initialize RedisDocumentIndex"""
+        if db_config is not None and getattr(db_config, 'index_name'):
```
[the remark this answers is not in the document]
jupyterjazz: BaseDocIndex init needs `index_name` to work, that's why I assign it in advance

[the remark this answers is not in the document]
JoanFM: where is this needed?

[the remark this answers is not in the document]
JoanFM: what u should do is make `index_name` propery read it from `db_config`

docarray/index/backends/redis.py
JoanFM: how do u break out of the loop?

JoanFM: Review this because I feel is not needed. it seems that u will infintely loop over the same keys and iterator

JoanFM (changes requested):

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,609 @@
...
+    ) -> Union[List[DocList], List[List[Dict]]]:
+        """
+        Filters the index based on the given batch of filter queries.
+
+        :param filter_queries: The filter conditions.
+        :param limit: The maximum number of results to return for each filter query.
+        :return: Filter results.
+        """
+        results = []
+        for query in filter_queries:
```
JohannesMessner, line 533: Same here, no native batching?

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,609 @@
...
+        """
+        Conducts a batched text-based search on the index.
+
+        :param queries: The queries to search.
+        :param limit: The maximum number of results to return for each query.
+        :param search_field: The field to search the queries.
+        :return: Search results.
+        """
+        docs, scores = [], []
+        for query in queries:
```
JohannesMessner, line 587: same

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,609 @@
...
+        :param item: The document to check.
+            It must be an instance of BaseDoc or its subclass.
+        :return: True if the document exists in the index, False otherwise.
+        """
+        if safe_issubclass(type(item), BaseDoc):
+            return self._doc_exists(item.id)
+        else:
+            raise TypeError(
+                f"item must be an instance of BaseDoc or its subclass, not '{type(item).__name__}'"
+            )
```
JohannesMessner, line 609: Maybe worthy of a separete PR, but I think we should move `__contains__` to the abstract class and tell backends to implement `_doc_exitsts()`.

JohannesMessner (commented):
Some small comments, but looking good!

docarray/index/backends/redis.py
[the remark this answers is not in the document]
jupyterjazz: I still have to check the index name before the init function and create myself in case it's not provided so I don't see why your option is better

It's needed when creating subindices in `_init_subindex`

docarray/index/backends/redis.py
jupyterjazz: One item is created on each iteration, stops when the iterator is exhausted, i.e. `item` is None (in this case I'm checking the iterator for ID because others might have None values)

```python
if key == 'id' and not item:
    if batch:
        yield batch
  return
```

What could be dangerous here?

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,609 @@
...
+
+        return ids
+
+    def num_docs(self) -> int:
+        """
+        Fetch the number of documents in the index.
+
+        :return: Number of documents in the index.
+        """
+        num_docs = self._client.ft(self._index_name).info()['num_docs']
```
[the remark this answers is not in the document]
jupyterjazz: It is, `num_docs` is included in the index info

but apparently there's a better way to get number of docs
`        return self._client.dbsize()
`

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,609 @@
...
+        """
+        Conducts a batched search on the index.
+
+        :param queries: The queries to search.
+        :param limit: The maximum number of results to return for each query.
+        :param search_field: The field to search the queries.
+        :return: Search results.
+        """
+        docs, scores = [], []
+        for query in queries:
```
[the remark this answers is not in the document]
jupyterjazz: nope multiple queries are not supported

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,609 @@
...
+
+        :param doc_ids: A sequence of document ids.
+        :return: A sequence of documents from the index.
+        """
+        if not doc_ids:
+            return []
+
+        ids = [self._prefix + id for id in doc_ids]
+        docs = self._client.json().mget(ids, '$')
+        docs = [doc[0] for doc in docs if doc]
```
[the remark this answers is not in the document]
jupyterjazz: yes but the abstract class expects a list and not a generator, we can refactor that in a separate pr maybe

docarray/index/backends/redis.py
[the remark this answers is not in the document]
jupyterjazz: But I think we should stick with the old way because dbsize returns number of keys the client stores, combining all indices so it's inaccurate if you have more than one index

docarray/index/backends/redis.py
jupyterjazz: https://github.com/docarray/docarray/issues/1686

docarray/index/backends/redis.py
```diff
@@ -0,0 +1,609 @@
...
+                raise ValueError(
+                    f"Invalid text scorer '{self.text_scorer}' provided. "
+                    f"Must be one of: {', '.join(VALID_TEXT_SCORERS)}"
+                )
+
+    @dataclass
+    class RuntimeConfig(BaseDocIndex.RuntimeConfig):
+        """Dataclass that contains all "dynamic" configurations of RedisDocumentIndex.
+
+        :param batch_size: Batch size during indexing.
```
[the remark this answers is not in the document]
jupyterjazz: Yes it's only index currently, I didn't think get and del would need it tbh, but I'll use it there too (search and other ops dont need it)

JoanFM (changes requested):

Issue #1439, opened 2023-04-24
Bug: Type coercion from torch to numpy does not work


JohannesMessner:
**Description:**
casting a TorchTensor NdArray fails

**How to reproduce:**
```python
from docarray import BaseDoc
from docarray.typing import NdArray
import torch

class MyDoc(BaseDoc):
    tens: NdArray

t1 = torch.rand(512)
doc = MyDoc(tens=t1)
```
```bash
  Expected a numpy.ndarray compatible type, got <class 'torch.Tensor'> (type=value_error)
```

**TODO**:
- Fix this in the `validate` function of `NdArray`
- what about coercions to/from tf? do we have tests for this?

JohannesMessner:
Update: the easy way of fixing this produces unwanted side effects. Basicaly, if we make coercion to np too powerful, then it will break `AnyTensor == Union[NdArray, TorchTensor, TensorFlowTensor]`, since it would cast everything to np. This is due to how pydantic resolves validation for unions.

The only solution i see here is defining our own Union with our own validation rules, and then re-defining AnyTensor in terms of that.
This is a bit more involved than the naive fix, so we de-emphasize that for now.

makram93:
pining so that you can assign it to me!

PR #1552, opened 2023-05-17
refactor: make AnyTensor a class

makram93:
This PR solves #1439 .

As mentioned by @JohannesMessner, simply updating the `validate()` method of `NdArray` to support type coercion from Torch/TensorFlow to NumPy produces unwanted side-effects with `AnyTensor`. 

`AnyTensor` is currently defined as: `AnyTensor == Union[NdArray, TorchTensor, TensorFlowTensor]`. `Union` validates the type from left to right, as such if we give torch tensors then it gets converted to NdArray. If we change the order as
`AnyTensor == Union[TorchTensor, TensorFlowTensor, NdArray]` then the NumPy value gets converted to TorchTensor. 

In this PR:
- Make `AnyTensor` a class that inherits from the `AbstractTensors` and implements the `override()` method only. This method basically checks the type of value and returns the class accordingly. 
Note: `AnyTensor` does not inherit `[TorchTensor, TensorflowTensor, NdArray]` neither does it implement any functions from the classes
- Keep the union of tensors inside `TYPE_CHECKING` so that mypy does not complain about the missing attribute.
- Similar changes have been done for `AnyEmbedding`, `AudioTensor`, `ImageTensor`, `VideoTensor`.
- Allow coercion of [`torch.Tensor`, `TorchTensor`, `tf.Tensor`, `TensorFlowTensor`] to np in `NdArray`

### DIFF
--- a/docarray/utils/_internal/_typing.py
+++ b/docarray/utils/_internal/_typing.py
@@ -1,6 +1,7 @@
 from typing import Any, Optional
 
-from typing_inspect import get_args, is_union_type
+from typing_extensions import get_origin
+from typing_inspect import get_args, is_typevar, is_union_type
 
 from docarray.typing.tensor.abstract_tensor import AbstractTensor
 
@@ -32,3 +33,18 @@ def change_cls_name(cls: type, new_name: str, scope: Optional[dict] = None) -> N
         scope[new_name] = cls
     cls.__qualname__ = cls.__qualname__[: -len(cls.__name__)] + new_name
     cls.__name__ = new_name
+
+
+def safe_issubclass(x: type, a_tuple: type) -> bool:
+    """
+    This is a modified version of the built-in 'issubclass' function to support non-class input.
+    Traditional 'issubclass' calls can result in a crash if the input is non-class type (e.g. list/tuple).
+
+    :param x: A class 'x'
+    :param a_tuple: A class, or a tuple of classes.
+    :return: A boolean value - 'True' if 'x' is a subclass of 'A_tuple', 'False' otherwise.
+             Note that if the origin of 'x' is a list or tuple, the function immediately returns 'False'.
+    """
+    if (get_origin(x) in (list, tuple, dict, set)) or is_typevar(x):
+        return False
+    return issubclass(x, a_tuple)
--- a/docarray/index/abstract.py
+++ b/docarray/index/abstract.py
@@ -28,7 +28,7 @@ from docarray import BaseDoc, DocList
 from docarray.array.any_array import AnyDocArray
 from docarray.typing import ID, AnyTensor
 from docarray.typing.tensor.abstract_tensor import AbstractTensor
-from docarray.utils._internal._typing import is_tensor_union
+from docarray.utils._internal._typing import is_tensor_union, safe_issubclass
 from docarray.utils._internal.misc import import_library
 from docarray.utils.find import (
     FindResult,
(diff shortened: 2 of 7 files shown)


docarray/typing/tensor/image/image_tensor.py
```diff
@@ -1,23 +1,130 @@
...
+
+if TYPE_CHECKING:
+    from pydantic import BaseConfig
+    from pydantic.fields import ModelField
+
+
+T = TypeVar("T", bound="ImageTensor")
+
+
+class ImageTensor(AbstractImageTensor):
```
JohannesMessner, line 34: I think all of these generic ImageTensor, AudioTensor can be implemented without repeating every method on them.
Is it possible to just do `class ImageTensor(AnyTensor, AbstractImageTensor)` with an (almost) empty class body?
If not, then I think we should change the inheritance structure of these tensors a bit to enable this. We should not have this copy/pasted code if we can avoid it.

JohannesMessner (changes requested):

JohannesMessner:
@makram93 could you pleas update the description of the PR?
The problem you are solving, and how you are solving it. Thanks you!

docarray/utils/_internal/_typing.py
```diff
@@ -21,6 +21,19 @@ def is_tensor_union(type_: Any) -> bool:
         )
 
 
+def is_tensor_any(first_doc, field_name, field_type, tensor_type) -> bool:
+    """Return True if the `field_name` in the doc is a subclass of the tensor type."""
+    # all generic tensor types such as AnyTensor, AnyEmbedding, ImageTensor, etc. are subclasses of AbstractTensor
+    if issubclass(field_type, AbstractTensor):
```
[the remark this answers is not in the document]
makram93: I just wanted to avoid checking field type that does not inherit `AbstractTensor` but this can be clubbed together with the below check

JohannesMessner (commented):
Big picture looks good, I think this is the way we should solve this!! I will do a more thorough review when I have time (thursday, friday I am not so docarray focused)

docarray/typing/tensor/ndarray.py
```diff
@@ -105,6 +118,14 @@ def validate(
             return cls._docarray_from_native(value)
         elif isinstance(value, NdArray):
             return cast(T, value)
+        elif torch_available and isinstance(value, torch.Tensor):
+            return cls._docarray_from_native(value.detach().cpu().numpy())
+        elif torch_available and isinstance(value, TorchTensor):
```
samsja, line 123: TorchTensor is a torch.Tensor so no need for both I guess ?

docarray/typing/tensor/tensor.py
```diff
@@ -1,22 +1,145 @@
...
+
+        def __iter__(self):
+            pass
+
+        def __len__(self):
+            pass
+
+        @classmethod
+        def _docarray_from_native(cls: Type[T], value: Any):
+            raise AttributeError(f'This method should not be called on {cls}.')
```
samsja, line 95: Maybe RuntimeError is more clear, Technically this method is nto an Attribute right ?

samsja (commented):
Looks good, nice trick ! I added minor comments

docarray/typing/tensor/tensor.py
makram93: Technically, [methods are attributes](https://docs.python.org/3/tutorial/classes.html#:~:text=By%20the%20way%2C%20I%20use%20the%20word%20attribute%20for%20any%20name%20following%20a%20dot) of class instance but you are right, this error will pop up during runtime so raising it as `RuntimeError` should be more apt

samsja (approved):
Looks very good !

PR #1559, opened 2023-05-22
fix: dict method for document view

JohannesMessner:
Before this, calling `doc.dict()` when `doc.is_view() == True` would return an empty dict.
The core issue is that the `ColumnStorageView` did not behave like a proper dict: `.values()` and `.keys()` were not implemented, and would return empty values.
That is fixed in this PR.

### DIFF
--- a/docarray/array/doc_vec/column_storage.py
+++ b/docarray/array/doc_vec/column_storage.py
@@ -3,6 +3,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
+    ItemsView,
     Iterable,
     MutableMapping,
     Optional,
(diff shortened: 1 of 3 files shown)


docarray/array/doc_vec/column_storage.py
```diff
@@ -141,5 +143,16 @@ def __iter__(self):
...
 
+    def _local_dict(self):
+        """The storage.columns dictionary with every value at position self.index"""
+
+        return {key: self[key] for key in self.storage.columns.keys()}
+
     def keys(self):
         return self.storage.columns.keys()
+
+    def values(self):
```
samsja, line 154: can we add type hint for the return type ?

docarray/array/doc_vec/column_storage.py
```diff
@@ -141,5 +143,16 @@ def __iter__(self):
...
+        return ValuesView(self._local_dict())
+
+    def items(self):
```
samsja, line 157: same tipe hint here

docarray/array/doc_vec/column_storage.py
```diff
@@ -3,12 +3,14 @@
...
     Any,
     Dict,
+    ItemsView,
     Iterable,
     MutableMapping,
     Optional,
     Type,
     TypeVar,
     Union,
+    ValuesView,
```
samsja, line 13: same here

samsja (commented):
looks good

docarray/array/doc_vec/column_storage.py
```diff
@@ -3,12 +3,14 @@
     TYPE_CHECKING,
     Any,
     Dict,
+    ItemsView,
```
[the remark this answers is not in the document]
JohannesMessner: It takes a dict and gives you a view of it as iterable of tuples. It is the closes thing to the "real" return type `dict_values` which is private and thus not usable. More details here: https://github.com/python/typing/discussions/1033

docarray/array/doc_vec/column_storage.py
JohannesMessner: same as for itemsview :)

samsja (approved):

PR #1562, opened 2023-05-22
feat: i/o for DocVec

### LINKED ISSUE #1330: IOMixin for DocVec (to/from)
We want to support IO functionality in `DocVec` such as from/to json, csv, bytes, etc. For example, to/from json is needed to make it FastAPI-compatible.

However, we can not use the existing IOMixin here as the logic behind DocVec is different from DocList and we'd have to overwrite some functions. 

Solution:
Create an abstract mixin containing functions that the two classes share, and for other functions create separate mixins inheriting from the abstract class

JohannesMessner:
This PR implements the IOMixinArray for DocVec, enabling the following (de)serializations:

- [x] bytes
- [x] csv
- [x] json
- [x] pandas dataframe
- [x] proto (https://github.com/docarray/docarray/pull/1639)
  - [x] tensor_type for proto (https://github.com/docarray/docarray/pull/1645)
- [x] save/load
- [x] base64
- [x] Documentation

Depends on https://github.com/docarray/docarray/issues/1561 (protobuf serialization bug)

JSON data now contains four columns, like this:
```
{"tensor_columns":{"tens":[[0.18736880416567359,0.9418134463574087,0.23608414651162635,0.200237847284912,0.714415793571785],[0.2980003926683161,0.9763157895635218,0.605654616422409,0.3988857003524561,0.9359414534852715],[0.6382713675712178,0.7725828308842624,0.1442529379493993,0.9327649675846935,0.7312168445052849],[0.07432756375020888,0.84017563415619,0.895668448511279,0.636209667658674,0.299134816130166],[0.1693478076528051,0.039113459980241405,0.3005636243582387,0.06480147049887997,0.09461133654601517]],"tens_none":null},"doc_columns":{"inner":{"tensor_columns":{"tens":[[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915],[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915],[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915],[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915],[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915]]},"doc_columns":{},"docs_vec_columns":{"tens":[[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915],[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915],[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915],[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915],[0.27449474647200733,0.10957290479279247,0.6840416906244843,0.02922130361087716,0.1610742166462915]]},"any_columns":{"id":["1fa4159e51fdfa029b2e2bbb45be0e58","1fa4159e51fdfa029b2e2bbb45be0e58","1fa4159e51fdfa029b2e2bbb45be0e58","1fa4159e51fdfa029b2e2bbb45be0e58","1fa4159e51fdfa029b2e2bbb45be0e58"]}},"inner_none":null},"docs_vec_columns":{"tens":[[0.18736880416567359,0.9418134463574087,0.23608414651162635,0.200237847284912,0.714415793571785],[0.2980003926683161,0.9763157895635218,0.605654616422409,0.3988857003524561,0.9359414534852715],[0.6382713675712178,0.7725828308842624,0.1442529379493993,0.9327649675846935,0.7312168445052849],[0.07432756375020888,0.84017563415619,0.895668448511279,0.636209667658674,0.299134816130166],[0.1693478076528051,0.039113459980241405,0.3005636243582387,0.06480147049887997,0.09461133654601517]],"tens_none":null},"any_columns":{"id":["5f2575c95a52380be1a9a68e26cd19c5","a5015ccad05d1b7ead1dce92650e5e8f","f14b45ea6384f235f990f40cddf39444","fe0004643fb7aadb7813438727f42dd1","489f382c9887c3fe3ef075a735722ff4"],"text":["0","1","2","3","4"],"num":[null,null,null,null,null]}}
```

CSV is not supported, since it is a row-based format.

All other formats are used the same way as in DocList, but have a different serialized representation.

Closes #1330


docarray/array/doc_list/io.py
```diff
@@ -377,13 +377,12 @@ def from_csv(
     @classmethod
     def _from_csv_file(
         cls, file: Union[StringIO, TextIOWrapper], dialect: Union[str, csv.Dialect]
```
samsja, line 379:
```suggestion
        cls: Type[T], file: Union[StringIO, TextIOWrapper], dialect: Union[str, csv.Dialect]
```

docarray/array/doc_list/io.py
```diff
@@ -443,7 +442,7 @@ def to_csv(
                 writer.writerow(doc_dict)
 
     @classmethod
-    def from_dataframe(cls, df: 'pd.DataFrame') -> 'DocList':
+    def from_dataframe(cls, df: 'pd.DataFrame') -> 'T':
```
samsja, line 445:
```suggestion
    def from_dataframe(cls: Type[T], df: 'pd.DataFrame') -> 'T':
```

docarray/array/doc_list/io.py
```diff
@@ -515,6 +514,8 @@ class Person(BaseDoc):
             doc_dict = _access_path_dict_to_nested_dict(access_path2val)
             docs.append(doc_type.parse_obj(doc_dict))
 
+        if not isinstance(docs, cls):
+            return cls(docs)
```
samsja, line 518: How could it not be from instance `cls` ?

docarray/array/doc_vec/doc_vec.py
```diff
@@ -156,7 +157,7 @@ class DocVec(AnyDocArray[T_doc]):
         AnyTensor or Union of NdArray and TorchTensor
     """
 
-    doc_type: Type[T_doc]
+    doc_type: Type[T_doc] = AnyDoc  # TODO(johannes): should this be BaseDoc?
```
samsja, line 160: hmmm Maybe yes by default it should be BaseDoc since DocVec cannot be instantiate with AnyDoc

samsja (changes requested):
Lets discuss I don't think we should implement this that way.

If I am understanding correctly here the json representation of a DocList and a DocVec is the same right ?

IMO DocVec should have its own column based json representation. 

For csv etcc I would say we ask first user to instantiate a DocList then cast to docVec

docarray/array/doc_list/io.py
JohannesMessner: `docs` can also be a list of Documents etc.

JohannesMessner:
> IMO DocVec should have its own column based json representation.

Fair

> For csv etcc I would say we ask first user to instantiate a DocList then cast to docVec

Why? The whole point of this issue is to natively enable these, no? Why do you think DocVec should not have these methods?

samsja:
> > IMO DocVec should have its own column based json representation.
> 
> Fair
> 
> > For csv etcc I would say we ask first user to instantiate a DocList then cast to docVec
> 
> Why? The whole point of this issue is to natively enable these, no? Why do you think DocVec should not have these methods?

My point is that saving line by line in csv for instance is a bit weird for a column base format. But maybe yes we can have them, I just don't want it to be confusing

JohannesMessner:
> > > IMO DocVec should have its own column based json representation.
> > 
> > 
> > Fair
> > > For csv etcc I would say we ask first user to instantiate a DocList then cast to docVec
> > 
> > 
> > Why? The whole point of this issue is to natively enable these, no? Why do you think DocVec should not have these methods?
> 
> My point is that saving line by line in csv for instance is a bit weird for a column base format. But maybe yes we can have them, I just don't want it to be confusing

yeah this is fair, I will address that

JoanFM (approved):
Looks very nice!

Issue #1545, opened 2023-05-16
Validation bug: DocList and DocVec are not coerced to each other


JohannesMessner:
In the spirit of pydantic, a model (document) field that is annotated with a certain type should only ever hold data that is actually of that type.
To achieve that, incoming data of a different type should be coerced to the target type.

This does not happen between `DocList` and `DocVec`:

```python
from docarray import BaseDoc, DocList, DocVec
from docarray.documents import ImageDoc


class Doc(BaseDoc):
    docs: DocList[ImageDoc]

d = Doc(docs=DocVec[ImageDoc]([ImageDoc()]))
# below should give <DocList[ImageDoc] (length=1)> but give <DocVec[ImageDoc] (length=1)>
print(d.docs)


class OtherDoc(BaseDoc):
    docs: DocVec[ImageDoc]

d = Doc(docs=DocList[ImageDoc]([ImageDoc()]))
# below should give <DocVec[ImageDoc] (length=1)> but give <DocList[ImageDoc] (length=1)>
print(d.docs)

```

agaraman0:
Hey @JohannesMessner I was trying to figure out where type-casting logic has to be written, I checked the `_validate_docs` function for a brief idea but did not get much context.

agaraman0:
got it

PR #1568, opened 2023-05-24
Validation bug Fix: DocList and DocVec are not coerced to each other

agaraman0:
fixes issue #1545 

`
from docarray import BaseDoc, DocList, DocVec
from docarray.documents import ImageDoc

class Doc(BaseDoc):
    docs: DocList[ImageDoc]

d = Doc(docs=DocVec[ImageDoc]([ImageDoc()]))
# below gives <DocList[ImageDoc] (length=1)>
print(d.docs)

class OtherDoc(BaseDoc):
    docs: DocVec[ImageDoc]

d = Doc(docs=DocList[ImageDoc]([ImageDoc()]))
# below gives <DocVec[ImageDoc] (length=1)>
print(d.docs)

`

### DIFF
--- a/docarray/array/doc_list/doc_list.py
+++ b/docarray/array/doc_list/doc_list.py
@@ -267,8 +267,10 @@ class DocList(
     ):
         from docarray.array.doc_vec.doc_vec import DocVec
 
-        if isinstance(value, (cls, DocVec)):
+        if isinstance(value, cls):
             return value
+        elif isinstance(value, DocVec):
+            return value.to_doc_list()
         elif isinstance(value, cls):
             return cls(value)
         elif isinstance(value, Iterable):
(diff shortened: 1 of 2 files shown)


docarray/array/doc_list/doc_list.py
```diff
@@ -267,8 +267,10 @@ def validate(
     ):
         from docarray.array.doc_vec.doc_vec import DocVec
 
-        if isinstance(value, (cls, DocVec)):
+        if isinstance(value, cls):
             return value
+        elif isinstance(value, DocVec):
+            return value.to_doc_list()
```
samsja, line 273: here we need to test if the documen_type of the value and self are compatible.

If they are you can silence mypy by return  `cast(T, )`

samsja, line 273:
```suggestion
        elif isinstance(value, DocVec):
            if issubclass(value.doc_type, self.doc_type) or value.doc_type == cls.doc_type:
                return cast(T, value.to_doc_list()
            else:
                raise ValueError(f'DocList[value.doc_type] is not compatible with {cls}')    
```

samsja: smth like this

samsja (approved):

Issue #1557, opened 2023-05-21
Not able to index when using List[str] in custom document class


lazyhope:
System: `macOS 13.3.1`
Python version: `3.11.0`
IPython version: `8.10.0`

In the latest docarray version, when building hnsw index from the following simple document class:
```python
from docarray import BaseDoc
from docarray.index import HnswDocumentIndex

class MyDoc(BaseDoc):
    test: List[str]

index = HnswDocumentIndex[MyDoc]()
```
The following error will pop up:
```
File ~/miniconda3/envs/test/lib/python3.11/site-packages/docarray/index/backends/hnswlib.py:76, in HnswDocumentIndex.__init__(self, db_config, **kwargs)
     73 if db_config is not None and getattr(db_config, 'index_name'):
     74     db_config.work_dir = db_config.index_name.replace("__", "/")
---> 76 super().__init__(db_config=db_config, **kwargs)
     77 self._db_config = cast(HnswDocumentIndex.DBConfig, self._db_config)
     78 self._work_dir = self._db_config.work_dir

File ~/miniconda3/envs/test/lib/python3.11/site-packages/docarray/index/abstract.py:111, in BaseDocIndex.__init__(self, db_config, subindex, **kwargs)
    109 self._runtime_config = self.RuntimeConfig()
    110 self._logger.info('Runtime config created')
--> 111 self._column_infos: Dict[str, _ColumnInfo] = self._create_column_infos(
    112     self._schema
    113 )
    114 self._is_subindex = subindex
    115 self._subindices: Dict[str, BaseDocIndex] = {}

File ~/miniconda3/envs/test/lib/python3.11/site-packages/docarray/index/abstract.py:880, in BaseDocIndex._create_column_infos(self, schema)
    873 """Collects information about every column that is implied by a given schema.
    874
    875 :param schema: The schema (subclass of BaseDoc) to analyze and parse
    876     columns from
    877 :returns: A dictionary mapping from column names to column information.
    878 """
    879 column_infos: Dict[str, _ColumnInfo] = dict()
--> 880 for field_name, type_, field_ in self._flatten_schema(schema):
    881     # Union types are handle in _flatten_schema
    882     if issubclass(type_, AnyDocArray):
    883         column_infos[field_name] = _ColumnInfo(
    884             docarray_type=type_, db_type=None, config=dict(), n_dim=None
    885         )

[...]

PR #1569, opened 2023-05-24
fix: support list in document class (#1557)

maxwelljin:
This pull request resolves the TypeError in BaseDocIndex._flatten_schema that occurs when a non-class type, such as List[str], is used as an argument for issubclass(). As observed in issue #1557, this causes the HnswDocumentIndex initialisation to crash.

As Johannes suggested, I use a custom issubclass function that can handle non-class types by returning False instead of failing. This change prevents crashes when a non-class type is passed as input.

### DIFF
--- a/docarray/index/abstract.py
+++ b/docarray/index/abstract.py
@@ -390,7 +390,7 @@ class BaseDocIndex(ABC, Generic[TSchema]):
         for field_name, type_, _ in self._flatten_schema(
             cast(Type[BaseDoc], self._schema)
         ):
-            if issubclass(type_, AnyDocArray):
+            if safe_issubclass(type_, AnyDocArray):
                 for doc_id in key:
                     nested_docs_id = self._subindices[field_name]._filter_by_parent_id(
                         doc_id
--- a/docarray/index/abstract.py
+++ b/docarray/index/abstract.py
@@ -18,6 +18,7 @@ from typing import (
     TypeVar,
     Union,
     cast,
+    get_origin,
 )
 
 import numpy as np
@@ -707,6 +708,21 @@ class BaseDocIndex(ABC, Generic[TSchema]):
     ##########################################################
 
     @staticmethod
+    def custom_issubclass(x, A_tuple) -> bool:
+        """
+        This is a modified version of the built-in 'issubclass' function to support non-class input.
+        Traditional 'issubclass' calls can result in a crash if the input is list/tuple.
+
+        :param x: A class 'x'
+        :param A_tuple: A class, or a tuple of classes.
+        :return: A boolean value - 'True' if 'x' is a subclass of 'A_tuple', 'False' otherwise.
+                 Note that if the origin of 'x' is a list or tuple, the function immediately returns 'False'.
+        """
+        if get_origin(x) in (list, tuple):
+            return False
+        return issubclass(x, A_tuple)
+
+    @staticmethod
     def _get_values_by_column(docs: Sequence[BaseDoc], col_name: str) -> List[Any]:
         """Get the value of a column of a document.
 
@@ -776,7 +792,7 @@ class BaseDocIndex(ABC, Generic[TSchema]):
         for field_name, type_, _ in self._flatten_schema(
             cast(Type[BaseDoc], self._schema)
         ):
-            if issubclass(type_, AnyDocArray):
+            if self.custom_issubclass(type_, AnyDocArray):
                 for doc in docs:
                     _list = getattr(doc, field_name)
                     for i, nested_doc in enumerate(_list):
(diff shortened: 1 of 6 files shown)


JoanFM (changes requested):
Please, add tests showing the underlying issue is fixed

docarray/index/abstract.py
```diff
@@ -776,7 +792,7 @@ def _update_subindex_data(
         for field_name, type_, _ in self._flatten_schema(
             cast(Type[BaseDoc], self._schema)
         ):
-            if issubclass(type_, AnyDocArray):
+            if self.custom_issubclass(type_, AnyDocArray):
```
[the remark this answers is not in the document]
maxwelljin: thanks for pointing it out. will fix it

docarray/index/abstract.py
```diff
@@ -706,6 +707,21 @@ def _filter_by_parent_id(self, id: str) -> Optional[List[str]]:
     # These might be useful in your subclass implementation  #
     ##########################################################
 
+    @staticmethod
+    def custom_issubclass(x, A_tuple) -> bool:
```
JohannesMessner, line 711: Let's make this a function instead of a staticmethod, and let's more it to `utils._internal._typing`.
Maybe we can also rename it to `safe_issubclass` to indicate a bit more about what makes it custom?

docarray/index/abstract.py
```diff
@@ -706,6 +707,21 @@ def _filter_by_parent_id(self, id: str) -> Optional[List[str]]:
...
+        """
+        This is a modified version of the built-in 'issubclass' function to support non-class input.
+        Traditional 'issubclass' calls can result in a crash if the input is list/tuple.
+
+        :param x: A class 'x'
+        :param A_tuple: A class, or a tuple of classes.
+        :return: A boolean value - 'True' if 'x' is a subclass of 'A_tuple', 'False' otherwise.
+                 Note that if the origin of 'x' is a list or tuple, the function immediately returns 'False'.
+        """
+        if get_origin(x) in (list, tuple):
```
JohannesMessner, line 721: What about other origins? I think Dict for example applies in the same way.
Also, are there other things that aren't classes (= break normal issubclass) but also don't have an origin? TypeVars for example? Would be nice to cover as many cases as possible here.

docarray/index/abstract.py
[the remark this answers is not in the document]
JohannesMessner: Actually I think this should not even be a method, as I say above. That way we can use it in all kinds of places, I don't see a reason why this needs to be tied to BaseDocIndex.

docarray/index/abstract.py
```diff
@@ -706,6 +707,21 @@ def _filter_by_parent_id(self, id: str) -> Optional[List[str]]:
...
     ##########################################################
 
+    @staticmethod
+    def custom_issubclass(x, A_tuple) -> bool:
+        """
+        This is a modified version of the built-in 'issubclass' function to support non-class input.
+        Traditional 'issubclass' calls can result in a crash if the input is list/tuple.
+
+        :param x: A class 'x'
+        :param A_tuple: A class, or a tuple of classes.
```
JohannesMessner, line 717: let's make this argument lowercase, i.e. `a_tuple`

docarray/index/abstract.py
```diff
@@ -706,6 +707,21 @@ def _filter_by_parent_id(self, id: str) -> Optional[List[str]]:
     # These might be useful in your subclass implementation  #
     ##########################################################
 
+    @staticmethod
+    def custom_issubclass(x, A_tuple) -> bool:
+        """
+        This is a modified version of the built-in 'issubclass' function to support non-class input.
+        Traditional 'issubclass' calls can result in a crash if the input is list/tuple.
```
JohannesMessner, line 714: Nice docstring!

JohannesMessner (changes requested):

maxwelljin:
Thanks for the feedback! I have made the following changes:

I have added tests (docarray/tests/hnswlib/test_index_get_del.py) to demonstrate that the issue is fixed. These tests verify the functionality and ensure the problem is resolved.
I have converted the method to a function and moved it to utils._internal._typing, as suggested. Additionally, I renamed it to "safe_issubclass" to provide a clearer indication of its purpose.
I have also included support for other origins like "Dict" in addition to "list" and "tuple" and handled cases like "TypeVars" that are not classes but still need to be checked.

Once again, thank you for the valuable feedback

docarray/utils/_internal/_typing.py
```diff
@@ -32,3 +32,17 @@ def change_cls_name(cls: type, new_name: str, scope: Optional[dict] = None) -> N
         scope[new_name] = cls
     cls.__qualname__ = cls.__qualname__[: -len(cls.__name__)] + new_name
     cls.__name__ = new_name
+
+def safe_issubclass(x, a_tuple) -> bool:
```
JoanFM, line 36: Please add unittests specific to this method, also add proper type hinting

JoanFM (changes requested):

JoanFM (approved):

PR #1572, opened 2023-05-24
docs: add mention and example for type coercion

Tanguyabel:
This PR brings changes to the docs where we mention type coercion for Tensor, with an example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,39 @@ class MyDocument(BaseDoc):
 
 So not only can you define the types of your data, you can even **specify the shape of your tensors!**
 
+> :bulb: **Type coercion:** When submitting a tensor-like structure to a tensor field, the resulting `TorchTensor` is typed after the provided data.
+
+See:
+
+```python
+from docarray import BaseDoc
+from docarray.typing import TorchTensor
+import numpy as np
+
+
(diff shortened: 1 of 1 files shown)


samsja (changes requested):
looks good.

Tho readme is not the right section for this. 


IMO the best would be that you put it in a Tensor section here : https://docs.docarray.org/data_types/first_steps/

Tanguyabel:
> looks good.
> 
> Tho readme is not the right section for this.
> 
> IMO the best would be that you put it in a Tensor section here : https://docs.docarray.org/data_types/first_steps/

@samsja I understand what you say. I added the mention there because:
- it is in this openable section "Click for more details", where we already mention that one can specify the shape of a tensor, so I thought it could go well together, as it is kinda related
- in the data type section that you mention, there is no subsection related to a generic tensor, instead we focus on specific modality; however, type coercion applies to any kind of tensor. It also looks a bit too much detailed for the data type intro page imo

Any suggestion?

makram93 (commented):
should we also highlight type coercion to AnyTensor in this doc?

samsja:
Then lets move it to the BaseDoc section at the end. I don't think we should overload the readme

PR #1587, opened 2023-05-30
fix: make example payload a string and not bytes

JoanFM:
Before:

```python

from docarray import DocList, BaseDoc
from docarray.documents import TextDoc
import numpy as np


class MyDoc(BaseDoc):
    text: str
    embedding: NdArray[128]

print(f'{type(MyDoc.schema()["properties"]["embedding"]["example"])}')
```

This would show `bytes`.

Now this would show `str`


samsja (approved):
good catch

Issue #1555, opened 2023-05-19
Url types are not aware of extension during validation


alaeddine-13:
Url types do not perform validation on the url extension. For instance, if a url has extension to an audio file (.wav), ImageURL can still accept it.

in the screenshot below, when a field is defined like this `item: Union[ImageUrl, AudioUrl, str]`, when it is initialized like this : MyClass(item='link to audio') it will be validated against a ImageUrl. This is because ImageUrl considers a .wav file a valid image

![Image](https://github.com/docarray/docarray/assets/15269265/81369255-f062-48b6-953b-42ceea4380e5)

makram93:
pinging to pick this up

PR #1606, opened 2023-05-31
feat: validate file formats in url

makram93:
Solves: #1555 

Solution:
- rely on the `mime-types` library to detect the mime type from the URL and raise an error if it is not the expected type for that class
- if auto-detection of mime-type fails then we check if the file ending is among the `extra extensions`
  - Check for exceptional cases, if there is still no match then raise the error

CAVEAT: Any file without an extension will raise an error, for example, a `wav` audio file with no extension will raise an error.


JoanFM:
I think doing a request to the URL is an overkill and it will make it very slow

docarray/typing/url/any_url.py
```diff
@@ -61,6 +70,33 @@ def validate(
 
         url = super().validate(abs_path, field, config)  # basic url validation
 
+        # Use mimetypes to validate file formats
+        mimetype, encoding = mimetypes.guess_type(value.split("?")[0])
+        if not mimetype:
+            # try reading from the request headers if mimetypes failed - could be slow
+            try:
```
JoanFM, line 77: avoid this step

JoanFM (changes requested):

docarray/typing/url/any_url.py
makram93: okay, will remove it!

JoanFM (approved):

PR #1611, opened 2023-06-01
fix: hnswlib must be able to search with limit more than num docs

JoanFM:
Before, running `find` or `find_batched` on a HNSWLibIndex, the search would raise an Exception from the hnswlib. This PR fixes it by limiting the `limit` parameter to the actual number of Documents in the index


jupyterjazz (approved):
nice, are other backends already following this behavior?

PR #1615, opened 2023-06-02
docs: add forward declaration steps to example to avoid pickling error

JoanFM:


docs/how_to/multimodal_training_and_serving.md
```diff
@@ -135,6 +135,18 @@ class PairTextImage(BaseDoc):
     image: ImageDoc
 ```
 
+You then need to forward declare the following types, this will allow the objects to be properly pickled and unpickled.
```
alexcg1, line 138: What does "forward declare" mean? Declare beforehand?

JoanFM: `forward declaration` is a common term in programming to say that u make aware to the context the existence of a symbol, method, variable, etc... before it exists. Like for some compilers and so it is necessary, it is necessary here for the pickler to know that that class exist.

The thing is that `DocVec[specific_document]` is a dynamically created class, and here we are saying to the modules that such a class exists

alexcg1, line 138:
```suggestion
You then need to forward declare the following types. This will allow the objects to be properly pickled and unpickled.
```

alexcg1 (changes requested):

samsja (approved):

PR #1617, opened 2023-06-05
feat: support subindex on ExactNNSearch

maxwelljin:
Currently, the subindex feature is not available on exact in-memory NNIndexer. This PR aims to support nested documents for ExactNNIndexer

### DIFF
--- a/tests/index/in_memory/test_persist_data.py
+++ b/tests/index/in_memory/test_persist_data.py
@@ -25,7 +25,8 @@ class MyDoc(BaseDoc):
     my_tens: NdArray[30]
 
 
-def test_subindex_index(tmp_path):
+@pytest.fixture
+def nested_doc():
     my_docs = [
         MyDoc(
             id=f'{i}',
@@ -67,11 +68,14 @@ def test_subindex_index(tmp_path):
         )
         for i in range(5)
     ]
+    return my_docs
 
+
+def test_persist_restore(nested_doc, tmp_path):
     stored_path = str(tmp_path) + "/in_memory_index.bin"
 
     index = InMemoryExactNNIndex[MyDoc]()
-    index.index(my_docs)
+    index.index(nested_doc)
 
     assert index.num_docs() == 5
     assert index._subindices['docs'].num_docs() == 25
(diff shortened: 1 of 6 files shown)


maxwelljin:
This PR introduces support for sub-indexing in the exactNNSearch by enabling nested documents. Here's a summary of the changes:
1. Abstract class methods are utilized to create a parent_id / initialization for each nested document. 
2. The backend stores everything in the new_schema type, including the parent_id, but user need raw type. To handle this, the PR includes helper methods to convert the raw type back into the original type that the user inputted to the Indexer.
3. To ensure that the data type internally stored in the backend is not changed during casting, a copy method is included. The built-in document copy method does not support PyTorch tensors, so a new shallow copy method was created and tested.
4. The PR also includes test code, which is similar to the tests for other backends.
These changes aim to enhance the functionality of the exactNNSearch by supporting nested documents.

JoanFM (changes requested):
what happens when I call persist, does it persist and reload all the subindices?

JoanFM:
> > 
> 
> I'll take a look on that. All doc contents should be persist. If the load_binary function call the initialization, it should reconstruct subindices. I'll try some examples

And test it

maxwelljin:
This new commit supports the persist storage on subindex document and includes the relevant test.

tests/index/in_memory/test_persist_data.py
```diff
@@ -0,0 +1,105 @@
...
+    list_tens: NdArray[20]
+
+
+class MyDoc(BaseDoc):
+    docs: DocList[SimpleDoc]
+    list_docs: DocList[ListDoc]
+    my_tens: NdArray[30]
+
+
+def test_subindex_index(tmp_path):
```
JoanFM, line 28: I think we should also check that find and find_subindex work

JoanFM (changes requested):

tests/index/in_memory/test_persist_data.py
maxwelljin: Sure, I'll add that

JoanFM (approved):

PR #1621, opened 2023-06-06
refactor: minor changes in weaviate

jupyterjazz:
https://github.com/docarray/docarray/issues/1620

two non-breaking changes:
1. build() is more flexible and follows abstract signature
2. query_builder's `find()` can receive search_field even though it won't / can't use it

### DIFF
--- a/docarray/index/backends/weaviate.py
+++ b/docarray/index/backends/weaviate.py
@@ -332,20 +332,19 @@ class WeaviateDocumentIndex(BaseDocIndex, Generic[TSchema]):
         query: Union[AnyTensor, BaseDoc],
         search_field: str = '',
         limit: int = 10,
-        **kwargs,
+        **kwargs: Any,
     ):
         """
         Find k-nearest neighbors of the query.
 
         :param query: query vector for KNN/ANN search. Has single axis.
-        :param search_field: name of the field to search on
         :param limit: maximum number of documents to return per query
         :return: a named tuple containing `documents` and `scores`
         """
         self._logger.debug('Executing `find`')
-        if search_field != '':
-            raise ValueError(
-                'Argument search_field is not supported for WeaviateDocumentIndex.\nSet search_field to an empty string to proceed.'
+        if kwargs.get('search_field'):
+            logging.warning(
+                'The search_field argument is not supported for the WeaviateDocumentIndex and will be ignored.'
             )
         embedding_field = self._get_embedding_field()
         if isinstance(query, BaseDoc):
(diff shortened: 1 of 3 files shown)


JoanFM (changes requested):
I think this is too much of a breaking change, what is the justification?

JoanFM (changes requested):
I think this is too much of a breaking change, what is the justification?

jupyterjazz:
@JoanFM can you check now?

For most of the backends arg names is `filter_query` but for elastic it's `query` and for weaviate it's `where_filter`, I think we should keep interface same where we can and this is one clear example. But I can pass filters without specifying arg name so it's not really needed for langchain stuff.

samsja:
@jupyterjazz can u update the description ? If they are breaking change please state them clearly with pro and cons in doing them

docarray/index/backends/weaviate.py
```diff
@@ -831,15 +831,22 @@ def find(
...
             :param query: query vector for search. Has single axis.
             :param score_name: either `"certainty"` (default) or `"distance"`
             :param score_threshold: the threshold of the score
+            :param search_field: name of the field to search on
             :return: self
             """
+            if search_field != '':
+                raise ValueError(
+                    'Argument search_field is not supported for WeaviateDocumentIndex.\nSet search_field to an empty string to proceed.'
+                )
```
samsja, line 848: then why do we add it ?

jupyterjazz: 1. I want to be able to call find(query=query_emb, search_field=self.search_field) for every backend
2. We already do this for a separate `find()` method in weaviate

jupyterjazz:
@samsja description updated, there are no breaking changes anymore

docarray/index/backends/weaviate.py
samsja: okay I see

samsja (approved):

JoanFM (changes requested):

docarray/index/backends/weaviate.py
```diff
@@ -831,15 +831,22 @@ def find(
...
             score_threshold: Optional[float] = None,
+            search_field: str = '',
         ) -> Any:
             """
             Find k-nearest neighbors of the query.
 
             :param query: query vector for search. Has single axis.
             :param score_name: either `"certainty"` (default) or `"distance"`
             :param score_threshold: the threshold of the score
+            :param search_field: name of the field to search on
```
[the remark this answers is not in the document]
jupyterjazz: in that case let's give a warning instead of a value error?

JoanFM (approved):

samsja:
> 



> @samsja
> 
> that was a possibility even before my changes (find and find_batched already accepted kwargs)
> 
> are you suggesting to remove kwargs from those fns completely?

yes you can inspect the kwargs keys, if you find anything that is not `search_field` you raise an error

PR #1631, opened 2023-06-07
feat: add partial support for pydantic v2

samsja:
# Context

this PR add partial support for Pydantic v2. Most likely this PR is a first draft trying to see how docarray play with pydantic v2. 

The real PR that will merge pydantic v2 support will allow to use both v1 and v2 at the same time


samsja:
This is how pydantic v1 and v2 support could look like internaly

```python
import pydantic
pydantic_v2 = pydantic.__version__.startswith('2')
print(pydantic_v2)

from typing import Any
from pydantic import BaseModel

if pydantic_v2:
    from pydantic import GetCoreSchemaHandler
    from pydantic_core import core_schema
else:
    from pydantic import BaseConfig
    from pydantic.fields import ModelField


class MyType:

    def __init__(self, value = 0):
        self.value = value

    if pydantic_v2:
        @classmethod
        def validate(cls, value: Any, _: Any) -> 'MyType':
            if value > 0:
                return cls(value)
            else:
                raise ValueError('value must be > 0')
    else:
        @classmethod
        def validate(cls, value: Any, field: 'ModelField',config: 'BaseConfig') -> 'MyType':
            if value > 0:
                return cls(value)
            else:
                raise ValueError('value must be > 0')


    if pydantic_v2:
        @classmethod
        def __get_pydantic_core_schema__(
            cls, _source_type: Any, _handler: GetCoreSchemaHandler
        ) -> core_schema.CoreSchema:
            return core_schema.general_plain_validator_function(
                cls.validate,
            )
    else:
        @classmethod
        def __get_validators__(cls):
            # one or more validators may be yielded which will be called in the
            # order to validate the input, each validator will receive as an input
            # the value returned from the previous validator
            yield cls.validate


class MyModel(BaseModel):
    my_type: MyType

m = MyModel(my_type=1)
```

PR #1633, opened 2023-06-09
feat: check if a document is already index

maxwelljin:
This PR is designed to enable the indexer (for all supported backends) to check whether a document has already been indexed. We aim to accommodate various backends including in_memory, hnswlib, elastic, qdrant, and weaviate. Given that different backends store and index documents in distinct ways, we need to custom-tailor our function for each backend.

Progress:

- [x] In-Memory
- [x] Hnswlib
- [x] Elastic
- [x] Qdrant
- [x] Weaviate

### DIFF
--- a/docarray/index/abstract.py
+++ b/docarray/index/abstract.py
@@ -1167,15 +1167,6 @@ class BaseDocIndex(ABC, Generic[TSchema]):
             )
             return self._get_root_doc_id(cur_root_id, root, '')
 
-    @abstractmethod
-    def __contains__(self, item: BaseDoc) -> bool:
-        """Checks if a given BaseDoc item is contained in the index.
-
-        :param item: the given BaseDoc
-        :return: if the given BaseDoc item is contained in the index
-        """
-        ...
-
     def subindex_contains(self, item: BaseDoc) -> bool:
         """Checks if a given BaseDoc item is contained in the index or any of its subindices.
 
--- a/docarray/index/abstract.py
+++ b/docarray/index/abstract.py
@@ -1167,6 +1167,7 @@ class BaseDocIndex(ABC, Generic[TSchema]):
             )
             return self._get_root_doc_id(cur_root_id, root, '')
 
+    @abstractmethod
     def __contains__(self, item: BaseDoc) -> bool:
         """Checks if a given BaseDoc item is contained in the index.
 
@@ -1175,6 +1176,7 @@ class BaseDocIndex(ABC, Generic[TSchema]):
         """
         ...
 
+    @abstractmethod
     def _get_all_documents(self) -> Union[AnyDocArray, List]:
         """Retrieve all documents from the index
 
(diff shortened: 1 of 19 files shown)


docarray/index/backends/hnswlib.py
```diff
@@ -392,6 +392,17 @@ def _get_items(self, doc_ids: Sequence[str], out: bool = True) -> Sequence[TSche
...
+    def __contains__(self, item: BaseDoc):
+        if safe_issubclass(type(item), BaseDoc):
+            hash_id = self._to_hashed_id(item.id)
+            self._sqlite_cursor.execute(
+                f"SELECT data FROM docs WHERE doc_id = '{hash_id}'"
+            )
+            rows = self._sqlite_cursor.fetchall()
+            return len(rows) > 0
+        else:
+            raise NotImplementedError
```
JoanFM, line 404: this does not seem a proper exception

JoanFM (changes requested):

docarray/index/backends/hnswlib.py
maxwelljin: This PR is still in progress, I'll change it later :) It should output proper hint to users

docarray/index/backends/in_memory.py
```diff
@@ -433,6 +433,9 @@ def _text_search_batched(
     ) -> _FindResultBatched:
         raise NotImplementedError(f'{type(self)} does not support text search.')
 
+    def __contains__(self, item: BaseDoc):
```
samsja, line 436:
```suggestion
    def _contains(self, item: BaseDoc):
```

the syntax `__smth__` means the function is a very special function, I think in this case the function should just be private and therefore be `_smth`

JoanFM: this contains operator allows user to do ' doc.id in index` in a very nice way

samsja: my bad

samsja (approved):
lgtm. Good job :tophat:

JoanFM (changes requested):

docarray/index/abstract.py
```diff
@@ -1166,3 +1166,46 @@ def _get_root_doc_id(self, id: str, root: str, sub: str) -> str:
...
+        """Checks if a given BaseDoc item is contained in the index or any of its subindices.
+
+        :param item: the given BaseDoc
+        :return: if the given BaseDoc item is contained in the index/subindices
+        """
+        if self.num_docs() == 0:
+            return False
+
+        if safe_issubclass(type(item), BaseDoc):
+            docs = self._get_all_documents()
```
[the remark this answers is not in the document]
maxwelljin: It's implemented in the line 1205 - 1207. I'll remove all `_get_all_documents` method, so it would only look for any sub-document inside the DocArray. (so the meaning for this subindex_contains method would similar to subindex_find)

JoanFM (approved):

PR #1641, opened 2023-06-12
fix: doc vec equality

### LINKED ISSUE #1638: Bug: Equality of DocVec
Equality between instances of DocVec does not seem to work properly:

```python


class CustomDocument(BaseDoc):
    image: str


da = DocList[CustomDocument](
    [CustomDocument(image='hi') for _ in range(10)]
).to_doc_vec()

da2 = DocList[CustomDocument](
    [CustomDocument(image='hi') for _ in range(10)]
).to_doc_vec()

print(da == da2)
```
```bash
False
```

npitsillos:
- Fixes #1638


JoanFM (changes requested):
Can u please add tests so that we can approve and merge?

JoanFM (approved):

PR #1642, opened 2023-06-12
fix: find_and_filter for inmemory

jupyterjazz:
fixes https://github.com/docarray/docarray/issues/1640

### DIFF
--- a/docarray/index/backends/helper.py
+++ b/docarray/index/backends/helper.py
@@ -21,13 +21,23 @@ def _collect_query_args(method_name: str):  # TODO: use partialmethod instead
 
 
 def _execute_find_and_filter_query(
-    doc_index: BaseDocIndex, query: List[Tuple[str, Dict]]
+    doc_index: BaseDocIndex, query: List[Tuple[str, Dict]], reverse_order: bool = False
 ) -> FindResult:
     """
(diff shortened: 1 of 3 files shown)


docarray/index/backends/helper.py
```diff
@@ -21,13 +21,23 @@ def inner(self, *args, **kwargs):
...
+
+    Args:
+        doc_index: Document index instance.
+            Either InMemoryExactNNIndex or HnswDocumentIndex.
+        query: Dictionary containing search and filtering configuration.
+        reverse_order: Flag indicating whether to sort in descending order. If set to
+            False (default), the sorting will be in ascending order.
+
+    Returns:
+        Sorted documents and their corresponding scores.
```
samsja, line 40: this is not the right style for docstring

samsja (changes requested):
Code looks good.

Tho I have to say it is unclear why adding this reverse sorting fix the original problem. Can you add a comment somewhere in the code to explain it ?

FYI: docstring are not in the right format

docarray/index/backends/helper.py
jupyterjazz: I got too used to langchain docstrings lol

samsja (approved):

PR #1645, opened 2023-06-13
feat: tensor type for protobuf deserialization

JohannesMessner:
This allows `DocVoc` to be deserilzed to a specific `tensor_type`, i.e. torch, tf, or numpy:

```python
# look at the unit test for more comprehensive example
class MyDoc(BaseDoc):
    tensor: TensorFlowTensor

da = DocVec[MyDoc](...)  # doesn't matter what tensor_type is here

proto = da.to_protobuf()
da_after = DocVec[MyDoc].from_protobuf(proto, tensor_type=TensorFlowTensor)

assert isinstance(da_after.tensor, TensorFlowTensor)
```
Note that the `tensor_type` passed to `from_protobuf()` does not need to match the tensor type from before serialization.
Since all tensor are represented the same way in the proto, any proto can be deserialized to any tensor type.

**TODO**:
- [x] docs

### DIFF
--- a/docarray/array/doc_vec/doc_vec.py
+++ b/docarray/array/doc_vec/doc_vec.py
@@ -589,8 +589,16 @@ class DocVec(AnyDocArray[T_doc]):
     ####################
 
     @classmethod
-    def from_protobuf(cls: Type[T], pb_msg: 'DocVecProto') -> T:
-        """create a DocVec from a protobuf message"""
+    def from_protobuf(
+        cls: Type[T], pb_msg: 'DocVecProto', tensor_type: Type[AbstractTensor] = NdArray
+    ) -> T:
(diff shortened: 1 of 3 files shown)


samsja (commented):
so what happened if I do `to_profobuf` with a TorchTensor in my doc_type ? It will be loaded as numpy tensor right ?

JohannesMessner:
> so what happened if I do `to_profobuf` with a TorchTensor in my doc_type ? It will be loaded as numpy tensor right ?

yes, if not `tensor_type` is specified in `from_protobuf` it will attempt to load it as numpy. This mirrors the behaviour of `DocVec.__init__()`.

samsja (approved):

PR #1648, opened 2023-06-14
fix: default column config should be DBConfig and not RuntimeConfig

JoanFM:


JohannesMessner:
Can we make this a non-breaking change for a few versions? And raise a warning?

JoanFM:
> Can we make this a non-breaking change for a few versions? And raise a warning?

What is breaking in your opinion? I do not see what can be breaking?

docarray/index/backends/hnswlib.py
```diff
@@ -188,6 +183,12 @@ class RuntimeConfig(BaseDocIndex.RuntimeConfig):
             }
         )
 
+    @dataclass
+    class RuntimeConfig(BaseDocIndex.RuntimeConfig):
+        """Dataclass that contains all "dynamic" configurations of HnswDocumentIndex."""
+
+        pass
```
JohannesMessner, line 190: I believe there are things that we could put here, but that is for a different PR I suppose

docs/user_guide/storing/index_hnswlib.md
```diff
@@ -58,36 +58,31 @@ To load existing data, you can specify a directory that stores data from a previ
...
-            'ef': 15,
-            'M': 8,
-            'allow_replace_deleted': True,
-            'num_threads': 5,
-        },
-        None: {},
-    }
-)
+
+db = HnswDocumentIndex[MyDoc](work_dir='/tmp/my_db', default_column_config={
```
JohannesMessner, line 72: can we `black` this?

JohannesMessner (approved):

PR #1649, opened 2023-06-14
refactor: implementation of InMemoryExactNNIndex follows DBConfig way

JoanFM:
The InMemoryExactNNIndexer gets an attribute in `__init__` instead of following the pattern of using `DBConfig`.

Waiting for #1648 to be merged updating Documentation

### DIFF
--- a/docarray/index/backends/in_memory.py
+++ b/docarray/index/backends/in_memory.py
@@ -1,5 +1,4 @@
 import os
-import warnings
 from collections import defaultdict
 from dataclasses import dataclass, field
 from typing import (
@@ -55,7 +54,6 @@ class InMemoryExactNNIndex(BaseDocIndex, Generic[TSchema]):
         """Initialize InMemoryExactNNIndex"""
         super().__init__(db_config=db_config, **kwargs)
         self._runtime_config = self.RuntimeConfig()
-        self._db_config = cast(InMemoryExactNNIndex.DBConfig, self._db_config)
         self._db_config = cast(InMemoryExactNNIndex.DBConfig, self._db_config)
         self._index_file_path = self._db_config.index_file_path
 
@@ -449,7 +447,7 @@ class InMemoryExactNNIndex(BaseDocIndex, Generic[TSchema]):
         DEFAULT_INDEX_FILE_PATH = 'in_memory_index.bin'
         file_to_save = self._index_file_path or file
         if file_to_save is None:
-            warnings.warn(
+            self._logger.warning(
                 f'persisting index to {DEFAULT_INDEX_FILE_PATH} because no `index_file_path` has been used inside DBConfig and no `file` has been passed as argument'
             )
         file_to_save = file_to_save or 'in_memory_index.bin'  # to keep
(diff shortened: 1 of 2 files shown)


docarray/index/backends/in_memory.py
```diff
@@ -442,9 +444,16 @@ def __contains__(self, item: BaseDoc):
                 f"item must be an instance of BaseDoc or its subclass, not '{type(item).__name__}'"
             )
 
-    def persist(self, file: str = 'in_memory_index.bin') -> None:
+    def persist(self, file: Optional[str] = None) -> None:
         """Persist InMemoryExactNNIndex into a binary file."""
-        self._docs.save_binary(file=file)
+        DEFAULT_INDEX_FILE_PATH = 'in_memory_index.bin'
```
jupyterjazz, line 449: I'd assign this inside init as a default value

docarray/index/backends/in_memory.py
```diff
@@ -442,9 +444,16 @@ def __contains__(self, item: BaseDoc):
...
+        file_to_save = self._index_file_path or file
+        if file_to_save is None:
+            warnings.warn(
```
jupyterjazz, line 452: let's do self._logger.warning here

jupyterjazz (commented):
I like this, makes more sense

docarray/index/backends/in_memory.py
JoanFM: I do not want, because I do not want it to have a default value at `__init__`.  This is just for backwards compatibility at `persist` time

jupyterjazz: ok but why wouldn't it be backward compatible?

JoanFM: I want it to be only at `persist level`. If I put at top level, it seems to apply at load time as well

docarray/index/backends/in_memory.py
```diff
@@ -442,9 +442,16 @@ def __contains__(self, item: BaseDoc):
...
+    def persist(self, file: Optional[str] = None) -> None:
         """Persist InMemoryExactNNIndex into a binary file."""
-        self._docs.save_binary(file=file)
+        DEFAULT_INDEX_FILE_PATH = 'in_memory_index.bin'
+        file_to_save = self._index_file_path or file
+        if file_to_save is None:
+            self._logger.warning(
+                f'persisting index to {DEFAULT_INDEX_FILE_PATH} because no `index_file_path` has been used inside DBConfig and no `file` has been passed as argument'
+            )
+        file_to_save = file_to_save or 'in_memory_index.bin'  # to keep
```
jupyterjazz, line 453:
```suggestion
        file_to_save = file_to_save or DEFAULT_INDEX_FILE_PATH
```

jupyterjazz (approved):
one small comment and lgtm

docs/user_guide/storing/index_in_memory.md
```diff
@@ -33,10 +33,16 @@ doc_index.index(docs)
...
+Alternatively, you can pass an `index_file_path` argument to make sure that the index can be restored if persisted from that specific file.
 ```python
 # Save your existing index as a binary file
-doc_index.persist('docs.bin')
+docs = DocList[MyDoc](MyDoc() for _ in range(10))
+
+doc_index = InMemoryExactNNIndex[MyDoc](index_file_path='docs.bin')
+doc_index.index(docs)
+
+# or in one step:
```
jupyterjazz, line 44: what do you mean by "# or in one step:" here?

JoanFM: Initialize it with the docs inserted

PR #1651, opened 2023-06-15
refactor: more robust method to detect duplicate index

### LINKED ISSUE #1357: Weaviate: Add more robust method to detect existing index
**Is your feature request related to a problem? Please describe.**
An index should be considered as exist when there is a schema with the exist name and properties.

**Describe the solution you'd like**
Use the client's `exists` instead of `contains` method to detect an existing index name.

**Describe alternatives you've considered**
NA

**Additional context**
Wait for weaviate/weaviate-python-client#232 to be implemented

hsm207:
fixes #1357


hsm207:
@JoanFM @JohannesMessner mypy and black is failing on files I did not modify and I already pulled from main

JoanFM (changes requested):
I guess u need to commit `pyproject` updated and not only `poetry.lock`?

hsm207:
> I guess u need to commit `pyproject` updated and not only `poetry.lock`?

the pyproject.toml file already has what I want when I pulled from main i.e.:

> weaviate-client = {version = ">=3.17, <3.18", optional = true}

I don't know if this is the expected behavior, but after the pull, I ran `poetry lock --no-update` and surprisingly there were changes to the poetry.lock file.

samsja:
Why is there a need to update the lock file ? Not sure that I follow

JoanFM (approved):

PR #1652, opened 2023-06-15
feat: hybrid pydantic support for both v1 and v2

samsja:
# Context

this PR allow support for pydantic v2 and v1 at the same time

Note: it is still very early progress


docarray/typing/abstract_type.py
```diff
@@ -1,12 +1,16 @@
...
+from typing import TYPE_CHECKING, Any, Type, TypeVar
 
 from pydantic import BaseConfig
 
 from docarray.utils._internal.pydantic import is_pydantic_v2
 
-if not is_pydantic_v2():
-    from pydantic.fields import ModelField
+if TYPE_CHECKING:
+    if not is_pydantic_v2():
```
JoanFM, line 9: why dont u do as jina did for docarray v2, have a variable instead of a method call each time?

samsja: yeah okay fair

docarray/base_doc/mixins/io.py
```diff
@@ -238,7 +242,7 @@ def from_protobuf(cls: Type[T], pb_msg: 'DocProto') -> T:
         for field_name in pb_msg.data:
             if (
                 not (cls.Config._load_extra_fields_from_protobuf)
-                and field_name not in cls.__fields__.keys()
+                and field_name not in cls._docarray_fields().keys()
```
JoanFM, line 245: why a dictionary is accessed as a method?

docarray/documents/audio.py
```diff
@@ -94,11 +94,11 @@ class MultiModalDoc(BaseDoc):
     ```
     """
 
-    url: Optional[AudioUrl]
-    tensor: Optional[AudioTensor]
-    embedding: Optional[AnyEmbedding]
-    bytes_: Optional[AudioBytes]
-    frame_rate: Optional[int]
+    url: Optional[AudioUrl] = None
```
JoanFM, line 97: why is this needed?

docarray/documents/image.py
```diff
@@ -92,10 +92,10 @@ class MultiModalDoc(BaseDoc):
     ```
     """
 
-    url: Optional[ImageUrl]
-    tensor: Optional[ImageTensor]
-    embedding: Optional[AnyEmbedding]
-    bytes_: Optional[ImageBytes]
+    url: Optional[ImageUrl] = None
```
JoanFM, line 95: also this?

docarray/documents/mesh/vertices_and_faces.py
```diff
@@ -23,7 +23,7 @@ class VerticesAndFaces(BaseDoc):
     faces: AnyTensor
 
     @classmethod
-    def validate(
+    def _docarray_validate(
```
JoanFM, line 26: why is this needed?

JoanFM (changes requested):

docarray/array/doc_vec/io.py
```diff
@@ -268,28 +262,26 @@ def to_protobuf(self) -> 'DocVecProto':
             NdArrayProto,
         )
 
-        self_ = cast('DocVec', self)
```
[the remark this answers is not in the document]
samsja: sure

docarray/base_doc/mixins/io.py
samsja: it is a method that return a dictionaries. I used initially a property but classmethd + property was failingwith python 3.8 (working with 3.9)

docarray/documents/audio.py
samsja: None is need each  time there is Optional.

This does not affect the behavior of pydantic v1 and in pydantic v2 you have to add none to have the same behavior as pydantic v1.


FYI: https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields

docarray/documents/image.py
samsja: same as above, you will see this at a lot of place in the codebase

docarray/documents/mesh/vertices_and_faces.py
samsja: this is an alias, the signture function of the validate function change between pydtnic v1 and v2, therefore I created this function to abstract away the validation logic in one place and then depending on the pydantic version install I translate it to the right function singature

docarray/utils/_internal/pydantic.py
```diff
@@ -0,0 +1,12 @@
+import pydantic
+
+is_pydantic_v2 = pydantic.__version__.startswith('2.')
+
+
+if not is_pydantic_v2:
+    from pydantic.validators import bytes_validator
+
+else:
+    from pydantic.v1.validators import bytes_validator
```
JohannesMessner, line 10: Am I understanding correctly that we are falling back to the bytes validator of v1 even if pydantic v2 is used? What is the reason behind this?

docarray/utils/create_dynamic_doc_class.py
```diff
@@ -49,6 +50,11 @@ class MyDoc(BaseDoc):
     :param model: The input model
     :return: A new subclass of BaseDoc, where every DocList type in the schema is replaced by List.
     """
+    if is_pydantic_v2:
+        raise NotImplementedError(
+            'This method is not supported in Pydantic 2.0. Please use Pydantic 1.8.2 or lower.'
+        )
+
```
JohannesMessner, line 57: Why is this?

docarray/base_doc/any_doc.py
```diff
@@ -28,7 +30,14 @@ def _get_field_type(cls, field: str) -> Type['BaseDoc']:
...
+    def _get_field_annotation_array(cls, field: str) -> Type:
         from docarray import DocList
 
         return DocList
+
+    if is_pydantic_v2:
+
+        def dict(self, *args, **kwargs):
+            raise NotImplementedError(
+                "dict() method is not implemented for pydantic v2. Now pydantic require the schema to dump the dict but AnyDoc is schemaless"
```
JohannesMessner, line 42:
```suggestion
                "dict() method is not implemented for pydantic v2. Now pydantic requires a schema to dump the dict, but AnyDoc is schemaless"
```

docarray/base_doc/doc.py
```diff
@@ -36,6 +45,13 @@
 
     from docarray.array.doc_vec.column_storage import ColumnStorageView
 
+if is_pydantic_v2:
+
+    IncEx: typing_extensions.TypeAlias = (
```
JohannesMessner, line 50: What does `IncEx` stand for?

docarray/base_doc/doc.py
```diff
@@ -78,35 +94,144 @@ class MyDoc(BaseDoc):
...
-    class Config:
-        json_loads = orjson.loads
-        json_dumps = orjson_dumps_and_decode
-        # `DocArrayResponse` is able to handle tensors by itself.
-        # Therefore, we stop FastAPI from doing any transformations
-        # on tensors by setting an identity function as a custom encoder.
-        json_encoders = {AbstractTensor: lambda x: x}
+    if is_pydantic_v2:
+
+        class Config:
```
JohannesMessner, line 99: Should we still have this Config inner class, it is not recommended anymore in v2, right?
Also, where did the orjson stuff and the extra protobuf setting go?

docarray/base_doc/doc.py
```diff
@@ -78,35 +94,144 @@ class MyDoc(BaseDoc):
...
+            json_dumps = orjson_dumps_and_decode
+            # `DocArrayResponse` is able to handle tensors by itself.
+            # Therefore, we stop FastAPI from doing any transformations
+            # on tensors by setting an identity function as a custom encoder.
+            json_encoders = {AbstractTensor: lambda x: x}
+
+            validate_assignment = True
+            _load_extra_fields_from_protobuf = False
+
+    if is_pydantic_v2:
```
JohannesMessner, line 117: Could you put a high level comment here explaining why v2 needs different treatment from v1?

docarray/base_doc/doc.py
```diff
@@ -207,73 +332,193 @@ def _docarray_to_json_compatible(self) -> Dict:
         """
         return self.dict()
 
-    ########################################################################################################################################################
-    ### this section is just for documentation purposes will be removed later once
-    # https://github.com/mkdocstrings/griffe/issues/138 is fixed ##############
-    ########################################################################################################################################################
+    def _exclude_doclist(
```
JohannesMessner, line 335: What does this do? Docstring please :)

docarray/base_doc/doc.py
```diff
@@ -207,73 +332,193 @@ def _docarray_to_json_compatible(self) -> Dict:
...
+                    val = getattr(self, field)
+                    data[field] = (
+                        [doc.dict() for doc in val] if val is not None else None
+                    )
+
+            return data
+
+    else:
+
+        def model_dump(  # type: ignore
```
JohannesMessner, line 470: Am I understanding correctly that in v2 json and dict are both subsumed by this model_dump? So if the user wants json they would to `model_dump()` first and then convert that to json themselves?

JohannesMessner (commented):
Okay first pass done, I have few questions and will probably make a more thorough pass tomorrow!

docarray/utils/create_dynamic_doc_class.py
samsja: the method we used has been removed from pydantic v2, will have to find a fix later

docarray/base_doc/doc.py
samsja: this is the named used in pydantic function signature. Not even sure what this means

docarray/base_doc/doc.py
samsja: good question,

basically now in pydantic v2 this way is "deprecated". But at the same time, it is still working. Therefore if we keep it as a class it means we don't have to repeat the config for pydantic v2. Might be easier to maintain ...

samsja: > Also, where did the orjson stuff and the extra protobuf setting go?

This is handle differently with pydantic v2 and therefore not needed

docarray/base_doc/doc.py
samsja: model_dump is used under the hood to both seriliaze to json and to "dict". Basically now in pydantic v2 calling .dict() and .json() will call under the hood this model_dump

docarray/utils/_internal/pydantic.py
JohannesMessner: ?

docarray/base_doc/doc.py
JohannesMessner: okay

docarray/base_doc/doc.py
JohannesMessner: hmm, using deprecated stuff is not great. Does it raise a warning?

docarray/base_doc/doc.py
JohannesMessner: ^^

JohannesMessner: sorry hehe

JohannesMessner (approved):
the only thing that feels icky to me is using the deprecated `Config` inner class

docarray/base_doc/doc.py
samsja: it does indeed raise a warning ...

samsja:
I will go with @JohannesMessner proposal and switch to ConfigDict instead of class

PR #1654, opened 2023-06-15
fix: update text search to match client's new sig

### LINKED ISSUE #1653: Weaviate: update text search to match latest client's requirements
Starting from 3.15.6, with_bm25 no longer expects a dictionary. 

See weaviate/weaviate-python-client#290 for details.

hsm207:
fix #1653


pyproject.toml
```diff
@@ -52,7 +52,7 @@ hnswlib = {version = ">=0.6.2", optional = true }
 lz4 = {version= ">=1.0.0", optional = true}
 pydub = {version = "^0.25.1", optional = true }
 pandas = {version = ">=1.1.0", optional = true }
-weaviate-client = {version = ">=3.15", optional = true}
+weaviate-client = {version = ">=3.17, <3.18", optional = true}
```
samsja, line 55:
```suggestion
weaviate-client = {version = ">=3.17", optional = true}
```

I think it is better to not cap it. It might be that 3.18 will not break anything for docarray

hsm207: @samsja we should cap it now as there are breaking changes in the weavaite implementation if we don't. I will make another PR to fix it and will remove the cap there.

samsja: but this means that the next version of docarray will never be compatible with the next one of weaviate. We cannot know in advance that it will be the case. It is better to react after wards

hsm207: If I remove the cap now, then will poetry download the latest client i.e. 3.20.1 into the CI? If it does, then some of the tests will fail.

JoanFM: and when will this be fixed?

hsm207: the week after next week

samsja: then lets merge like this for now. Sorry misunderstanding on my side I though this PR was fixing docarray for version 3.17 and above

PR #1655, opened 2023-06-16
fix: bugs when serialize union type

maxwelljin:
This bug is caused by the deserialization process. The deserialization function currently only checks the schema of the class, without specific information about the particular document type in the field of union type. To resolve this issue, we may need to modify the data storage. After discussing with Joan, we could temporarily remove support for the union type and add this info the document. Additionally, we provide users with a more informative error message.

### DIFF
--- a/docs/user_guide/sending/serialization.md
+++ b/docs/user_guide/sending/serialization.md
@@ -304,5 +304,5 @@ assert isinstance(dv_from_proto_numpy.tensor, NdArray)
 ```
 
 !!! note
-    Currently, union type hints are not supported for serialization to protobuf formats.
+    Serialization to protobuf formats is not supported for union type hints. Please consider using JSON serialization as an alternative.
 
--- a/docs/user_guide/sending/serialization.md
+++ b/docs/user_guide/sending/serialization.md
@@ -304,5 +304,7 @@ assert isinstance(dv_from_proto_numpy.tensor, NdArray)
 ```
 
 !!! note
-    Serialization to protobuf formats is not supported for union type hints. Please consider using JSON serialization as an alternative.
+    Serialization to protobuf formats is not supported for union type hints (unless your union type consists solely of basic types). Please use JSON serialization as an alternative for these cases.
+
+
 
--- a/docs/user_guide/sending/serialization.md
+++ b/docs/user_guide/sending/serialization.md
@@ -303,4 +303,6 @@ assert dv_from_proto_numpy.tensor_type == NdArray
 assert isinstance(dv_from_proto_numpy.tensor, NdArray)
 ```
 
+!!! note
+    Currently, union type hints are not supported for serialization to protobuf formats.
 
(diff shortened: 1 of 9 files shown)


docs/user_guide/sending/serialization.md
```diff
@@ -303,4 +303,6 @@ assert dv_from_proto_numpy.tensor_type == NdArray
 assert isinstance(dv_from_proto_numpy.tensor, NdArray)
 ```
 
+!!! note
+    Currently, union type hints are not supported for serialization to protobuf formats.
```
JoanFM, line 307: Any union type? or Union types of regular types are? Can u add tests for the exceptions?

JoanFM (changes requested):

docs/user_guide/sending/serialization.md
maxwelljin: I'll add more tests. It should be any union type as union type doesn't have the `from_protobuf` method

maxwelljin:
Currently, the serialization for union type is only supported through `from_json` and `to_json`. This new commit adds tests to this bug.

docs/user_guide/sending/serialization.md
```diff
@@ -304,5 +304,5 @@ assert isinstance(dv_from_proto_numpy.tensor, NdArray)
 ```
 
 !!! note
-    Currently, union type hints are not supported for serialization to protobuf formats.
+    Serialization to protobuf formats is not supported for union type hints. Please consider using JSON serialization as an alternative.
```
JoanFM, line 307: Avoid the consideration, just inform

tests/units/array/test_array_from_to_bytes.py
```diff
@@ -69,3 +69,17 @@ def test_from_to_base64(protocol, compress, show_progress):
...
     assert da[1].image.url is None
     assert da2[1].image.url is None
+
+
+def test_union_type_error(tmp_path):
+    from typing import Union
+
+    from docarray.documents import TextDoc
+
+    class CustomDoc(BaseDoc):
```
JoanFM, line 79: what about Union[int, float] or Union[int, str]

JoanFM (changes requested):
Still not answered what happens with `Union` of basic types

tests/units/array/test_array_from_to_bytes.py
maxwelljin: After the testing, the Union with basis types should work in all serialization methods. Only the Union with document types will cause those issues.

docs/user_guide/sending/serialization.md
```diff
@@ -303,4 +303,8 @@ assert dv_from_proto_numpy.tensor_type == NdArray
 assert isinstance(dv_from_proto_numpy.tensor, NdArray)
 ```
 
+!!! note
+    Serialization to protobuf formats is not supported for union type hints (unless your union type consists solely of basic types). Please use JSON serialization as an alternative for these cases.
```
JoanFM, line 307:
```suggestion
    Serialization to protobuf is not supported for union types involving `BaseDoc` types.
```

JoanFM (changes requested):

PR #1663, opened 2023-06-20
fix: docvec equality if tensors are involved

JohannesMessner:
Before this, DocVec equality checks that involved a tensor would fail by raising and Exception:

```python
    class Text(BaseDoc):
        tens: TorchTensor

    da = DocVec[Text](
        [Text(tens=[1, 2, 3, 4]) for _ in range(10)], tensor_type=TorchTensor
    )
    da2 = DocVec[Text](
        [Text(tens=[1, 2, 3, 4]) for _ in range(10)], tensor_type=TorchTensor
    )
    assert da == da2  # would raise
```
```terminal
RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

This is because checks of the form `tens1 == tens2` return a boolean tensor, not a single bool, which breaks an assumption in the previous implementation.

This PR solved that by introducing a helper function that let's us compare tensors and get a single bool as output.

### DIFF
--- a/docarray/helper.py
+++ b/docarray/helper.py
@@ -301,7 +301,7 @@ def _tensor_equals(tens1: Any, tens2: Any) -> bool:
 
         t1, t2 = getattr(tens1, 'tensor', None), getattr(tens2, 'tensor', None)
         if tf.is_tensor(t1) and tf.is_tensor(t2):
-            return t1.shape == t2.shape and tf.math.reduce_all(tf.equal(t1, t1))
+            return t1.shape == t2.shape and tf.math.reduce_all(tf.equal(t1, t1))  # type: ignore
 
     are_np_arrays = isinstance(tens1, np.ndarray) and isinstance(tens2, np.ndarray)
     if are_np_arrays:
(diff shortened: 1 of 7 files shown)


docarray/helper.py
```diff
@@ -256,3 +259,52 @@ def _shallow_copy_doc(doc):
...
+    if is_tf_available():
+        import tensorflow as tf  # type: ignore
+
+        t1, t2 = getattr(tens1, 'tensor', None), getattr(tens2, 'tensor', None)
+        if tf.is_tensor(t1) and tf.is_tensor(t2):
+            return t1.shape == t2.shape and tf.math.reduce_all(tf.equal(t1, t1))  # type: ignore
+
+    are_np_arrays = isinstance(tens1, np.ndarray) and isinstance(tens2, np.ndarray)
+    if are_np_arrays:
+        return np.array_equal(tens1, tens2)
```
samsja, line 308: should not this be part of our computational backend ?

JohannesMessner: good point, I will refactor

samsja: thx

samsja (approved):

PR #1666, opened 2023-06-21
feat: support milvus

maxwelljin:
This PR aims to support Milvus as a storage backend

- [x] Index creation
- [x] Get documents
- [x] Search
- [x] Filter
- [ ] Subindex
- [ ] Tests


docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,328 @@
...
+
+        for py_type, db_type in type_map.items():
+            if safe_issubclass(python_type, py_type):
+                return db_type
+
+        raise ValueError("No corresponding milvus type")
+
+    def _init_index(self) -> Collection:
+        if not utility.has_collection(self._db_config.collection_name):
+            print(self._db_config.collection_name)
```
JohannesMessner, line 122: Don't forget to remove the print

docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,328 @@
...
+                    is_primary=True,
+                    auto_id=True,
+                )
+            ]  # Initiliaze the id
+            fields.extend(
+                [
+                    FieldSchema(
+                        name="doc_id" if column_name == "id" else column_name,
+                        dtype=DataType.VARCHAR if column_name == "id" else info.db_type,
+                        is_primary=False,
```
JohannesMessner, line 136: shouldn't `id``be completely skipped here since it is alrady added to the schema above? Won't this add the id twice, once as `id` and once as `doc_id`?

docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,328 @@
...
+                    )
+                    for column_name, info in self._column_infos.items()
+                ]
+            )
+            self._logger.info("Collection has been created")
+            return Collection(
+                name=self._db_config.collection_name,
+                schema=CollectionSchema(
+                    fields=fields,
+                    description="Collection Schema",
```
JohannesMessner, line 156: Perhaps the description could also be passes as a config?

docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,328 @@
...
+            "metric_type": self._db_config.index_metric,
+            "params": self._db_config.index_params,
+        }
+
+        self._collection.create_index(self._db_config.index_name, index)
+        self._logger.info(
+            f"Index '{self._db_config.index_name}' has been successfully created"
+        )
+
+    def index(self, docs: Union[BaseDoc, Sequence[BaseDoc]], **kwargs):
```
JohannesMessner, line 208: What is the reason for implementing `index()` instead of implementing `_index()` and utilizing `index()` from the base class?

docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,328 @@
...
+        self._collection.insert(entities)
+        self._collection.flush()
+        self._logger.info(f"{len(docs)} documents has been indexed")
+
+    def _convert_to_vector(self, column_value: AbstractTensor) -> Sequence[float]:
+        """
+        Converts a column value to a float vector.
+
+        The database can only store float vectors, so this method is used to convert
+        TensorFlow or PyTorch tensors to a format compatible with the database.
```
JohannesMessner, line 240: The base class should already take care of torch, tf etc conversions. Are you sure you need to do this again?

docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,328 @@
...
+        if len(column_value.shape) != 1:
+            raise ValueError(
+                'Unsupported: Milvus backend only supports one-dimensional vectors'
+            )
+
+        if isinstance(column_value, np.ndarray):
+            return column_value.astype(float).tolist()
+        elif torch.is_tensor(column_value):
+            return column_value.float().numpy().tolist()
+        elif tf.is_tensor(column_value):
```
JohannesMessner, line 254: This is not safe, because not every user has torch and/or tf installed. Therefore we can't directly use torch/tf. [Here](https://github.com/docarray/docarray/blob/3fc6ecb71bdc0095f2c405c17492debcc3d8412d/docarray/typing/tensor/video/video_tensor.py#L19) you can see how that is handled in other places.

JohannesMessner (changes requested):
I think we are on a good path here! Just a few comments

docarray/index/backends/milvus.py
maxwelljin: Thanks for the feedback! I'll only keep doc_id

docarray/index/backends/milvus.py
maxwelljin: In my new implementation, the `index()` function would also do the serialization (there's a column for serialized data), so the index function need to be utilized

docarray/index/backends/milvus.py
maxwelljin: I'm not very sure how the base class handles conversions between different formats. The vector embedding is still stored in its raw format if we use the `__getattr__` method.

docarray/index/__init__.py
```diff
@@ -34,6 +34,9 @@ def __getattr__(name: str):
     elif name == 'WeaviateDocumentIndex':
         import_library('weaviate', raise_error=True)
         import docarray.index.backends.weaviate as lib
+    elif name == "MilvusDocumentIndex":
```
JoanFM, line 37: single quotes please

docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,422 @@
...
+    from pymilvus import (
+        Collection,
+        CollectionSchema,
+        DataType,
+        FieldSchema,
+        connections,
+        utility,
+    )
+else:
+    hnswlib = import_library('hnswlib', raise_error=True)
```
JoanFM, line 53: why do we need this library?

docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,422 @@
...
+        )
+
+        self._loaded = False
+
+        self._db_config.index_name = index_name
+        self._validate_columns()
+        self._create_collection_name()
+        self._collection = self._init_index()
+        self._build_index()
+        self._logger.info(f"{self.__class__.__name__} has been initialized")
```
JoanFM, line 92: use single quotes

docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,422 @@
...
+from docarray.utils._internal.misc import (
+    import_library,
+    is_tf_available,
+    is_torch_available,
+)
+from docarray.utils.find import _FindResult, _FindResultBatched
+
+torch_available, tf_available = is_torch_available(), is_tf_available()
+
+if torch_available:
```
JoanFM, line 36: why do we need to import these?

JoanFM (changes requested):

docarray/index/backends/milvus.py
maxwelljin: That's because when we perform a vector search, it requires the input in Python list format. If the user's input tensor is in torch/tensorflow format, we need to perform a conversion.

docarray/index/backends/milvus.py
maxwelljin: I just found base class' implementation. I'll remove my implementation here. Thanks for your suggestion!

jupyterjazz:
Great work @maxwelljin, I'll take over from here!

PR #1667, opened 2023-06-21
feat: add method to create BaseDoc from schema

JoanFM:
This code is used inside `Jina` to dynamically create `BaseDocs` in services where the actual BaseDoc is not created but still hold the validation capabilities.

You can think as being easy to validate the `BaseDoc` at the driver or client program of a server serving BaseDoc models.


docarray/utils/create.py
```diff
@@ -0,0 +1,220 @@
...
+from typing import Dict, List, Any, Union, Optional, Type
+
+
+def create_new_model_cast_doclist_to_list(model: Any) -> BaseDoc:
+    """
+    Take a Pydantic model and cast DocList fields into List fields.
+
+    This can be needed because of this limitation of Pydantic:
+
+    (https://docs.pydantic.dev/latest/blog/pydantic-v2/)
```
JohannesMessner, line 13: Can we have a more specific link here? This just points to the entire v2 plan

docarray/utils/create.py
```diff
@@ -0,0 +1,220 @@
...
+
+    ---
+
+    ```python
+    from docarray import BaseDoc
```
JohannesMessner, line 18: (nitpick) I think this is not blacked, there should be an empty line between the import and the code

docarray/utils/create.py
```diff
@@ -0,0 +1,220 @@
...
+        url: ImageUrl
+        title: str
+        texts: DocList[TextDoc]
+
+    MyDocCorrected = create_new_model_cast_doclist_to_list(CustomDoc)
+    ```
+
+    ---
+    :param model: The input model
+    :return: A BaseDoc class dynamically created with List instead of DocList in the schema.
```
JohannesMessner, line 30:
```suggestion
    :return: A new subclass of BaseDoc, where every DocList type in the schema is replaced by List.
```

docarray/utils/create.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def _get_field_from_type(
+    field_schema,
+    field_name,
+    root_schema,
+    cached_models,
+    is_tensor=False,
+    num_recursions=0,
+):
```
JohannesMessner, line 54: Can we have type hints to that mypy can do its job?

A small docstring just for internal purposes would also be nice, the code is quite complicated.

docarray/utils/create.py
```diff
@@ -0,0 +1,220 @@
...
+) -> Type:
+    """
+    Dynamically create a `BaseDoc` class from a `schema` of another `BaseDoc`.
+    This method is intended to dynamically create a `BaseDoc` compatible with the schema
+    of another BaseDoc that is not available in the context. For instance, you may have stored the schema
+    as a JSON, or sent it to another service, etc ...
+
+    Due to this Pydantic limitation (https://docs.pydantic.dev/latest/blog/pydantic-v2/), we need to make sure that the
+    input schema uses `List` and not `DocList`, therefore this is recommended to be used in combination with `create_new_model_cast_doclist_to_list`
+    to make sure that `DocLists` in schema are converted to `List`.
```
JohannesMessner, line 179:
```suggestion
    """
    Dynamically create a `BaseDoc` subclass from a `schema` of another `BaseDoc`.
    
    This method is intended to dynamically create a `BaseDoc` compatible with the schema
    of another BaseDoc. This is useful when that other `BaseDoc` is not available in the current scope. For instance, you may have stored the schema
    as a JSON, or sent it to another service, etc.

    Due to this Pydantic limitation  ### better link here, like above ### (https://docs.pydantic.dev/latest/blog/pydantic-v2/), we need to make sure that the
    input schema uses `List` and not `DocList`. Therefore this is recommended to be used in combination with `create_new_model_cast_doclist_to_list`
    to make sure that `DocLists` in schema are converted to `List`.
```

docarray/utils/create.py
```diff
@@ -0,0 +1,220 @@
...
+            )
+        else:
+            raise ValueError(
+                f"Unknown field type: {field_type} for field_name {field_name}"
+            )
+    return ret
+
+
+def create_base_doc_from_schema(
+    schema: Dict[str, Any], model_name: str, cached_models: Optional[Dict] = None
```
JohannesMessner, line 169:
```suggestion
    schema: Dict[str, Any], doc_name: str, cached_models: Optional[Dict] = None
```

docarray/utils/create.py
```diff
@@ -0,0 +1,220 @@
...
+
+    MyDocCorrected = create_new_model_cast_doclist_to_list(CustomDoc)
+    new_my_doc_cls = create_base_doc_from_schema(
+        CustomDocCopy.schema(), 'MyDoc'
+    )
+    ```
+
+    ---
+    :param schema: The schema of the original `BaseDoc` where DocLists are passed as regular Lists of Documents.
+    :param model_name: The name of the new pydantic model created.
```
JohannesMessner, line 199:
```suggestion
    :param doc_name: The name of the newly created BaseDoc subclass.
```

tests/units/util/test_create.py
```diff
@@ -0,0 +1,240 @@
+import pytest
+from typing import List, Dict, Union, Any
+from docarray.utils.create import (
```
JohannesMessner, line 3:
```suggestion
from docarray.utils.create import (
```
Maybe we should have these directly importable from `docarray.utils`?

docarray/utils/create.py
```diff
@@ -0,0 +1,220 @@
+from docarray import DocList, BaseDoc
```
JohannesMessner, line 1: I would call the file `create_doc_class.py` or something like that (and make the functions directly importable from `docarray.utils`)

JohannesMessner (changes requested):

tests/units/util/test_create.py
JoanFM: we can do this in another PR for the rest of the cases

docarray/utils/create.py
```diff
@@ -0,0 +1,235 @@
...
+from docarray.typing import AnyTensor
+from pydantic import create_model
+from typing import Dict, List, Any, Union, Optional, Type
+
+
+def create_new_model_cast_doclist_to_list(model: Any) -> BaseDoc:
+    """
+    Take a Pydantic model and cast DocList fields into List fields.
+
+    This can be needed because of this limitation of Pydantic:
```
scott-martens, line 11:
```suggestion
    This may be necessary due to limitations in Pydantic:
```

docarray/utils/create.py
JoanFM: I will change the file name, to make them importable, I would handle in another PR where all `utils` are exposed

docarray/utils/create.py
JoanFM: Done!

docarray/utils/create_dynamic_doc_class.py
```diff
@@ -0,0 +1,236 @@
+from docarray import DocList, BaseDoc
+from docarray.typing import AnyTensor
+from pydantic import create_model
+from typing import Dict, List, Any, Union, Optional, Type
+
+
+def create_new_model_cast_doclist_to_list(model: Any) -> BaseDoc:
```
samsja, line 7: the name is a bit confusing

samsja: but does it need to be a public method ? If not they lets just make it private and no need to change the name I guess

JoanFM: it needs to be yes

samsja: Okay I see, well maybe the name is good them,  I cannot find a better one

samsja: or maybe `create_pure_python_type_model` ? Who knows maybe when we will introduce DocDIct we will have to convert DOcDict to Dict as well in this method and therefore the name with DocList is not relevant anymore

JoanFM: okey

JoanFM: done

samsja (approved):

PR #1669, opened 2023-06-26
feat: validate file formats in url (#1606)

jupyterjazz:
Followup of https://github.com/docarray/docarray/pull/1606

Approach:
1. Validate given url based on what type `mimetypes` will guess
2. If the first step is not successful, try validating against extra extensions provided for each url type

Why was CI failing in Kalim's PR:
Apparently `mimetypes` additionally uses system's `mime.types` file which is unique for different operating systems, even for different versions of the same operating system. Because of this file, `mimetypes` was guessing different types locally and on CI, resulting in strange errors. I disabled it by `mimetypes.init([])` which means `mimetypes` will ignore system's `mime.types` and return same types every time


docarray/typing/url/text_url.py
```diff
@@ -13,6 +13,32 @@ class TextUrl(AnyUrl):
...
+    @classmethod
+    def mime_type(cls) -> str:
+        return 'text'
+
+    @classmethod
+    def extra_extensions(cls) -> List[str]:
+        """
+        List of extra file extensions for this type of URL (outside the scope of mimetype library).
+        """
+        return ['.md']
```
JoanFM, line 25: avoid magic numbers, have all these `strings` in some place with meaningful constant names

JoanFM (changes requested):

docarray/typing/url/any_url.py
```diff
@@ -38,6 +52,29 @@ def _to_node_protobuf(self) -> 'NodeProto':
...
+        """
+        Check if the file extension of the URL is allowed for this class.
+        First, it guesses the mime type of the file. If it fails to detect the
+        mime type, it then checks the extra file extension.
+
+        :param value: The URL or file path.
+        :return: True if the extension is allowed, False otherwise
+        """
+        if cls is AnyUrl:
+            return True
```
JoanFM, line 66: I do not understand this code, will this not always return True? what ks trying to do here?

docarray/typing/url/audio_url.py
```diff
@@ -17,6 +17,18 @@ class AudioUrl(AnyUrl):
     Can be remote (web) URL, or a local file path.
     """
 
+    @classmethod
+    def mime_type(cls) -> str:
+        return 'audio'
```
JoanFM, line 22: avoid using magic names, us constants with meaningful names

docarray/typing/url/text_url.py
```diff
@@ -13,6 +14,18 @@ class TextUrl(AnyUrl):
     Can be remote (web) URL, or a local file path.
     """
 
+    @classmethod
+    def mime_type(cls) -> str:
+        return 'text'
```
JoanFM, line 19: pleas use meaningful constants

docarray/typing/url/url_3d/url_3d.py
```diff
@@ -18,6 +18,10 @@ class Url3D(AnyUrl, ABC):
     Can be remote (web) URL, or a local file path.
     """
 
+    @classmethod
+    def mime_type(cls) -> str:
+        return 'application'
```
JoanFM, line 23: what is this mime type? use constants alsl

tests/integrations/predefined_document/test_audio.py
```diff
@@ -29,7 +29,6 @@
     str(TOYDATA_DIR / 'hello.ogg'),
     str(TOYDATA_DIR / 'hello.wma'),
     str(TOYDATA_DIR / 'hello.aac'),
-    str(TOYDATA_DIR / 'hello'),
```
JoanFM, line 32: why removed?

tests/integrations/predefined_document/test_audio.py
```diff
@@ -40,7 +39,6 @@
     (str(TOYDATA_DIR / 'hello.ogg'), 'ogg'),
     (str(TOYDATA_DIR / 'hello.wma'), 'asf'),
     (str(TOYDATA_DIR / 'hello.aac'), 'adts'),
-    (str(TOYDATA_DIR / 'hello'), 'wav'),
```
JoanFM, line 43: why removed?

tests/units/typing/url/test_text_url.py
```diff
@@ -8,7 +8,7 @@
 from docarray.typing import TextUrl
 from tests import TOYDATA_DIR
 
-REMOTE_TEXT_FILE = 'https://de.wikipedia.org/wiki/Brixen'
+REMOTE_TEXT_FILE = 'https://www.gutenberg.org/files/1065/1065-0.txt'
```
JoanFM, line 11: is this test needed? I do not like to have a test depemding on an external website that can remove the image

JoanFM (changes requested):

docarray/typing/url/any_url.py
jupyterjazz: No this is a direct comparison (same as ==)
issubclass would return True always

tests/units/typing/url/test_text_url.py
jupyterjazz: but we were already doing that with wikipedia

jupyterjazz: the purpose is to test loading remote text files, I think gutenberg project is very reliable, if not we can easily remove it

tests/integrations/predefined_document/test_audio.py
jupyterjazz: Because it's an Audio URL without an audio extension and should not be validated

docarray/typing/url/url_3d/url_3d.py
jupyterjazz: This is a broad category for mimetypes. Contains obj, pdf, json, xml and many other files.. Now that I think about it, we should make it more specific (whatever is associated with .obj extension because that's what we usually use) and for other non-obj files rely on extra extensions. Changed accordingly.

tests/integrations/predefined_document/test_audio.py
jupyterjazz: Answered above

tests/integrations/predefined_document/test_audio.py
JohannesMessner: I am not sure about that, (at least on unix) i can store an audio file without extension I believe, why should that not be allowed? Admittedly, for audio it may not be common to do that, but text files do it, e.g. `Dockerfile`

jupyterjazz: Yes you can, but in order to avoid issues like https://github.com/docarray/docarray/issues/1555 we need to look at extensions. 

But you have a good point, text files without extensions are very common. Does it make sense to ignore validating text URLs that have no extensions? I don't really have another solution, we can't guess extensions or types in that scenario, and trying to read them during validation will be slow

JoanFM: let's ignore validation of TextURL then?

JohannesMessner: Can't we have a rule that is like "_if_ there is an extension, validate it; if there is no extension, don't validate"? We could have that for all url types, no?

jupyterjazz: > "if there is an extension, validate it; if there is no extension, pass validation"

yeap this is what I meant, but for text urls only. 
ok let's do it for all urls

docarray/typing/url/any_url.py
```diff
@@ -38,6 +52,29 @@ def _to_node_protobuf(self) -> 'NodeProto':
...
+
+        url_parts = value.split("?")
+        mimetype, _ = mimetypes.guess_type(url_parts[0])
+        if mimetype and mimetype.startswith(cls.mime_type()):
+            return True
+
+        filename = url_parts[0].split('.')
+        extension = filename[-1] if len(filename) > 1 else None
```
JohannesMessner, line 74: maybe I am being overly cautious here, but do we know for a fact that there are no corner cases where this splitting into filename and extension could break? Is there some resource or standard that we can reference?
Alternatively, I think pydantic implements some of this internally, Maybe we could repurpose some of their logic?

jupyterjazz: but I think this is ok for now, wdyt?

PR #1672, opened 2023-06-26
fix: allow additional_headers parameters to Weaviate client

girishc13:
Weaviate v3.9.0 supports `additional_headers` config param to pass in authentication headers.


girishc13:
Let me know how to handle unit tests.

PR #1675, opened 2023-06-27
fix: better error message when docvec is unusable

### LINKED ISSUE #1563: bug: better error message after calling `to_doc_list` on DocVec
# Context

When calling `to_doc_list` on a DocVec it destroys (delete) the storage of DocVec and therefore the DocVec is not usable.

Nevertheless we can still have a reference to the DocVec object, if we try to call any method on it it will fail with a weird error message. While it is expected that it fail the error message should clearly explain what is the problem: the DocVec should not be used anymore because its content has been moved to a DocList.

## Example to reproduce the problem

```python
from docarray import DocVec
from docarray.documents import ImageDoc


docs = DocVec[ImageDoc](
    [ImageDoc(url='http://url.com/foo.png') for _ in range(10)]
)

docs.to_doc_list()

print(docs)
```

```bash
  File "/home/sami/.config/JetBrains/PyCharmCE2023.1/scratches/scratch_2.py", line 11, in <module>
    print(docs)
  File "/home/sami/Documents/workspace/Jina/docarray2/docarray/docarray/array/any_collections.py", line 38, in __repr__
    return f'<{self.__class__.__name__} (length={len(self)})>'
  File "/home/sami/Documents/workspace/Jina/docarray2/docarray/docarray/array/doc_vec/doc_vec.py", line 507, in __len__
    return len(self._storage)
  File "/home/sami/Documents/workspace/Jina/docarray2/docarray/docarray/array/any_array.py", line 51, in __getattr__
    return super().__getattribute__(item)
AttributeError: 'DocVec[ImageDoc]' object has no attribute '_storage'
```

## Expected results

We should rather yield a more explicit error message

```bash
RuntimeError: 'DocVec[ImageDoc] has been destroy and cannot be used anymore. Probably because you call `to_doc_list()` on it'
```

JohannesMessner:
closes #1563 

The approach is to set a flag `_is_unasable` as soon as the instance gets unusable, and from that point onward raise an Exception at every getatts/setattr interaction.

Unfortunately we cannot override getattr/setattr on the fly (and get rid of the `_is_unasable`) flag: https://stackoverflow.com/questions/10376604/overriding-special-methods-on-an-instance


JoanFM (changes requested):
Why is it unusable?

JohannesMessner:
> Why is it unusable?

Because `to_doc_list()` directly modifies the data stored in the columns in order to not copy the data. This "destroys" the columns.
@samsja knows more about the implementation details.

JoanFM (approved):

Issue #1670, opened 2023-06-26
Refactor: use `Literal` in type hints


JohannesMessner:

srini047:
Hey, @JohannesMessner I would like to work on this issue.

JoanFM:
hey @srini047 ,

It would be great if you can take up this issue.

bpshaver:
Is `protobuf-array` also a possible value for `protocol` in the intended context? (If not... maybe the parameter name `protocol` should be used to refer to two different things.)

bpshaver:
Opened a [PR](https://github.com/docarray/docarray/pull/1827).

I'm in favor of an enum over a `Literal` consisting of different strings, but I can see how the latter is easier for end users. I can seek out other instances of this same issue if this PR looks good.

JoanFM:
Yes, it would be nice to make sure there are other instances or if we can close this

bpshaver:
I believe these parameters are candidates for similar treatment:

- [ ] `audio_codec`

- [ ]`audio_format`

- [ ] `video_codec`

But there may be more

PR #1680, opened 2023-06-28
Refactor: use `Literals` in type hints

srini047:
Fixes: #1670 

- I have made sure to `protocol` with `Literals` and options being ['pickle', 'protobuf']


JoanFM:
We need you to comply with our contributing guidelines and signoff the commits.

https://github.com/docarray/docarray/blob/main/CONTRIBUTING.md

JoanFM:
> @JoanFM I have signed the commits. Could you have a look at it?
> 
> image

THE CI is complaining about that, as well as mypy, black and ruff

srini047:
> > @JoanFM I have signed the commits. Could you have a look at it?
> > image
> 
> THE CI is complaining about that, as well as mypy, black and ruff

Well how do I proceed with. Any leads appreciated.
Thanks in advance!

docarray/store/helpers.py
```diff
@@ -131,12 +132,27 @@ def close(self):
 
 
 def _to_binary_stream(
+    self, protocol: Literal['pickle', 'protobuf'],
```
samsja, line 135: default value should still be `protobuf` here and we should not change the order of the argument (protocol should stay at the same place as before)

docarray/base_doc/mixins/io.py
```diff
@@ -146,7 +146,7 @@ def __bytes__(self) -> bytes:
         return self.to_bytes()
 
     def to_bytes(
-        self, protocol: str = 'protobuf', compress: Optional[str] = None
+        self, protocol: Literal['pickle', 'protobuf'], compress: Optional[str] = None
```
samsja, line 149: we need to keep the default vlaue

samsja (changes requested):
Thanks for you contribution ! Indeed litteral makes more sense in most of the case.

Two overall comments:
* Lets keep default value
* If possible we should not change the order of the parameters of existing function

docarray/base_doc/mixins/io.py
srini047: Updated the changes on all the occurrences.

docarray/store/helpers.py
srini047: I would want to make it clear if the `protocol` is in the right place and order (i.e. 2nd argument).

Is my understanding correct? At the same time let me make sure to add `protobuf ` as the default value.

srini047: Also, if we want `protocol` to act as a default argument then it has to be placed at the end.

samsja: I think you can do what you did on the other and kept it at the end

docarray/store/helpers.py
```diff
@@ -167,13 +183,28 @@ def _to_binary_stream(
 
 
 def _from_binary_stream(
+    self, protocol: Literal['pickle', 'protobuf'],
```
samsja, line 186: same here we can do like you id in the other

samsja (commented):
It starts to look very good. 

One idea, maybe we could wrap the `Literal['protobuf','pickle']` into a more formal type that we could call each time instead or repeating it ?

Something like this:

```python

Protocols = Literal['pickle','protobuf']

def my_func(protocol: Protocol):
   ...
```

Wdyt ?

srini047:
> It starts to look very good.
> 
> One idea, maybe we could wrap the `Literal['protobuf','pickle']` into a more formal type that we could call each time instead or repeating it ?
> 
> Something like this:
> 
> ```python
> Protocols = Literal['pickle','protobuf']
> 
> def my_func(protocol: Protocol):
>    ...
> ```
> 
> Wdyt ?

It increases the modularity of the code and maybe easier to maintain as you wish, what do you say?

docarray/store/helpers.py
srini047: image

Facing this error When I try to commit the changes. Any help on the same appreciated.

samsja:
> > It starts to look very good.
> > One idea, maybe we could wrap the `Literal['protobuf','pickle']` into a more formal type that we could call each time instead or repeating it ?
> > Something like this:
> > ```python
> > Protocols = Literal['pickle','protobuf']
> > 
> > def my_func(protocol: Protocol):
> >    ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Wdyt ?
> 
> It increases the modularity of the code and may be easier to maintain as well, what do you say?

then lets do it

docarray/store/helpers.py
srini047: @samsja Can you have a look at this? Meanwhile let me work on the function implementation.

srini047:
Hey @samsja!
- Please have a look at the approach if this helps. Or else let me revert to any other logic
- Also, I need help in `docarray/store/helpers.py` and I have tagged you in the issue I am facing.

docarray/base_doc/mixins/io.py
```diff
@@ -146,7 +147,10 @@ def __bytes__(self) -> bytes:
         return self.to_bytes()
 
     def to_bytes(
-        self, protocol: str = 'protobuf', compress: Optional[str] = None
+        self,
+        data,
```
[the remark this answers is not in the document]
samsja: what is this data ?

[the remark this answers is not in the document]
srini047: `data` is base64 encoded string. This is the same parameter available as in `from_base64` function.
I am actually using it to `_decompress_bytes`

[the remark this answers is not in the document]
srini047: @samsja Any updates?
Just tagging to ensure this doesn't remain the backlog.

samsja:
@srini047 can you install the pre commit hook and apply black, ruff and mypy for the CI to pass !

PR #1681, opened 2023-07-02
feat: support milvus

jupyterjazz:
Taking over from here https://github.com/docarray/docarray/pull/1666

- [x] Find (vector search)
- [x] Filter
- [x] Get/Del
- [x] Hybrid Search (find+filter)
- [x] Subindex
- [x] Tests

Functionality-wise everything's ready, only thing left is to make the execution faster (milvus job takes 13 mins on CI)

Docs will be done in a separate PR

Note:
I had to bump protobuf and tensorflow versions


docarray/index/backends/milvus.py
```diff
@@ -0,0 +1,841 @@
...
+        type_map = {
+            int: DataType.INT64,
+            float: DataType.FLOAT,
+            str: DataType.VARCHAR,
+            bytes: DataType.VARCHAR,
+            np.ndarray: DataType.FLOAT_VECTOR,
+            list: DataType.FLOAT_VECTOR,
+            AnyTensor: DataType.FLOAT_VECTOR,
+            AbstractTensor: DataType.FLOAT_VECTOR,
+        }
```
samsja, line 192: We can move this dict away from the function imo

samsja (commented):
added one minor comments

docarray/index/backends/milvus.py
jupyterjazz: hmm but all other backends have it inside

.github/workflows/ci.yml
```diff
@@ -198,7 +198,7 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install poetry
           poetry install --all-extras
-          poetry run pip install protobuf==3.19.0 # we check that we support 3.19
+          poetry run pip install protobuf==3.20.0 # we check that we support 3.19
```
samsja, line 201:
```suggestion
          poetry run pip install protobuf==3.20.0 # we check that we support 3.20
```

samsja (approved):

JoanFM (approved):

PR #1689, opened 2023-07-07
refactor: show what review means

JoanFM:


docarray/index/backends/redis.py
```diff
@@ -101,6 +94,23 @@ def __init__(self, db_config=None, **kwargs):
         self._create_index()
         self._logger.info(f'{self.__class__.__name__} has been initialized')
 
+    @property
+    def index_name(self):
+        default_index_name = (
+            self._schema.__name__.lower() if self._schema is not None else None
+        )
```
jupyterjazz, line 101: Ok I get the idea. However, I don't think this default naming convention is a good idea, we're essentially basing the index name on the schema class name which will be confusing to many.

```python

index1 = RedisDocumentIndex[MyDoc]()
index2 = RedisDocumentIndex[MyDoc]()
```
these should be two different indices but won't be with this convention

JoanFM: if u want to avoid this, u pass a different index name

jupyterjazz: Yeah, I see we're doing that for other backends too. Sounds good, I still think the other approach is slightly better and offers more control though

PR #1696, opened 2023-07-10
fix: fix search in memory with AnyEmbedding

JoanFM:
Allows to do easy search in Memory when AnyEmbedding is used as with `TextDoc` or `ImageDoc`:

```python
index = InMemoryExactNNIndex[TextDoc]()

docs = DocList[TextDoc](
    [TextDoc(text='hey', embedding=np.random.rand(128)) for _ in range(200)]
)
index.index(docs)
res = index.find_batched(docs[0:10], search_field='embedding', limit=5)
```

### DIFF
--- a/docarray/array/list_advance_indexing.py
+++ b/docarray/array/list_advance_indexing.py
@@ -1,5 +1,4 @@
 from typing import (
-    TYPE_CHECKING,
     Any,
     Iterable,
     List,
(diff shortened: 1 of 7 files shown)


jupyterjazz (commented):
why was `AnyEmbedding` failing?

JoanFM:
> why was `AnyEmbedding` failing?

It does not have a computational backend attached, so it cannot stack or do anything on tensors

samsja (commented):
I think it would be better to change 

`_extract_embedding` directly to look at the type to the AnyTensor

JoanFM:
> I think it would be better to change
> 
> `_extract_embedding` directly to look at the type to the AnyTensor

and what about the compBackend, is it okey If I return the NumpyBackend for AnyTensor as default?

samsja (approved):

PR #1706, opened 2023-07-18
fix: slow hnsw by caching num docs

jupyterjazz:
As the user reported, num_docs() operation is expensive and slows down the search. This PR addresses the issue by storing/caching num_docs and updating it after index/del

I tested on this simple code snippet and while index time increases slightly (from 6.13 to 6.14 seconds) we can see the speedup in the search time (from 0.0238 to 0.0018, 13x speedup)

```python
from docarray.index import HnswDocumentIndex
import numpy as np
import time

class MyDoc(BaseDoc):
    text: str
    embedding: NdArray[128]


docs = [MyDoc(text='hey', embedding=np.random.rand(128)) for _ in range(20000)]
index = HnswDocumentIndex[MyDoc](work_dir='tst', index_name='index')

index_start = time.time()
index.index(docs=DocList[MyDoc](docs))
index_time = time.time() - index_start

query = docs[0]

find_start = time.time()
matches, _ = index.find(query, search_field='embedding', limit=10)
find_time = time.time() - find_start

assert len(matches) == 10
assert query.id == matches[0].id
```


docarray/index/backends/hnswlib.py
```diff
@@ -403,7 +406,7 @@ def num_docs(self) -> int:
         """
         Get the number of documents.
         """
-        return self._get_num_docs_sqlite()
+        return self._num_docs
```
JoanFM, line 409: I think it may be better to handle it here totally: here do:

```python
self._num_docs = self._num_docs or self._get_num_docs_sqlite()
return self._num_docs
```

 And simply in all updates, deletes, index u set self._num_docs to 0.

JoanFM (changes requested):
small comment

docarray/index/backends/hnswlib.py
jupyterjazz: we update the value in the end of index and del, I think this way is more intuitive rather than setting it to 0

JoanFM (approved):

PR #1711, opened 2023-07-18
fix: weaviate handles lowercase index names

jupyterjazz:
WeaviateDocumentIndex capitalizes the index name when creating an index, so if we pass lowercase index name, there's a mismatch leading to various errors

```python
from docarray.index import WeaviateDocumentIndex
from pydantic import Field
import numpy as np

# Define the document schema.
class MyDoc(BaseDoc):
    title: str
    embedding: NdArray[128] = Field(is_embedding=True)

# Create dummy documents.
docs = DocList[MyDoc](MyDoc(title=f'title #{i}', embedding=np.random.rand(128)) for i in range(10))

# Initialize a new WeaviateDocumentIndex instance and add the documents to the index.
doc_index = WeaviateDocumentIndex[MyDoc](index_name='zdzd')
doc_index.index(docs)

# Perform a vector search.
query = np.ones(128)
retrieved_docs = doc_index.find(query, limit=10)
```

> File "/Users/jinaai/Desktop/docarray/docarray/index/backends/weaviate.py", line 354, in find
> docs, scores = self._find(
> File "/Users/jinaai/Desktop/docarray/docarray/index/backends/weaviate.py", line 408, in _find
> results["data"]["Get"][index_name], score_name
> KeyError: 'zdzd'
> 

Solution: add a post_init function that capitalizes the first letter of the index name

### DIFF
--- a/tests/index/weaviate/test_column_config_weaviate.py
+++ b/tests/index/weaviate/test_column_config_weaviate.py
@@ -48,3 +48,9 @@ def test_index_name():
 
     index = WeaviateDocumentIndex[StringDoc]()
     assert index.index_name == StringDoc.__name__
+
+    index = WeaviateDocumentIndex[StringDoc](index_name='Index_name')
+    assert index.index_name == 'Index_name'
+
+    index = WeaviateDocumentIndex[StringDoc](index_name='Index_name')
+    assert index.index_name == 'Index_name'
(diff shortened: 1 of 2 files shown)


tests/index/weaviate/test_column_config_weaviate.py
```diff
@@ -48,3 +48,9 @@ class StringDoc(BaseDoc):
 
     index = WeaviateDocumentIndex[StringDoc]()
     assert index.index_name == StringDoc.__name__
+
+    index = WeaviateDocumentIndex[StringDoc](index_name='Index_name')
+    assert index.index_name == 'Index_name'
+
+    index = WeaviateDocumentIndex[StringDoc](index_name='Index_name')
```
JoanFM, line 55: this test must be changed, it has to test non capital ones

JoanFM (changes requested):

PR #1713, opened 2023-07-19
feat: InMemoryExactNNIndex pre filtering

jupyterjazz:
Support arbitrary number of find/filter operations for `InMemoryExactNNIndex`, enabling pre+post filtering

From now on, you can build queries like this:
```python
query = (
    doc_index.build_query()
    .filter(filter_query={'price': {'$lte': 3}})
    .find(query=np.ones(10), search_field='tensor')
    .filter(filter_query={'text': {'$eq': 'hello 1'}})
    .build()
)
```

Note:
**how limits work**
Since developers could provide `limit` in any component of the query, and we also had an internal default value for `find`, the results were confusing. What I'm doing is the following: I'm not applying any limits during the operations, but I'm remembering the lowest one provided, and apply that limit in the end.

### DIFF
--- a/docarray/index/backends/in_memory.py
+++ b/docarray/index/backends/in_memory.py
@@ -291,11 +291,6 @@ class InMemoryExactNNIndex(BaseDocIndex, Generic[TSchema]):
             raise ValueError(
                 f'args and kwargs not supported for `execute_query` on {type(self)}'
             )
-        # find_res = _execute_find_and_filter_query(
-        #     doc_index=self,
-        #     query=query,
-        #     reverse_order=True,
-        # )
         find_res = self._hybrid_search(query)
         return find_res
 
@@ -331,6 +326,10 @@ class InMemoryExactNNIndex(BaseDocIndex, Generic[TSchema]):
                 out_docs = filter_docs(out_docs, op_kwargs['filter_query'])
             else:
                 raise ValueError(f'Query operation is not supported: {op}')
+
+        # if limit was not provided, use the default
+        if limit == sys.maxsize:
+            limit = 10
 
         out_docs = out_docs[:limit]
         scores_and_docs = zip([doc_to_score[doc.id] for doc in out_docs], out_docs)
--- a/docarray/index/backends/in_memory.py
+++ b/docarray/index/backends/in_memory.py
@@ -294,9 +294,7 @@ class InMemoryExactNNIndex(BaseDocIndex, Generic[TSchema]):
 
     def _find_and_filter(self, query: List[Tuple[str, Dict]]) -> FindResult:
         """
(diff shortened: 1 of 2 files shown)


docarray/index/backends/in_memory.py
```diff
@@ -293,13 +291,53 @@ def execute_query(self, query: List[Tuple[str, Dict]], *args, **kwargs) -> Any:
...
             )
-        find_res = _execute_find_and_filter_query(
-            doc_index=self,
-            query=query,
-            reverse_order=True,
-        )
+        find_res = self._hybrid_search(query)
         return find_res
 
+    def _hybrid_search(self, query: List[Tuple[str, Dict]]) -> FindResult:
```
JoanFM, line 297: hybrid search is not that. This is find and filter, let's change the name of this method for clarity

JoanFM (changes requested):

docarray/index/backends/in_memory.py
JoanFM: no, hybrid search is about mixing bm25 scores with embedding scores. Not about filtering + search

JoanFM (changes requested):
also test name to be changed

JoanFM (changes requested):

docarray/index/backends/in_memory.py
jupyterjazz: I think hybrid search means combining more than one search methods. bm25 + vector search is just one case.

Even we use it like that 😄 
https://docs.docarray.org/user_guide/storing/docindex/#hybrid-search-through-the-query-builder

docarray/index/backends/in_memory.py
```diff
@@ -293,12 +291,51 @@ def execute_query(self, query: List[Tuple[str, Dict]], *args, **kwargs) -> Any:
...
-            doc_index=self,
-            query=query,
-            reverse_order=True,
-        )
-        return find_res
+        return self._find_and_filter(query)
+
+    def _find_and_filter(self, query: List[Tuple[str, Dict]]) -> FindResult:
+        """
+        Executes a hybrid search on documents based on the provided query.
```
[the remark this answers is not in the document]
jupyterjazz: I think we should keep this term, we (and not only us) use it in docs and even though most ppl think of bm25 when talking about hybrid search, this is not wrong

JoanFM (changes requested):

docarray/index/backends/in_memory.py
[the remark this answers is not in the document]
JoanFM: we use find and filter which is what it is

JoanFM (changes requested):

PR #1718, opened 2023-07-20
feat: filtering in hnsw

jupyterjazz:
Support Filtering for `HnswDocumentIndex`, separately as well as inside the query builder

Main changes:
- Add filtrable fields separately to the sqlite table (so now we store `id`, `doc_blob`, and fields that can be filtered)
- Support `filter` function which does not touch hnsw and executes filtering on the sqlite table
- Refactor the query builder (hybrid search) to enable pre+post filtering. Post filtering is done in the same way; Pre filtering first queries the SQL table, gets hashed_ids for the filtered documents, and passes it to hnsw's new feature that can filter vectors based on their ids during search. 

Notes:
query builder became quite complex even tho I tried decomposing it as much as possible.
i think codecov has some problems?

### DIFF
--- a/docarray/index/backends/hnswlib.py
+++ b/docarray/index/backends/hnswlib.py
@@ -2,6 +2,7 @@ import glob
 import hashlib
 import os
 import sqlite3
+from collections import defaultdict
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import (
@@ -13,6 +14,7 @@ from typing import (
     List,
     Optional,
     Sequence,
+    Set,
     Tuple,
     Type,
     TypeVar,
@@ -30,16 +32,14 @@ from docarray.index.abstract import (
     _raise_not_composable,
     _raise_not_supported,
 )
-from docarray.index.backends.helper import (
-    _collect_query_args,
-    _execute_find_and_filter_query,
-)
+from docarray.index.backends.helper import _collect_query_args
 from docarray.proto import DocProto
 from docarray.typing.tensor.abstract_tensor import AbstractTensor
 from docarray.typing.tensor.ndarray import NdArray
 from docarray.utils._internal._typing import safe_issubclass
 from docarray.utils._internal.misc import import_library, is_np_int
-from docarray.utils.find import _FindResult, _FindResultBatched
+from docarray.utils.filter import filter_docs
+from docarray.utils.find import FindResult, _FindResult, _FindResultBatched
 
 if TYPE_CHECKING:
     import hnswlib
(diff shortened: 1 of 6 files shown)


tests/index/hnswlib/test_find.py
```diff
@@ -304,33 +304,6 @@ class MyDoc(BaseDoc):
         assert q.id == matches[0].id
 
 
-@pytest.mark.parametrize(
-    'find_limit, filter_limit, expected_docs', [(10, 3, 3), (5, None, 5)]
-)
-def test_query_builder_limits(find_limit, filter_limit, expected_docs, tmp_path):
```
JoanFM, line 310: why is this test not relevant anymore?

docarray/index/backends/hnswlib.py
```diff
@@ -167,21 +177,22 @@ class DBConfig(BaseDocIndex.DBConfig):
...
+                        'max_elements': 1024,
+                        'ef_construction': 200,
+                        'ef': 10,
+                        'M': 16,
+                        'allow_replace_deleted': True,
+                        'num_threads': 1,
+                    },
                 },
-                # `None` is not a Type, but we allow it here anyway
-                None: {},  # type: ignore
```
JoanFM, line 183: not really know why it was here, but why removing it?

docarray/index/backends/hnswlib.py
```diff
@@ -206,6 +217,15 @@ def python_type_to_db_type(self, python_type: Type) -> Any:
             if safe_issubclass(python_type, allowed_type):
                 return np.ndarray
 
+        type_map = {
```
JoanFM, line 220: so we must be clear on what items are filterable and which are not. To be mindful when documenting

JoanFM (changes requested):
I like it in general. Do we have an idea on the amount of extra space it takes to index all these into `SQLIte`? Should we have a `DBConfig` entry abou tthe fields that are to be filterable?

docarray/index/backends/hnswlib.py
jupyterjazz: I'll update docs here https://github.com/docarray/docarray/pull/1678

docarray/index/backends/hnswlib.py
jupyterjazz: I think it's because python_type_to_db_type function was returning None for types other than vector, but I changed it with defaultdict which returns an empty dict by default

JoanFM (approved):

PR #1720, opened 2023-07-21
fix: fix deepcopy torchtensor

JoanFM:


docarray/typing/tensor/torch_tensor.py
```diff
@@ -271,3 +271,10 @@ def _docarray_from_ndarray(cls: Type[T], value: np.ndarray) -> T:
     def _docarray_to_ndarray(self) -> np.ndarray:
         """cast itself to a numpy array"""
         return self.detach().cpu().numpy()
+
+    def new_empty(self, *args, **kwargs):
```
samsja, line 275: can we copy the full signature of the original method ?

JoanFM: if it is always the same, I do not think it makes sense, this is way more mantainable

samsja: But this break everything related to mypy and pycharm feature. In DocArray v2 we always repeat the full signature of the function

JoanFM: but mypy check passes

JoanFM: The method on TorchEmbedding does the same.
This is a method that noone should use
It is easier to forget to update this method than anything else.

samsja: okay oaky

samsja: mypy pass because it does not look at it. It will only check if there is type hint

PR #1723, opened 2023-07-24
fix: collection and index name in qdrant

JoanFM:
The subindex is initialized with a different_index name, so we need to make sure that the collection name is updated accordingly in the constructor.


jupyterjazz (commented):
I don't understand why this change was needed, can you add some context?

JoanFM:
Check the issue and the test. The 0.36.0 version does not work. The `subindex` is initialized with a different_index name, so we need to make sure that the collection name is updated accordingly in the constructor.

jupyterjazz (approved):

PR #1724, opened 2023-07-24
feat: update for inmemory index

jupyterjazz:
Implements Update for `InMemoryExactNNIndex` by creating a mapping between document IDs and their corresponding positions in the DocList.

In terms of performance, we see significant ~130x (from 0.0266s to 0.0002s) improvement in get, and a slight increase in the index time - approx. from 3.50s to 3.57s; Other operation times stayed very similar. Experiment was made on 200 000 docs, vector dim 128.

### DIFF
--- a/docarray/index/backends/in_memory.py
+++ b/docarray/index/backends/in_memory.py
@@ -94,6 +94,7 @@ class InMemoryExactNNIndex(BaseDocIndex, Generic[TSchema]):
                 )()
 
         self._embedding_map: Dict[str, Tuple[AnyTensor, Optional[List[int]]]] = {}
+        self._ids_to_positions: Dict[str, int] = {}
 
     def python_type_to_db_type(self, python_type: Type) -> Any:
         """Map python type to database type.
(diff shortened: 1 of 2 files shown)


JoanFM:
Can you also add how much more memory is used? I am not sure this is desired.

jupyterjazz:
@JoanFM 

Not much change in terms of memory too.

Here's the experiment code
```python
from docarray.index import InMemoryExactNNIndex
from docarray import BaseDoc
import numpy as np
import time
import tracemalloc

tracemalloc.start()

class MyDoc(BaseDoc):
    text: str
    embedding: NdArray[128]


data = [MyDoc(text=f'text {i}', embedding=np.random.rand(128)) for i in range(200000)]
doc_index = InMemoryExactNNIndex[MyDoc]()

doc_index.index(data)

start_find = time.time()
docs, scores = doc_index.find(data[0], search_field='embedding')
end_find = time.time()

if data[10] in data:
    print('wohoo')

ids_to_get = [data[200].id, data[250].id, data[350].id]

ids_to_del = [data[200].id, data[250].id, data[350].id]
del doc_index[ids_to_del]


print(tracemalloc.get_traced_memory())

tracemalloc.stop()
```
Before change prints (768773718, 1032141402) showing current memory usage and max memory usage
After change - (784741328, 1048109012)
So there's ~2% increase

JoanFM (approved):

PR #1727, opened 2023-07-26
refactor: hnswlib performance

JoanFM:


docarray/index/backends/hnswlib.py
```diff
@@ -332,7 +340,19 @@ def _filter(
         limit: int,
     ) -> DocList:
         rows = self._execute_filter(filter_query=filter_query, limit=limit)
-        return DocList[self.out_schema](self._doc_from_bytes(blob) for _, blob in rows)  # type: ignore[name-defined]
+        hashed_ids = [doc_id for doc_id, _ in rows]
+        embeddings: OrderedDict[str, list] = OrderedDict()
```
samsja, line 344: Why specifically OrderedDict? I think normal dict in python will already be ordered by default

docarray/index/backends/hnswlib.py
```diff
@@ -127,7 +132,12 @@ def __init__(self, db_config=None, **kwargs):
             else:
                 self._hnsw_indices[col_name] = self._create_index(col_name, col)
                 self._logger.info(f'Created a new index for column `{col_name}`')
+            if self._hnsw_indices[col_name].space == 'cosine':
+                cosine_metric_index_exist = True
```
jupyterjazz, line 136: why do we care about this?

jupyterjazz (commented):
do we have any measure on how much faster hnsw works now?

docarray/index/backends/hnswlib.py
JoanFM: I would still like to be specific about the requirement for order

JoanFM:
> do we have any measure on how much faster hnsw works now?

We have observed around 10% improvement

jupyterjazz (approved):

PR #1729, opened 2023-07-26
refactor: do not recompute every time num_docs

JoanFM:


docarray/index/backends/hnswlib.py
```diff
@@ -628,8 +627,11 @@ def accept_hashed_ids(id):
 
         # If hashed_ids is provided, k is the minimum of limit and the length of hashed_ids; else it is limit
         k = min(limit, len(hashed_ids)) if hashed_ids else limit
-
-        labels, distances = index.knn_query(queries, k=k, **extra_kwargs)
+        try:
+            labels, distances = index.knn_query(queries, k=k, **extra_kwargs)
+        except RuntimeError:
+            k = min(k, self.num_docs())
+            labels, distances = index.knn_query(queries, k=k, **extra_kwargs)
```
jupyterjazz, line 634: we should just make sure that value of `k` is not exceeding number of vectors, no need for try catch

jupyterjazz (changes requested):
value of `k` should not be more than number of vectors, otherwise lgtm

docarray/index/backends/hnswlib.py
jupyterjazz: I already modified the code so you can revert this change

JoanFM: but this would involve having a call to num_docs very often which is what I am trying to avoid

jupyterjazz: why does it matter so much it's cached

JoanFM: I have seen it take 13 seconds for when the index is of the order of 1M documents. Just to solve a couple of very edge cases I do not think is worth

jupyterjazz (approved):
I wasn't aware of this issue. Let's add 1-2 sentences in the code to explain why we're doing this

PR #1745, opened 2023-08-05
feat(index): allow passing negative limits for find method

vkrishnamurthy11:


jupyterjazz (commented):
Hi @vkrishnamurthy11, this looks good, I think `get_current_count` is a reasonable way to get the number without touching the sql table.

Can you please add tests and implement this functionality for other backends too?

PR #1746, opened 2023-08-07
fix: fix readme test

samsja:
# Context

will fix broken ci and test coverage on main

### DIFF
--- a/tests/documentation/test_docs.py
+++ b/tests/documentation/test_docs.py
@@ -72,5 +72,12 @@ def test_readme():
     check_md_file(
         fpath='README.md',
         memory=True,
-        keyword_ignore=['tensorflow', 'fastapi', 'push', 'langchain', 'MovieDoc'],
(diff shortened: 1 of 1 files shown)


JohannesMessner:
Not sure to understand how this fixes it

samsja:
> Not sure to understand how this fixes it

basically the test is falling bc we try to run the jina part of the readme without installing jina. I just added jina to the ignore argument like we do with fastapi and all. So it should fix it (hopefully)

JohannesMessner:
I see, that makes sense.
But the readme alone can't crash our coverage from ~80% to ~30%, right?

JohannesMessner (approved):

samsja:
> I see, that makes sense. But the readme alone can't crash our coverage from ~80% to ~30%, right?

I have no clue honestly

jupyterjazz (approved):

PR #1764, opened 2023-08-24
solved the good first issue in the hnswlib.py file

shobhit9957:
Hey Joan, I've done it, according to the issue. I hope it was correct! Thanks for helping Joan.

### DIFF
--- a/docarray/index/backends/hnswlib.py
+++ b/docarray/index/backends/hnswlib.py
@@ -428,7 +428,7 @@ class HnswDocumentIndex(BaseDocIndex, Generic[TSchema]):
         Get the number of documents.
         """
         if self._num_docs == 0:
-            self._num_docs = self._get_num_docs_sqlite()
+            self._num_docs = self._get_num_docs_hnsw()
         return self._num_docs
 
     ###############################################
--- a/docarray/index/backends/hnswlib.py
+++ b/docarray/index/backends/hnswlib.py
@@ -430,7 +430,13 @@ class HnswDocumentIndex(BaseDocIndex, Generic[TSchema]):
         if self._num_docs == 0:
             self._num_docs = self._get_num_docs_hnsw()
         return self._num_docs
-
+    def _get_num_docs_hnsw(self) -> int:
+        """
+        Get the number of documents using the HNSW method.
+        """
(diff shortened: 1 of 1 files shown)


JoanFM:
Hey @shobhit9957,

Sorry for the missunderstanding, this is correct, but you will need to implement the private method.

shobhit9957:
Can you please elaborate on that meaning private method..., how can I do that in private method

JoanFM:
> hey joan, once check this :
> 
> ```python
> def num_docs(self) -> int:
>     """
>     Get the number of documents.
>     """
>     if self._num_docs == 0:
>         # Replace the old method with the new one
>         self._num_docs = self._get_num_docs_hnsw()
>     return self._num_docs
> 
> def _get_num_docs_hnsw(self) -> int:
>     """
>     Get the number of documents using the HNSW method.
>     This method should return the count of documents using the new technique.
>     """
>     # Implement your logic here to get the count using the HNSW method
>     # For example, you might access some data or perform calculations
>     # that help you quickly determine the number of documents.
>     # Then return the count.
>     return calculated_num_docs
> ```
> 
> is this correct?

Yes, and this computation shpuld come from inside the hnsw indices

docarray/index/backends/hnswlib.py
```diff
@@ -430,7 +430,13 @@ def num_docs(self) -> int:
         if self._num_docs == 0:
             self._num_docs = self._get_num_docs_hnsw()
         return self._num_docs
-
+    def _get_num_docs_hnsw(self) -> int:
+        """
+        Get the number of documents using the HNSW method.
+        """
+        # Access your HNSW index here and retrieve the count of documents
+        hnsw_num_docs = self.hnsw_index.get_document_count()
```
JoanFM, line 438: I think there is not a `self.hnsw_index` attribute. I think there are multiples inside the `self._hnsw_indices`.

I think u can use the first inside the map to retrieve the count.

Also please make sure to comply with the style contribution standards that you can find in https://github.com/docarray/docarray/blob/main/CONTRIBUTING.md

Also, make sure to signoff your commits so that we can merge your contribution

JoanFM (changes requested):

PR #1773, opened 2023-09-05
docs: explain how to set document config

JohannesMessner:
there is a small things that users need to be aware of when setting configs, so i am explaining that here.

~~TODO: small test~~ done

### DIFF
--- a/docs/user_guide/representing/first_step.md
+++ b/docs/user_guide/representing/first_step.md
@@ -117,6 +117,21 @@ This representation can be used to [send](../sending/first_step.md) or [store](.
 
     [BaseDoc][docarray.base_doc.doc.BaseDoc] can be nested to represent any kind of data hierarchy.
 
+## Setting a Pydantic `Config` class
+
+Documents support setting a `Config` [like any other Pydantic `BaseModel`](https://docs.pydantic.dev/latest/usage/model_config/).
+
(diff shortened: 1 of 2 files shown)


samsja (approved):

JohannesMessner:
@samsja what will be the effect of pydantic v2 on this?

samsja:
> @samsja what will be the effect of pydantic v2 on this?

this way will still be compatible with pydantic v2 but pydantic will show a warning saying it is deprecated. I suggest we update the docs after the merge

docs/user_guide/representing/first_step.md
```diff
@@ -117,6 +117,21 @@ This representation can be used to [send](../sending/first_step.md) or [store](.
...
+Documents support setting a `Config` [like any other Pydantic `BaseModel`](https://docs.pydantic.dev/latest/usage/model_config/).
+
+However, if you set a config, you should inherit from the `BaseDoc` config class:
+
+```python
+from docarray import BaseDoc
+
+
+class MyDoc(BaseDoc):
+    class Config(BaseDoc.Config):
```
deepankarm, line 131: Minor comment. In pydantic v1, we only had to pass a `class Config` in the model. It didn't need to be inherited from another config class.

PR #1777, opened 2023-09-06
fix: skip doc attributes in __annotations__ but not in __fields__

JoanFM:
Porting fix https://github.com/jina-ai/jina/pull/6035 to docarray

### DIFF
--- a/docarray/utils/create_dynamic_doc_class.py
+++ b/docarray/utils/create_dynamic_doc_class.py
@@ -51,6 +51,8 @@ def create_pure_python_type_model(model: Any) -> BaseDoc:
     """
     fields: Dict[str, Any] = {}
     for field_name, field in model.__annotations__.items():
+        if field_name not in model.__fields__:
+            continue
         field_info = model.__fields__[field_name].field_info
         try:
             if safe_issubclass(field, DocList):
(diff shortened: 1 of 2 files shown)


NarekA:
Thanks for doing this!

docarray/utils/create_dynamic_doc_class.py
```diff
@@ -51,6 +51,8 @@ class MyDoc(BaseDoc):
     """
     fields: Dict[str, Any] = {}
     for field_name, field in model.__annotations__.items():
+        if field_name not in model.__fields__:
+            continue
         field_info = model.__fields__[field_name].field_info
```
NarekA, line 56: FWIW: It might be worth using something like `pydash.get` here in case traversal breaks at a different location.

NarekA (approved):

JoanFM:
Actually, @NarekA it would be nice if you could open this PR on your own so that we can properly credit the actual contributor. I just did not want to annoy you with it, but since you have already seen this and you are now aware of this need, it may be good to have this.

NarekA:
Opening a new PR now

NarekA:
Here: https://github.com/docarray/docarray/pull/1779

Issue #1782, opened 2023-09-13
Bug with DocList[]


samsja:
# Context
```python
from docarray import DocList, BaseDoc

docs = DocList[BaseDoc()]# 1

class MyDoc(BaseDoc):
    text: str

DocList[MyDoc][MyDoc] # 2
```

Both 1 and 2 lines should failed with a nice error message

## description

DocList[X] should only accept BaseDoc class and should not be called twice

## How to do it

This should be as easy as adding if else statement and raise properly error in this part of the codebase:

https://github.com/docarray/docarray/blob/715252a72177e83cb81736106f34d0ab2960ce56/docarray/array/any_array.py#L56

punndcoder28:
Hi @samsja I can work on this. From the docs I understand that the correct way is
```
from docarray import DocList, BaseDoc

class MyDoc(BaseDoc):
    text: str

docs = DocList[BaseDoc]([MyDoc])
```

punndcoder28:
Understood. Will go through the custom implementation of `safe_issubclass` method and propose a solution for this issue

JoanFM:
I do not think is that code u have to check, is the specific line, the `safe_issubclass` method should not be touched I think

punndcoder28:
Not related to this issue but how do I test out my changes by locally building the library and importing it in a local python script

punndcoder28:
Checked the implementation of classes `DocList` and `AnyDocArray` and noticed that https://github.com/docarray/docarray/blob/d0b99909052d1640670764015e4e385319fdc776/docarray/array/doc_list/doc_list.py#L338
the above line fails when we do
```
from docarray import DocList, BaseDoc

doc = DocList[BaseDoc()]
```
and the control goes to
https://github.com/docarray/docarray/blob/d0b99909052d1640670764015e4e385319fdc776/docarray/array/doc_list/doc_list.py#L341
but does not end up in the class `AnyDocArray` so we can handle this case in `__class_getitem__` method of `DocList` by adding a condition to check
`if isinstance(item, object)` and raise an exception of type `ValueError`

JoanFM:
> Checked the implementation of classes `DocList` and `AnyDocArray` and noticed that
> 
> https://github.com/docarray/docarray/blob/d0b99909052d1640670764015e4e385319fdc776/docarray/array/doc_list/doc_list.py#L338
> 
> 
> the above line fails when we do
> ```
> from docarray import DocList, BaseDoc
> 
[...]

PR #1794, opened 2023-09-16
fix: raise exception when type of DocList is object

punndcoder28:
* Fixes being able to pass an object for `DocList` without causing code to raise expection as mentioned in #1782 

Test Plan - Running the below code raises exception
```
from docarray import DocList, BaseDoc

doc = DocList[BaseDoc()]
```

Exception raise
```
  File "test.py", line 3, in <module>
    doc = DocList[BaseDoc()]
  File "docarray/docarray/array/doc_list/doc_list.py", line 340, in __class_getitem__
    raise TypeError(f'Expecting a type, got {item} of type {type(item)}')
TypeError: Expecting a type, got BaseDoc(id='9ab01e333a55195e5bb3f97604b97f10') of type <class 'docarray.base_doc.doc.BaseDoc'>
```


docarray/array/doc_list/doc_list.py
```diff
@@ -337,8 +337,11 @@ def __class_getitem__(cls, item: Union[Type[BaseDoc], TypeVar, str]):
 
         if isinstance(item, type) and safe_issubclass(item, BaseDoc):
             return AnyDocArray.__class_getitem__.__func__(cls, item)  # type: ignore
-        else:
-            return super().__class_getitem__(item)
+
+        if isinstance(item, object) and not isinstance(item, TypeVar):
```
punndcoder28, line 341: added a check `isinstance(item, TypeVar)` since item of `TypeVar` was also being passed when `__class_getitem__` was called

punndcoder28:
@samsja @JoanFM I am unsure on how to handle the case `doc = DocList[MyDoc][MyDoc]` because the item parameter which gets passed to `__class_getitem__` of both `AnyDocArray` and `DocList` is of type `BaseDoc`

JoanFM:
Hey @punndcoder28, please make sure to add relevant tests for this

docarray/array/doc_list/doc_list.py
```diff
@@ -337,8 +337,14 @@ def __class_getitem__(cls, item: Union[Type[BaseDoc], TypeVar, str]):
 
         if isinstance(item, type) and safe_issubclass(item, BaseDoc):
             return AnyDocArray.__class_getitem__.__func__(cls, item)  # type: ignore
-        else:
-            return super().__class_getitem__(item)
+        if (
```
JoanFM, line 340: should we better have an `elif`?

JoanFM (changes requested):

docarray/array/doc_list/doc_list.py
punndcoder28: Since each condition has a terminating condition from the function wouldn't multiple `if`'s be similar to `elif`? If there was a scenario like below an `elif` would definitely make it much better. In the above case, we check the instance of `item` and based on the condition we return a type or raise an exception. But if the code base follows `if ... elif ... else` blocks I would love to keep it uniform so that it is easy to maintain
```
computation
if condn1:
   computation = value1
elif condn2:
   computation = value2
else:
   computation = value3

return computation

```

punndcoder28:
@JoanFM Can we pass string literal as types for a class? Like so
```
    def g(a: DocList['BaseDoc']) -> DocList['BaseDoc']:
        return a
```

This is being done here https://github.com/docarray/docarray/blob/8f32866e1cd3aa139a88cfca2beaa502656dc76b/tests/units/array/test_array.py#L383

JoanFM:
> @JoanFM Can we pass string literal as types for a class? Like so
> 
> ```
>     def g(a: DocList['BaseDoc']) -> DocList['BaseDoc']:
>         return a
> ```
> 
> This is being done here
> 
> https://github.com/docarray/docarray/blob/8f32866e1cd3aa139a88cfca2beaa502656dc76b/tests/units/array/test_array.py#L383

u can in the type hints

JoanFM (approved):

PR #1800, opened 2023-09-19
fix: fix double subscriptable error

JoanFM:


tests/units/array/test_array.py
```diff
@@ -493,7 +492,24 @@ def test_parameterize_list():
...
+    from docarray import DocList, DocVec
+    from docarray.documents import TextDoc
+
+    with pytest.raises(TypeError) as excinfo:
+        da = DocList[TextDoc][TextDoc]
+        assert da is None
+
+    assert 'not subscriptable' in str(excinfo.value)
+
+    with pytest.raises(TypeError) as excinfo:
```
punndcoder28, line 511: I don't know if I am allowed to comment on PR but this looks like duplicated code

JoanFM: this is on tests, nothing wrong about that

samsja (approved):

PR #1802, opened 2023-09-21
fix: milvus connection para missing

lvzii:
missing parameter port


JoanFM (approved):

JoanFM:
Thank you very much for the fix!

PR #1803, opened 2023-09-25
feat: add BaseDocWoId

samsja:
# Context

allow to have doc without Id

### DIFF
--- a/docarray/base_doc/doc.py
+++ b/docarray/base_doc/doc.py
@@ -61,38 +61,11 @@ T_update = TypeVar('T_update', bound='UpdateMixin')
 ExcludeType = Optional[Union['AbstractSetIntStr', 'MappingIntStrAny']]
 
 
-class BaseDoc(BaseModel, IOMixin, UpdateMixin, BaseNode):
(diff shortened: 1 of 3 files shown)


deepankarm (approved):

docarray/base_doc/doc.py
```diff
@@ -61,39 +61,12 @@
...
-        image: ImageUrl
-
-
-    doc = MyDoc(embedding=np.zeros(512), image='https://example.com/image.jpg')
-    ```
-
-
-    BaseDoc is a subclass of [pydantic.BaseModel](
-    https://docs.pydantic.dev/usage/models/) and can be used in a similar way.
+    BaseDocWoId is the class behind BaseDoc, it should not be used directly unless you know what you are doing.
```
JoanFM, line 66: Add that it will not be possible to use if with DocumentIndices

JoanFM (changes requested):

docarray/base_doc/doc.py
samsja: good point

samsja: done

PR #1823, opened 2023-10-18
fix: from_dataframe with numpy==1.26.1 and type handling in python 3.9

### LINKED ISSUE #1821: bug: deserializing from dataframe with numpy==1.26.1
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://docs.docarray.org) and still think this feature is missing

### Description

I am looking for an efficient way of creating a DocList or DocVec from a large DataFrame. Currently I am converting the DataFrame to a Dict and then looping over this to create the various objects which is slow.

I did come across the from_dataframe method: https://docs.docarray.org/API_reference/array/da_stack/#docarray.array.doc_vec.doc_vec.DocVec.from_dataframe

However this does not support list like fields, which is quite common when you have field that contains the text embeddings.

### Affected Components

- [ ] [Vector Database / Index](https://docs.docarray.org/user_guide/storing/docindex/)
- [ ] [Representing](https://docs.docarray.org/user_guide/representing/first_step)
- [X] [storing](https://docs.docarray.org/user_guide/storing/first_step/)
- [ ] [multi modal data type](https://docs.docarray.org/data_types/first_steps/)

JohannesMessner:
We perform a check that doesn't work anymore with new versions of numpy, because they switches equality semantics to broadcast over every element in the array.

This PR fixes that by treating np arrays as a special cases in the affected check. It also does the same for tf, torch, and jax, just because == semantics on tensor-like objects can always be a bit surprising.

The numpy version in question (`1.26.1`) doesn't support Python 3.8 or down, but we do. Therefore, poetry can't lock to that version. To circumvent this, I am here manually installing the "problematic" numpy version in the CI. For the same reason **I am bumping some tests to run on Python 3.9**.

No new tests since the existing tests already fail with the numpy version in question, and are fixed by the fix (see failing check of the CI run after the new numpy version, but before the fix was pushed, [here](https://github.com/docarray/docarray/actions/runs/6560445470/job/17818195764?pr=1823)).

**Also includes a fix to handle typing changes in python 3.9 and above**, by using the `get_args` helper instead of `.__args__`.

closes #1821

### DIFF
--- a/docarray/display/document_summary.py
+++ b/docarray/display/document_summary.py
@@ -1,4 +1,4 @@
-from typing import Any, List, Optional, Type, Union
+from typing import Any, List, Optional, Type, Union, get_args
 
 from rich.highlighter import RegexHighlighter
 from rich.theme import Theme
@@ -83,7 +83,7 @@ class DocumentSummary:
 
                 if is_union_type(field_type) or is_optional_type(field_type):
                     sub_tree = Tree(node_name, highlight=True)
-                    for arg in field_type.__args__:
+                    for arg in get_args(field_type):
                         if safe_issubclass(arg, BaseDoc):
                             sub_tree.add(
                                 DocumentSummary._get_schema(
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.9]
         pydantic-version: ["pydantic-v2", "pydantic-v1"]
         test-path: [tests/integrations, tests/units, tests/documentation]
     steps:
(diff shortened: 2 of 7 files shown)


JoanFM (changes requested):
Will a np.array of one dimension with 0 be considered a None?

JohannesMessner:
> Will a np.array of one dimension with 0 be considered a None?

No, it will not. It also wasn't before.

JoanFM:
> > Will a np.array of one dimension with 0 be considered a None?
> 
> No, it will not. It also wasn't before.

can we add a test for it?

docarray/display/document_summary.py
```diff
@@ -1,4 +1,4 @@
-from typing import Any, List, Optional, Type, Union
+from typing import Any, List, Optional, Type, Union, get_args
```
JoanFM, line 1: does this exist in Python 3.8?

JoanFM (changes requested):

docarray/display/document_summary.py
JoanFM: Ok it is

PR #1829, opened 2023-10-31
feat(array): allow for type object with DocList

### LINKED ISSUE #1828: DocList raises exception for type object.
### Initial Checks

- [X] I have read and followed [the docs](https://docs.docarray.org/) and still think this is a bug

### Description

This [commit](https://github.com/docarray/docarray/commit/2f3b85e333446cfa9b8c4877c4ccf9ae49cae660) introduced a check to verify that DocList is not used with an object: 
```
        if (
            isinstance(item, object)
            and not is_typevar(item)
            and not isinstance(item, str)
            and item is not Any
        ):
               raise TypeError('Expecting a type, got object instead')
 ```

This is quite a broad condition (it breaks things like `DocList[TorchTensor]`, or nested `DocList[DocList[...]]` for me for instance), because:
- Almost everything will be an object so the first line if almost a catch-all. 
- `is_typevar` only check for TypeVar objects.

Should this not check as well for something `not isinstance(item, type)` instead of `not is_typevar` to allow for classes?
This way only non class objects (like instances of class object that are not classes themselves) will raise the `TypeError`. 

### Example Code

```Python
from docarray import DocList
test = DocList[TorchTensor]
```

### Python, DocArray & OS Version

```Text
0.39.1
```

### Affected Components

- [ ] [Vector Database / Index](https://docs.docarray.org/user_guide/storing/docindex/)
- [X] [Representing](https://docs.docarray.org/user_guide/representing/first_step)
- [ ] [storing](https://docs.docarray.org/user_guide/storing/first_step/)
- [X] [multi modal data type](https://docs.docarray.org/data_types/first_steps/)

corentinmarek:
See issue https://github.com/docarray/docarray/issues/1828 for context

### DIFF
--- a/docarray/array/doc_list/doc_list.py
+++ b/docarray/array/doc_list/doc_list.py
@@ -343,6 +343,7 @@ class DocList(
         if (
             isinstance(item, object)
             and not is_typevar(item)
+            and not isinstance(item, type)
             and not isinstance(item, str)
             and item is not Any
         ):


JoanFM (commented):
May we add a test to show what is exactly what we are aiming to enable or disable?

PR #1835, opened 2023-12-04
feat(index): add epsilla connector

tonyyanga:
This PR adds the `EpsillaDocumentIndex`. It allows docarray users to use Epsilla vector database for queries.

It is the 1st PR for issue https://github.com/docarray/docarray/issues/1832. I intend to make a follow-up PR to add documentation for the connector, but want to limit the size of the PR to make it easier to review.


docarray/index/backends/epsilla.py
```diff
@@ -0,0 +1,510 @@
...
+                "column is of a vector type with the is_embedding=True attribute specified."
+            )
+        elif len(vector_columns) > 1:
+            raise ValueError("Specifying multiple vector fields is not supported.")
+
+    def _create_table_self_hosted(self):
+        """Use _column_infos to create a table in the database."""
+        table_fields = []
+
+        primaryKeys = []
```
JoanFM, line 128: please use `primary_keys`

docarray/index/backends/epsilla.py
```diff
@@ -0,0 +1,510 @@
...
+
+        def validate_self_hosted_config(self):
+            if not all(
+                [self.protocol, self.host, self.port, self.db_path, self.db_name]
+            ):
+                raise ValueError("Missing required attributes for self-hosted version.")
+
+        def validate_cloud_config(self):
+            if not all([self.cloud_project_id, self.cloud_db_id, self.api_key]):
+                raise ValueError("Missing required attributes for cloud version.")
```
JoanFM, line 282: it would be nice to show which ones are missing?

docarray/index/backends/epsilla.py
```diff
@@ -0,0 +1,510 @@
...
+            if self.is_self_hosted:
+                self.validate_self_hosted_config()
+            else:
+                self.validate_cloud_config()
+
+        def validate_self_hosted_config(self):
+            if not all(
+                [self.protocol, self.host, self.port, self.db_path, self.db_name]
+            ):
+                raise ValueError("Missing required attributes for self-hosted version.")
```
JoanFM, line 278: would be nice to know what is missinf

docarray/index/backends/epsilla.py
```diff
@@ -0,0 +1,510 @@
...
+                        'dataType': column_info.db_type,
+                        'dimensions': dim,
+                    }
+                )
+
+        status_code, response = self._db.create_table(
+            table_name=self._table_name,
+            table_fields=table_fields,
+        )
+        if status_code != 200:
```
JoanFM, line 166: is there a constant feom the library that could allow us to avoid using magic numbers?

JoanFM (changes requested):
Thanks for the nice contribution, I have some comments about the PR.

tonyyanga:
@JoanFM Thanks for the review!

I have updated the input validation logic to report specific parameters missing.

It is unfortunate that Epsilla's python library does not include pythonic handling of error (throw instead of status codes). It directly returns HTTP status codes. For now I replaced it with from http import HTTPStatus.

When there is a conflict on loading the db, Epsilla returns HTTP 500 when it should return HTTP 409. I have filed [an issue](https://github.com/epsilla-cloud/vectordb/issues/93), but parsing the string error message is the best workaround for now.

What do you think? Please let me know if there are other issues we should address.

docarray/index/backends/epsilla.py
```diff
@@ -60,10 +61,12 @@ def __init__(self, db_config=None, **kwargs):
                 db_path=self._db_config.db_path,
             )
 
-            if status_code != 200:
-                if status_code == 500 and (
-                    "already loaded" in response["message"]
-                    or "already exists" in response["message"]
+            if status_code != HTTPStatus.OK:
+                # Epsilla returns HTTP 500 when multiple clients connect to the same db
+                # Bug filed at https://github.com/epsilla-cloud/vectordb/issues/93
```
JoanFM, line 66: if the quick is fix on Epsilla side, would it be possible to wait for a patch on epsille side?

JoanFM:
> @JoanFM Thanks for the review!
> 
> I have updated the input validation logic to report specific parameters missing.
> 
> It is unfortunate that Epsilla's python library does not include pythonic handling of error (throw instead of status codes). It directly returns HTTP status codes. For now I replaced it with from http import HTTPStatus.
> 
> When there is a conflict on loading the db, Epsilla returns HTTP 500 when it should return HTTP 409. I have filed [an issue](https://github.com/epsilla-cloud/vectordb/issues/93), but parsing the string error message is the best workaround for now.
> 
> What do you think? Please let me know if there are other issues we should address.

Do you plan to have the fix for that behavior anytime soon on epsilla side?

tonyyanga:
Epsilla is going to merge a quick fix the `load_db` behavior: https://github.com/epsilla-cloud/vectordb/issues/93 Will wait for them.

docarray/index/backends/epsilla.py
```diff
@@ -0,0 +1,527 @@
...
+        )
+
+        if status_code != HTTPStatus.OK:
+            raise IOError(
+                f"Failed to insert documents. "
+                f"Error code: {status_code}. Error message: {response}."
+            )
+
+    def num_docs(self) -> int:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
tonyyanga: There is currently no public API for counting the number of docs. One workaround is to run `filter` query without any condition, but I worry that is quite inefficient. Would you prefer to implement `num_docs` with that rather than throwing here?

tonyyanga:
Epsilla fixed on their end. I've pushed a change to use HTTP 409 status codes rather than relying on error messages.

Love to know what you think about the `num_docs` implementation. :-)

docarray/index/backends/epsilla.py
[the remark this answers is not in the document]
tonyyanga: I double checked. `num_docs` impacts these two methods:
- `__len__` 
- `_is_index_empty`

I pushed a `_filter` based implementation for `_is_index_empty`. Confirmed there is nothing that depends on `num_docs` is limited (nothing in `abstract.py` depends on this). But there is no easy way to check all the `len()` locations.

```
$ grep -rn -l --exclude-dir=./tests "num_docs()"
./docarray/index/abstract.py
./docarray/index/backends/milvus.py
./docarray/index/backends/redis.py
./docarray/index/backends/hnswlib.py
./docs/user_guide/storing/index_hnswlib.md
./docs/user_guide/storing/index_elastic.md
./docs/user_guide/storing/index_in_memory.md
./docs/user_guide/storing/index_qdrant.md
./docs/user_guide/storing/index_redis.md
```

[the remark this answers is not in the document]
JoanFM: and _is_index_empty, I think it was checked in many queries. I am not sure if the result is kind of cached. I have little access to check properly now the code

tonyyanga:
@JoanFM Any other issues to address before we can merge this change?

JoanFM:
I will review the implications of num_docs in detail on Monday.

JoanFM:
Is it possible for you to have an API that decides whether the table is empty or not? Otherwise, u can try to override `_is_index_empty` to say always `False`. And make sure that all the methods work when the table is empty returning empty results properly.

tonyyanga:
@JoanFM 

Unfortunately, there is no API for that. I have pushed a change to override `_is_index_empty` to return `False`, and added test cases with the index being empty.

Please let me know if there are other things to address.

tonyyanga:
Hey @JoanFM 

Gently bumping this - please let me know if there are other things to address.

JoanFM:
Hello @tonyyanga ,

Thanks for the contribution, we will be waiting for the documentation PR

tonyyanga:
Thank you! will follow up with it, hopefully later this week.

Issue #1824, opened 2023-10-21
Integrate new Vector search library from Spotify


JoanFM:
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://docs.docarray.org) and still think this feature is missing

### Description

Spotify has released VOYAGER, a new vector search library 

I propose to have a VoyagerIndex expecting to be very similar to HNSWIndex.

### Affected Components

- [X] [Vector Database / Index](https://docs.docarray.org/user_guide/storing/docindex/)
- [ ] [Representing](https://docs.docarray.org/user_guide/representing/first_step)
- [ ] [Sending](https://docs.docarray.org/user_guide/sending/first_step/)
- [ ] [storing](https://docs.docarray.org/user_guide/storing/first_step/)
- [ ] [multi modal data type](https://docs.docarray.org/data_types/first_steps/)

darshi1337:
@JoanFM 

I'm eager to contribute to the "Voyager Index Implementation" (#1824) in DocArray. Please consider assigning me the task. I'm prepared to:
- [ ] Implement Voyager Index
- [ ] Documentation
- [ ] Testing

JoanFM:
Perfect @darshi1337

darshi1337:
@JoanFM 
Just saw that you assigned me. Will start working shortly

PR #1846, opened 2024-01-11
feat: added voyager in backend

darshi1337:
See issue #1824.


JoanFM:
Hello @darshi1337 ,

We need for you to sign the DCO agreement by signing off your commits

JoanFM:
Thanks for the contribution,

since it seems to be a work in progress (no tests, print statements, etc...) I will put it as a Draft.

docarray/index/backends/voyager.py
```diff
@@ -0,0 +1,263 @@
...
+        search_field: str = '',
+    ) -> _FindResultBatched:
+        # You can implement batched text search logic here
+        # For example, execute each text search query separately and combine the results
+        raise NotImplementedError(
+            f'{type(self)} does not support text search batched queries.'
+        )
+
+
+class NdArray:
```
JoanFM, line 239: why do u need this?

docarray/index/backends/voyager.py
```diff
@@ -0,0 +1,263 @@
...
+                or info.db_type == np.ndarray
+            ):
+                continue
+            columns.append((col, info.db_type))
+
+        columns_str = ', '.join(f'{name} {type}' for name, type in columns)
+        if columns_str:
+            columns_str = ', ' + columns_str
+
+        query = f'CREATE TABLE IF NOT EXISTS docs (doc_id INTEGER PRIMARY KEY, data BLOB{columns_str})'
```
JoanFM, line 48: does this make that column data will not be filterable at all?

JoanFM (changes requested):
I do not see any obvious missing thing. Just to clarify what is possible and not possible with the backend.

docarray/index/backends/voyager.py
darshi1337: The SQL query provided creates a table named `docs` with a `data` column defined as `BLOB` (Binary Large Object). The `BLOB` data type is generally used for storing binary data, and it does not inherently support filtering or indexing.

If the intention is to make the `data` column filterable, I might want to consider using a different data type based on the nature of the data being stored. For example, if the `data` column contains text-based information, changing the data type to `TEXT` could be more appropriate.

Here's an example modification to the query:

```python
if columns_str:
    columns_str = ', ' + columns_str

query = f'CREATE TABLE IF NOT EXISTS docs (doc_id INTEGER PRIMARY KEY, data TEXT{columns_str})'

docarray/index/backends/voyager.py
darshi1337: The `NdArray` class was created to provide a versatile and efficient solution for managing multidimensional arrays in Python. It aims to enhance efficiency, enabling optimized storage and access compared to standard Python data structures. The class also supports mathematical operations relevant to scientific computing, machine learning, and data analysis. Additionally, it offers custom functionality, including methods for reshaping, slicing, and applying filters. The abstraction of data representation enhances code readability, and seamless integration with libraries like NumPy ensures compatibility. 

If you want I can remove it completely.

darshi1337: I am removing this.

docarray/index/backends/voyager.py
JoanFM: this is what is done in HNSWDocumentIndex I believe

docarray/index/backends/voyager.py
JoanFM: Please remove this

JoanFM (changes requested):

docarray/index/backends/voyager.py
darshi1337: Yes like something similar

JoanFM:
Hello @darshi1337 ,

is there any progress to be expected here?

PR #1847, opened 2024-01-16
Fixed incorrect type hint in hnswlib.py

James4Ever0:
The `filter` parameter in `knn_query(data, k = 1, num_threads = -1, filter = None)` accepts an iterable of `int`, instead of `str`.

Accordingly, `hashed_ids` in `_search_and_filter(self, queries, limit, search_field = '', hashed_ids = None,` shall be `Optional[Set[int]]`.


JoanFM (approved):
Thanks a lot for the contribution.

PR #1852, opened 2024-02-08
feat: support list of strings for Weaviate

### LINKED ISSUE #1849: Enhancement to the weaviate datatypes support (text[], object, object[]) with WeaviateDocumentIndex
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://docs.docarray.org) and still think this feature is missing

### Description

There is a limitation with the translation of docarray data types with the data types in weaviate.

Currently the list of datatypes is limited to the list included here: 
docarray/index/backends/weaviate.py line 247
```
        default_column_config: Dict[Any, Dict[str, Any]] = field(
            default_factory=lambda: {
                np.ndarray: {},
                docarray.typing.ID: {},
                'string': {},
                'text': {},
                'int': {},
                'number': {},
                'boolean': {},
                'number[]': {},
                'blob': {},
```
line 710
```
        py_weaviate_type_map = {
            docarray.typing.ID: 'string',
            str: 'text',
            int: 'int',
            float: 'number',
            bool: 'boolean',
            np.ndarray: 'number[]',
            bytes: 'blob',
        }
```
line 197 create_schema
-  would need to accommodate the new data types in the schema creation.

The lists outlined above are more limited than the supported data types in weaviate:
https://weaviate.io/developers/weaviate/config-refs/datatypes

We are looking to support text[] -> list of strings, object and object[] data types in order to fully leverage the weaviate storage.

One of the motivations is to have simpler data storage and also to be able to make use of weaviate's new filters : ContainsAny, ContainsAll - https://weaviate.io/developers/weaviate/api/graphql/filters#filter-structure
At the moment we need to serialize our array as a string and use Like operators which is not ideal and detrimental to performance of queries.

These filters are essentially to find conditions within arrays of values. These data types (object[] and text[] aren't supported when indexing data with the WeaviateDocumentIndex.

 
                

### Affected Components

- [X] [Vector Database / Index](https://docs.docarray.org/user_guide/storing/docindex/)
- [ ] [Representing](https://docs.docarray.org/user_guide/representing/first_step)
- [X] [storing](https://docs.docarray.org/user_guide/storing/first_step/)
- [ ] [multi modal data type](https://docs.docarray.org/data_types/first_steps/)

JoanFM:


docarray/index/backends/weaviate.py
```diff
@@ -715,8 +719,26 @@ def python_type_to_db_type(self, python_type: Type) -> Any:
             bool: 'boolean',
             np.ndarray: 'number[]',
             bytes: 'blob',
+            list: 'text[]',
         }
 
+        if get_origin(python_type) == list:
+            py_weaviate_list_type_map = {
+                int: 'number[]',
+                float: 'number[]',
```
hsm207, line 728: a list of int in python should be mapped to 'int[]' in weaviate, since 'number[]' is for float64

hsm207 (changes requested):
thanks for implementing this.

i just have 1 comment.

hsm207 (approved):

PR #1865, opened 2024-02-14
Hotfix/optional leaf object

oytuntez:
This PR aims to improve `DocList` fields that are `Optional` or that are given `None` default value. Giving empty `[]` as default value would create entries in the subindex (with just the ID field, no other data), so I opted for `None`.

The biggest issue that this PR aims to solve is `DocList` type fields would be required no matter if they have a default value or are marked as `Optional`.

### DIFF
--- a/docarray/base_doc/doc.py
+++ b/docarray/base_doc/doc.py
@@ -178,6 +178,7 @@ class BaseDocWithoutId(BaseModel, IOMixin, UpdateMixin, BaseNode):
         :param field: name of the field
         :return:
         """
+
         if is_pydantic_v2:
             annotation = cls._docarray_fields()[field].annotation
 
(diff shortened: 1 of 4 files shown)


JoanFM (changes requested):
Thanks for the PR, is there a chance for you to add tests for the fixes? Also, I see code related to ES, does it belong to the same problem?

JohannesMessner:
+1 for adding a test, once there is a test I could take over for the review if you want @JoanFM

JoanFM:
> +1 for adding a test, once there is a test I could take over for the review if you want @JoanFM

Perfect!

JoanFM:
Also @oytuntez, please be aware that we will need to have all the commits `signed-off` for us to merge the contributions.

Issue #1875, opened 2024-03-06
Loading audio tensors fails: `ValueError: all input arrays must have the same shape`


chrisammon3000:
### Initial Checks

- [X] I have read and followed [the docs](https://docs.docarray.org/) and still think this is a bug

### Description

I have created subclips of a video in .mp4 using ffmpeg (through moviepy):
```py
# moviepy.video.io.ffmpeg_tools.ffmpeg_extract_subclip

def ffmpeg_extract_subclip(filename, t1, t2, targetname=None):
    """ Makes a new video file playing video file ``filename`` between
        the times ``t1`` and ``t2``. """
    name, ext = os.path.splitext(filename)
    if not targetname:
        T1, T2 = [int(1000*t) for t in [t1, t2]]
        targetname = "%sSUB%d_%d.%s" % (name, T1, T2, ext)
    
    cmd = [get_setting("FFMPEG_BINARY"),"-y",
           "-ss", "%0.2f"%t1,
           "-i", filename,
           "-t", "%0.2f"%(t2-t1),
           "-map", "0", "-vcodec", "copy", "-acodec", "copy", targetname]

    subprocess_call(cmd)
```

Output:
<img width="500" alt="image" src="https://github.com/docarray/docarray/assets/20426965/cf828df3-82e7-4eea-9e67-81838a1e8ecc">

The subclip path is passed to `VideoUrl`:
```py
subclip = VideoUrl("<subclip_path>")
```

Trying to load the tensors fails:
```py
tensors = subclip.load()
```
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[21], line 1
----> 1 tensors = subclip.load()

File ~/Projects/chrisammon3000/experiments/docarray/docarray-test/.venv/lib/python3.11/site-packages/docarray/typing/url/video_url.py:96, in VideoUrl.load(self, **kwargs)
     33 """
     34 Load the data from the url into a `NamedTuple` of
     35 [`VideoNdArray`][docarray.typing.VideoNdArray],
   (...)
     93     [`NdArray`][docarray.typing.NdArray] of the key frame indices.
     94 """
     95 buffer = self.load_bytes(**kwargs)
---> 96 return buffer.load()

File ~/Projects/chrisammon3000/experiments/docarray/docarray-test/.venv/lib/python3.11/site-packages/docarray/typing/bytes/video_bytes.py:86, in VideoBytes.load(self, **kwargs)
[...]

PR #1880, opened 2024-04-03
fix: pad audio arrays to same shape if sample rates differ

chrisammon3000:
## Description
Please refer to #1875 for a detailed description of the issue this fixes. When loading


JoanFM (changes requested):
It would be nice to add a test case to show that the problem is fixed.

JoanFM:
Hey @chrisammon3000 ,

Thanks a lot for the contribution.

However we would ask you to submit a test case to make sure the error case is covered and to make sure that the commits are `signed-off` to ensure that we can merge the contribution.

Please take a look at our contributing guidelines (https://github.com/docarray/docarray/blob/main/CONTRIBUTING.md)

